PR #14, opened 2023-05-13
[BUG] made build tools files executable

yarnabrina:
Changing file modes of shell scripts in `build_tools`.

```zsh
❯ chmod +x build_tools/fail_on_missing_init_files.sh
❯ chmod +x build_tools/run_examples.sh
```


fkiraly:
thanks!

can you explain why this works and how you arrived there?
Would be keen to learn.

fkiraly (approved):

yarnabrina:
This was the [error](https://github.com/sktime/skpro/actions/runs/4967743363/jobs/8889923544#step:7:5):

> /home/runner/work/_temp/eb86de82-ffe2-410b-a2c0-761c6bcf4458.sh: line 1: build_tools/fail_on_missing_init_files.sh: Permission denied

And this was the code:

```yaml
      - name: Check for missing init files
        run: build_tools/fail_on_missing_init_files.sh
        shell: bash
```

Now, I knew `bash` can't run any file directly in shell unless it's executable, one has to make them executable or do `bash <file_name>`. In case there are special scenarios, I am unaware, but don't really care much as I am not a shell person and will probably never be.

I faced similar issues in Gitlab CI before with my work projects, and switched to auto check of this sort of things with `pre-commit` hooks since then (`check-shebang-scripts-are-executable` and `check-executables-have-shebangs`). It was same here, checked the file mode, saw it's not executable, changed and it worked.

fkiraly:
I see. But, how is this different from `sktime` or `skbase` where the same works? It is a content identical copy, this confuses me.

yarnabrina:
File contents do not make a file executable, file mode does. The files in `skbase` and `sktime` are already executable. You can try changing file modes on these repositories as well, e.g. make them non-executable with `chmod -x <file_name>`, and they should fail in CI afterwards.

I know too less about these stuff to explain well. These can give some ideas, and there can be other better references:

https://en.wikipedia.org/wiki/File-system_permissions
https://linuxhint.com/linux_chmod_command_tutorial_beginners/
https://www.redhat.com/sysadmin/linux-file-permissions-explained
https://kb.iu.edu/d/abdb

Issue #10, opened 2023-05-11
umbrella issue - `scikit-base` based rearchitecture for `skpro` v2


fkiraly:
- [ ] 1st part: rearchitecting on scikit-base https://github.com/sktime/skpro/pull/9 - 
  this unlocks creating new estimators, distributions, and moving legacy estimators to scikit-base compatible interface
- [ ] 2nd part to cover probabilistic metrics and example notebooks
  this unlocks more metrics and moving legacy metrics

Moving individual estimators from legacy to `scikit-base` interface

- [ ] complete missing features in residual two-step
- [ ] bagging estimator
- [ ] density estimator naive baseline (should include kernel and empirical
- [ ] composite parametric
- [ ] bounding wrapper
- [ ] grid and random tuning (requires metrics)

Moving distirbutions from legacy to `scikit-base` interface

- [ ] Laplace
- [ ] Cauchy
- [ ] t
- [ ] empirical
- [ ] mixture (compositor)

Moving metrics from legacy to `scikit-base` interface - can be reused from `sktime`

- [ ] CRPS
- [ ] log-loss
- [ ] capped log-loss
- [ ] squared loss

PR #18, opened 2023-05-20
Rearchitecting `skpro` on `skbase` - part 3: `predict_interval`, `predict_quantiles`

fkiraly:
This PR is a major rearchitecting of `skpro` based on `skbase` and the `sktime` probabilistic prediction framework.
Part 3 of #10.
Contents:

* `datatypes` module for checks and conversions, `Proba` and `Table` scitypes only, adapted from `sktime`
* methods for interval and quantile predictions, mirroring interface in `sktime`
* probabilistic performance metrics and tests for interval and quantile predictions, adapted from `sktime`
* simple notebook examples with the base vignette for interval and quantile predictions
* some bugfixes and added test coverage for base vignette

Also contains some fixes and changes:

* corrections to distribution metrics' `multioutput` logic and test
* duplicative `multioutput` arguments removed
* corrections to docstrings, e.g., structure assumptions corrected or clarified
* `BaseDistribution.quantile` column creation logic fixed


fkiraly:
ready to review, @frthjf !

skpro/datatypes/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    "get_examples",
+    "mtype_to_scitype",
+    "MTYPE_REGISTER",
+    "MTYPE_LIST_PROBA",
+    "MTYPE_LIST_TABLE",
+    "scitype",
+    "scitype_to_mtype",
+    "SCITYPE_LIST",
+    "SCITYPE_REGISTER",
+    "update_data",
```
frthjf, line 44: `update_data` seems to be missing

skpro/datatypes/_check.py
```diff
@@ -0,0 +1,482 @@
...
+from skpro.datatypes._registry import AMBIGUOUS_MTYPES, SCITYPE_LIST, mtype_to_scitype
+from skpro.datatypes._table import check_dict_Table
+
+# pool convert_dict-s
+check_dict = dict()
+check_dict.update(check_dict_Table)
+check_dict.update(check_dict_Proba)
+
+
+def _check_scitype_valid(scitype: str = None):
```
frthjf, line 42: Shouldn't this be `_check_scitype_valid(scitype: str) -> None`?

skpro/datatypes/_check.py
```diff
@@ -0,0 +1,482 @@
...
+    return obj
+
+
+def check_is_mtype(
+    obj,
+    mtype: Union[str, List[str]],
+    scitype: str = None,
+    return_metadata=False,
+    var_name="obj",
+):
```
frthjf, line 88: I guess the type annotations are incomplete, so ignore my previous comment.

skpro/datatypes/_convert.py
```diff
@@ -0,0 +1,310 @@
...
+        # note: this is a side effect on store
+        store.clear()
+    elif store_behaviour == "update":
+        # store is passed to convert_obj by reference, unchanged
+        # this "elif" is here for clarity, to cover all three values
+        pass
+    else:
+        raise RuntimeError(
+            "bug: unrechable condition error, store_behaviour has unexpected value"
+        )
```
frthjf, line 164: nit: seems like a `ValueError` rather than runtime since it could be caught statically as well.

frthjf (approved):
Looks all good to me! 

Reading through it I had two more general thoughts that shouldn't hold up this PR but are perhaps worth noting:

1) Could the type checking infrastructure (mainly in `skpro/datatypes`) be moved into `skbase` (this may already be the plan, in this case ignore)
2) Or, if the sci-types vary, could the sci-type checking be realized within the type annotations instead of runtime checks so that static analyzers (mypy, IDEs) can pick up on it? (again, I am not too familiar with sci-types, this may be something that requires runtime information, in this case ignore)

skpro/datatypes/_convert.py
fkiraly: yes, ok, changed

skpro/datatypes/__init__.py
fkiraly: yes, removed that, only needed for forecasting.

PR #33, opened 2023-08-25
DIAGNOSTIC [MNT] Break some objects to check validity of tests

fkiraly:
This PR intentionally breaks some methods of objecets to check whethere the tests find the broken methods:

* `Mixture.mean`
* `Normal.mean` and `Normal.var`
* `ResidualDouble.predict`


fkiraly:
correctly validates test in https://github.com/sktime/skpro/pull/34, closing

PR #45, opened 2023-08-26
[MNT] `pre-commit` hook upgrade

fkiraly:
Upgrades `pre-commit` versions and steps to the same as `sktime`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.0.1
+    rev: v4.4.0
     hooks:
       - id: check-added-large-files
         args: ["--maxkb=1000"]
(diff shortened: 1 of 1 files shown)


fkiraly:
@yarnabrina, can I ask you a question about pre-commit?

This is just copy-pasting `sktime`-s, and I remember when you did that a while ago, it complained about the `# -*- coding: utf-8 -*-` lines and `super()` with arguments.

However, this does not seem to be happening here - and I'm confused why not. Is that not controlled by pre-commit?

yarnabrina:
`pre-commit` runs by default only on the modified files, which in this case
is just the configuration file itself and nothing else. So, new hooks have
no effect whatsoever.

You have to handle this locally as a one time process. You can force it by
running this:

```shell
pre-commit run --all-files
```

This will run all hooks on all files tracked by `git`. If you want to
restrict which hooks to run, just pass afterwards.

fkiraly:
> `pre-commit` runs by default only on the modified files

Oh, of course - thanks! That's the main point I missed.

PR #60, opened 2023-09-08
[MNT] repository linting

fkiraly:
Lints repository files after upgrading pre-commit hooks in https://github.com/sktime/skpro/pull/45


fkiraly:
remaining linting failures only in legacy code

PR #61, opened 2023-09-08
[DOC] fix loc/scale parameter names in `TDistribution` docstring

fkiraly:
Fixes location-scale parameter names in the docstring of `TDistribution`, the parameter names in the docstring were inconsistent with the actual parameter names.

FYI @Alex-JG3, before we release we can easily switch that either way - do you have a preference?
My default action is always to fix the docstring if the logic and the docstring disagree.


Alex-JG3:
> FYI @Alex-JG3, before we release we can easily switch that either way - do you have a preference?

I have no preference, I am happy with fixing the docstring. Thanks for making this change.

PR #102, opened 2023-09-13
[MNT] simplify the CI trigger

fkiraly:
Simplifys the CI tigger condition, now same as `sktime`


fkiraly:
closed for now - not sure if it's an improvement

PR #108, opened 2023-10-05
[ENH] multiple quantile regression

### LINKED ISSUE #107: [ENH] multiple quantile regression

For quantile regression, often more than one quantile probability is of interest. However, existing Sklearn compatible quantile regressors always fit and predict a single quantile probability. To the best of my knowdlegde, there is no standardized way to integrate multiple quantile regression with Sktime/Skpro probabilistic prediction methods such as predict_quantiles/predict_intervals.

New Skpro regressor that wraps multiple quantile regressors and supports probabilistic predictions from wrapped regressors.

None, as the proposed solution is already discussed in Sktime issue https://github.com/sktime/sktime/issues/5357

Ram0nB:
#### Reference Issues/PRs

Fixes #107

For quantile regression, often more than one quantile probability is of interest. However, existing Sklearn compatible quantile regressors always fit and predict a single quantile probability. To the best of my knowdlegde, there is no standardized way to integrate multiple quantile regression with Sktime/Skpro probabilistic prediction methods such as predict_quantiles/predict_intervals. This PR adds new Skpro regressor that wraps multiple quantile regressors and supports probabilistic predictions from wrapped regressors.

No

All added/changed files :)

No - I ran the Skpro tests locally without issues

#### Any other comments?

No

##### For all contributions

##### For new estimators


skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,362 @@
...
+    Parameters
+    ----------
+    mean_regressor : Sklearn compatible regressor, optional
+        Tabular mean regressor for predict.
+    quantile_regressor : Sklearn compatible quantile regressor, optional
+        Tabular quantile regressor. In fit, for every alpha a clone of
+        quantile_regressor is made whereafte the alpha parameter
+        quantile_regressor_alpha_param is set using the set_params method.
+        Subsequently, all regressors are fitted.
+    quantile_regressor_alpha_param : str, optional
```
fkiraly, line 31: docstring should also list defaults, i.e., what happens if parameter is not passed.

skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,362 @@
...
+        the crossing of quantiles, also known as the quantile crossing problem. This can
+        happen because each quantile is predicted by a single independent instance of
+        the quantile_regressor. To resolve this issue, the quantiles for
+        all predictions per row index will be sorted if sort_quantiles=True. This
+        methodology will never lead to a higher quantile loss and solves the quantile
+        crossing problem [1]_.
+
+    [1] Victor Chernozhukov, Iván Fernández-Val, and Alfred Galichon. Quantile and
+        probability curves without crossing. Econometrica, 78(3):1093–1125, 2010.
+
```
fkiraly, line 50: I would recommend to list `regressors_` in an `Attributes` section.

skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,362 @@
...
+        # put fitted regressors in dict and write to self
+        regressors_ = {}
+        if not self._no_mean_reg:
+            regressors_["mean"] = fitted_regressors.pop(0)
+        for i, a in enumerate(self.alpha):
+            regressors_[a] = fitted_regressors[i]
+        self.regressors_ = regressors_
+
+        # write varname to self
+        if isinstance(y, pd.DataFrame):
```
fkiraly, line 166: the default boilerplate in `fit` already coerces `y` to `pd.DataFrame`

fkiraly: (maybe the extension template docstring should be updated)

skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,362 @@
...
+        Returns
+        -------
+        y : pandas DataFrame, same length as `X`
+            labels predicted for `X`
+        """
+        if self._no_mean_reg:
+            raise ValueError("Can't predict: no mean_regressor provided")
+
+        # predict
+        X_pred = np.array(X)
```
fkiraly, line 199: `X.values` is the canonical way to coerce `DataFrame` to 2D `np.ndarray`

skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,362 @@
...
+            * if `deep=True`, also contains keys/value pairs of component parameters
+              parameters of components are indexed as `[componentname]__[paramname]`
+              all parameters of `componentname` appear as `paramname` with its value
+            * if `deep=True`, also contains arbitrary levels of component recursion,
+              e.g., `[componentname]__[componentcomponentname]__[paramname]`, etc
+        """
+        params = super().get_params(deep=deep)
+        # remove quantile_regressor_alpha_param from params
+        # managed by self, so removal prevents user confusion
+        params.pop(f"quantile_regressor__{self.quantile_regressor_alpha_param}", None)
```
fkiraly, line 302: I don't think we want that? Or, do you want to use different regressors per alpha? That would work differently (let me know if that is what you want to build)

fkiraly (changes requested):
Nice addition, thanks! This will be useful in `sktime`!

Blocking comments below.

Docs:

* defaults should be stated in docstring
* the exact algorithm should be stated in docstring (even if it is simple)

Algorithm:

* can we implement something for `predict_proba`? I would simply put masses on the quantiles we predict, such that any quantiles closest to that will be supported at the same point.
    * let's say, we predict three quantiles, 0.1, 0.5, 0.9. Then, we put masses 0.3, 0.4, 0.3 on the respective predictions, and get a weighted empirical distribution with support at three points.
    * this will allow sidestep the "no regressors for alpha" problem, and give the correct predictions for the correct `alpha`.
    * if this `predict_proba` is implemented, the remaining methods are filled in automatically, although one might think about efficiency.

Interface requirements:

* `get_params` should not be overwritten like this.
    * do you want to allow different quantile regressors per quantile point? If yes, we would need to use the heterogenous ensemble base class. Perhaps better in a separate PR since it's a bit fiddly?
* `get_test_params` should return two or more test parameter settings.

Ram0nB:
Many thanks for your very quick feedback, much appreciated! 

> * `get_params` should not be overwritten like this.
>   
>   * do you want to allow different quantile regressors per quantile point? If yes, we would need to use the heterogenous ensemble base class. Perhaps better in a separate PR since it's a bit fiddly?


Per probability level in alpha I intend to use one regressor. Since an instance of a QuantileRegressor is passed in __init__, it likely already has a quantile probability level assigned to it. The value assigned to the quantile probability parameter is however overridden by the logic in fit: per quantile probability in alpha, I make a copy of the quantile regressor and set the corresponding probability with set_params. This initial quantile probability level would also be exposed to the user with get/set deep params, although changing it with those methods wouldn't make any difference. Therefore, I thought it would be best if the quantile probability of the QuantileRegressor instance passed in __init__ wouldn't be exposed to the user with get/set params methods. Can you think of a better way to handle this?

I didn't know that the heterogenous ensemble forecaster base class exists and I've just had a look at it. It's nice that it has functionality for fitting and predicting with multiple estimators (similar to what I'm doing). Using a similar base class for a probabilistic regressor would resolve the aforementioned issue, however I can also think about a few cons:

- For __init__, the user would have to create and list all the QuantileRegressor instances for the desired probabilies rather than just provide the desired probabilities and one QuantileRegressor instance.
- For parameter optimisation, the user would need to set the hyperparameters for each quantileregressor seperately if we'd use a heterogenous ensemble base. (with that base class the hyperparameters of the nested estimators can be set seperately, right?). Currently, hyperparameters can only be set for all QuantileRegressors at once. Since fitting multiple quantile regressors is a computationally intensive task, I don't think that many users would tune hyperparameters per quantile regressor, but rather for all quantile regressors in once. Therefore, if we'd use a heterogenous ensemble base it would complicate the hyperparameter tuning process for most users compared to the current implementation.
- Current heterogenous ensemble base class not available for probabilistic regressors but rather only for Sktime's forecasters.

Personally (with my current understanding), I tend towards not using the heterogenous ensemble base class as it seems like it would complicate things without significant benefit. What are your thoughts on this?



> Algorithm:
> 
> * can we implement something for `predict_proba`? I would simply put masses on the quantiles we predict, such that any quantiles closest to that will be supported at the same point.
>   
>   * let's say, we predict three quantiles, 0.1, 0.5, 0.9. Then, we put masses 0.3, 0.4, 0.3 on the respective predictions, and get a weighted empirical distribution with support at three points.
>   * this will allow sidestep the "no regressors for alpha" problem, and give the correct predictions for the correct `alpha`.
>   * if this `predict_proba` is implemented, the remaining methods are filled in automatically, although one might think about efficiency.

Something for `predict_proba` would indeed be a nice addition. Can you expand a bit on why we'd put masses on the quantiles rather than use similar weights?

I'm not really familiar with the implementation of distributions in Sktime/Skpro, but would it also be possible to generate and return empirical distributions from the predicted quantiles in `predict_proba`? For the cdf we could interpolate between (sorted) quantiles and use step functions for the tails since we don't have any information about it. This would reflect the main logic of the MultipleQuantileRegressor best I think. Moreover, we could use the aforementioned cdf function to sidestep the "no regressors for alpha" problem in `predict_quantiles`. What are your thoughts on this?

fkiraly:
> Therefore, I thought it would be best if the quantile probability of the QuantileRegressor instance passed in init wouldn't be exposed to the user with get/set params methods. Can you think of a better way to handle this?

I see. The important point here is, get/set params methods *must* return and access the same parameters as passed to `__init__`. This is one of the "great laws of scikit-learn" that should not be deviated from. One can of discuss sense and sensibilities, but deviating means becoming incompatible with a large ecosystem, so we (and the tests) have to enforce it.

But of course that still leaves up to discussion and choice the exact parameterization. So, regarding "better ways", I think your parameterization is probably the best and most user friendly. Overall I can think of:

* estimator, alpha-name, alpha-vals, that's basically yours and would be also my a-priori first preference (up to, perhaps shortening the parameter names a bit)
* `[(est, alpha1), (est2, alpha2), ...]` syntax with list of tuples like in `sklearn` pipelines. Worse imo since a user using the same estimator has to do a lot of manual cloning
* that, but with an option to pass sth like `[(est, alpha1, ..., alphan), (est2, alphaprime)]`. I think this is only slightly but still worse than your choice, even if it allows more freedom in the ensembling.

fkiraly:
> however I can also think about a few cons

Regarding the heteorogenous ensemble, I think you list its main drawbacks correctly.

* the drawbacks in the parameterization are as discussed, I agree with your assessment, and that's why I think your choice in parameterization is better. I think it's worth sacrificing flexibility for usability. Flexibility can be achieved by chaining with a multiplexer.
* though, I mean the heterogenous base class in `scikit-base`, which can be used for arbitrary objects. True though, it has not been used in `skpro` extensively yet (but it exists, e.g., look at the pipeline), so there would likely be some fiddling around. Probably not worth the effort in this special case.

> I tend towards not using the heterogenous ensemble base class as it seems like it would complicate things without significant benefit. What are your thoughts on this?

Agreed from a general perspective - I would only suggest to use it if you want different regressor per alpha, but I think you have explained that that's not what you actually want to do, so we agree on the "should" and on the "how" both.

fkiraly:
> Something for `predict_proba` would indeed be a nice addition. Can you expand a bit on why we'd put masses on the quantiles rather than use similar weights?

If we put same weights, then the quantiles will in general not be the quantiles anymore. If you go with empirical with the same support, you need to put weights with the property so the cumulative weights at the chosen quantile match. There is in general an infinity of such choices, and the "vornoi/closest" by measure seems most canonical.

> For the cdf we could interpolate between (sorted) quantiles and use step functions for the tails since we don't have any information about it. This would reflect the main logic of the MultipleQuantileRegressor best I think. Moreover, we could use the aforementioned cdf function to sidestep the "no regressors for alpha" problem in predict_quantiles. What are your thoughts on this?

I have thought about the above as well, it was my first thought actually and I rejected it because:

* both proposed solutions sidestep the "no regressors for alpha" issue
* the choice of cdf above would require to implement new "interpolative" distribution types and will require to make arbitrary choices for the tails. The cdf would hence carry severely more implementative burden, as well as more information than the predicted quantiles.
* whereas, the proposed empirical distribution carries exactly as much information as the prediced quantiles - in fact in the parameterization of `Empirical` you just vertical-stack the matrix of predictive quantiles, and there is no further fiddling, it's just "plugin". (well, there are the weights, but they are easy to get).

fkiraly:
PS: I'm happy to implement the `predict_proba` if you produce for me a private utility function which, given the `predict`-`X`, produces a `pd.DataFrame` with the following specifications:

* row multiindex (`alpha`, `X.index`)
* column index as `y` in `fit`
* entry is the quantile prediction at quantile `alpha` and the respective `X`-index point

fkiraly:
here's another argument for implementing the variant with "empirical":

Imagine a probability calibrator (a transformer distr to distr) acting on empiricals, applying a smoother or linear interpolator like you propose, @Ram0nB.

Now your version can be easily obtained from my version, by appending. The other way round is not true, as the transformation is not invertible, or at least not easily so in the generic case.

Therefore, the "@Ram0nB version" is actually a pipeline (transformed target regressor for proba) of the "@fkiraly version" with a probability calibrator that does the smoothin or interpolation of the empirical.

Ram0nB:
Thanks for the clarification @fkiraly, I agree with all your points. I processed all the requested changes, looking forward to your feedback :smiley: .

> PS: I'm happy to implement the `predict_proba` if you produce for me a private utility function which, given the `predict`-`X`, produces a `pd.DataFrame` with the following specifications:
> 
> * row multiindex (`alpha`, `X.index`)
> * column index as `y` in `fit`
> * entry is the quantile prediction at quantile `alpha` and the respective `X`-index point

This would be great, thanks! The utility function is called `_predict_proba_util`.

fkiraly:
ok, will look at it after work today - 3.12 compatible releases of skpro and sktime having priority first, but that should be out of the way quickly (I hope...)

skpro/regression/multiquantile.py
```diff
@@ -0,0 +1,335 @@
...
+    ----------
+    mean_regressor : Sklearn compatible regressor or None, default=None
+        Tabular mean regressor for `predict`. Can't predict mean if not provided.
+    quantile_regressor : Sklearn compatible quantile regressor or None, default=None
+        Tabular quantile regressor for `predict_quantiles`, `predict_interval` and
+        `predict_proba`. Can't do probabilistic prediction if not provided.
+    alpha_name : str or None, default=None
+        Alpha parameter name that sets the quantile probability level for the
+        quantile_regressor. Can't do probabilistic prediction if not provided.
+    alpha : list with float or None, default=None
```
fkiraly, line 40: what happens if `alpha` is None? It looks to me like it will just fail then. So, should `alpha` not be compulsory?

Ram0nB:
Many thanks for the quick `_predict_proba` draft, worked right away! 

> what happens if `alpha` is None? It looks to me like it will just fail then. So, should `alpha` not be compulsory?

Only the prob. prediction methods wouldn't work as I did an input check in init, but it indeed makes more sense to make all parameters for the prob prediction methods non-optional. 

The newest version has the following changes:
- prob prediction parameters non-optional
- finalised `_predict_proba`
- updated docstring

Do you think that the docstring explains the "algorithm" for all prob prediction methods extensive enought?

fkiraly:
ok, let's see if it runs! 😁

fkiraly:
> Do you think that the docstring explains the "algorithm" for all prob prediction methods extensive enought?

I suppose as good as it can be without math (perhaps instead of "nearest probability", "nearest quantile probability"). ALthough I think it can be further improved by adding a little math.

fkiraly:
linting fails because you have an empty notebook in the branch, kindly remove

Ram0nB:
> One thing that might be worth to check before we merge, if the user passes non-sorted `alpha` either to `predict_quantiles`, or to `__init__`, is all the logic still correct? I have a nagging feeling that it might not be, and I'm not sure whether we check in the tests either.

Good one! This was indeed not the case. The following changes are made:
- Where appropriate alpha is sorted to prevent anomalies. 
- Docstring updated with a little math.

Unfortunately, I couldn't get the doc building for Skpro to work to check if it renders correctly (doc building not in Skpro's dev guide but only in Sktime's dev guide). Can you check this for me or provide me with a resource that explains how I can do so myself? Many thanks!

fkiraly:
> I see that the tests are failing, I'll have a look at it tomorrow :)

Is this readthedocs again - might be unrelated to you, I thought I fixed it: https://github.com/sktime/skpro/issues/122

docs/source/api_reference/regression.rst
```diff
@@ -63,6 +63,14 @@ take one or multiple ``sklearn`` estimators and adda probabilistic prediction mo
 
     ResidualDouble
 
+.. currentmodule:: skpro.regression.multiquantile
+
+.. autosummary::
```
fkiraly, line 68: I think this is the source of the readthedocs problem, but I cannot figure out what would b the problem. All seems fine to me.

fkiraly:
ah, I think it's the docstring. If you use TeX, and hence the backslash character, the string needs to be preceded by an "r" character, i.e.,

```python
r"""This is a docstring.

This is a docstring description with :math:`\pi`
"""

File: skpro/regression/multiquantile.py [hunk 8]
```diff
@@ -0,0 +1,369 @@
...
+
+    In `fit`, for every probability in alpha, the quantile_regressor is cloned and the
+    probability is set. Subsequently all regressors are fitted.
+
+    For every desired quantile probability in the probabilistic prediction methods, the 
+    fitted quantile regressor that is nearest to the desired quantile probability is 
+    selected to make the corresponding prediction. This is done as follows: we have a 
+    list of fitted quantile probabilities (init alpha) :math:`\alpha = [\alpha_1, 
+    \alpha_2, \ldots, \alpha_n]` and a list of desired quantile probabilities (predict 
+    alpha) :math:`\alpha = [\alpha_1, \alpha_2, \ldots, \alpha_n]`. The selected
```
fkiraly, line 31: these should not use the same symbol

fkiraly:
updated the docstring with the `r` char, and also made some small changes to the math paragraph (hope that's ok) - now it should reach the tests and not fail readthedocs.

fkiraly:
FYI, the regressor was failing a test since the `alpha` in the columns of `predict_quantiles` are expected in the same order as the `alpha` argument, which needs not be sorted.

I made two small changes which should fix that - feel free to revert or change.

fkiraly (approved):
Works now, with my changes, see above.
Happy to merge if you are happy with my changes.

Ram0nB:
Many thanks for your changes, all looks great to me! 👍

fkiraly:
Thanks for your contribution!

Once released in 2.1.1, you should be able to use this with `sktime`.

PR #127, opened 2023-10-10
[MNT] ensure pytest options are set in `make test`

fkiraly:
`make test` does not seem to respect the pytest options - hypothesis, this is due to a lacking pytest options reference in the makefile.


fkiraly:
hm, doesnt seem to have worked...

Issue #134, opened 2023-10-28
[ENH] MAPIE interface


fkiraly:
MAPIE is one of the more popular "favourite algorithm" repositories in the probabilistic supervised learning space.

Their scope is more geared towards multiple tasks (incl classification, time series forecasting, etc), whereas `skpro` has a more stringent framework layer imo.

The specific estimator that would be great to interface is the `MapieRegressor`:
https://github.com/scikit-learn-contrib/MAPIE/blob/master/mapie/regression/regression.py

should be straightforward with the extension template.
https://github.com/sktime/skpro/blob/main/extension_templates/regression.py

FYI @gmartinonQM, @thibaultcordier, @LacombeLouis, @vincentblot28, @vtaquet, @SZiane in case there might be interest in close collaboration - I see you are planning to add further models, and you are also following closely sklearn as a design template, so perhaps we should merge base class architectures?

For context, `skpro` is a 2017 package for proba supervised (tabular) regression which is one of the design antecessors of `sktime`. Parts of it became the `sktime` probabilistic forecasting module, now it has been rearchitected using `skbase`. Of course a number of proba supervised regression packages have seen the light of day since 2017.

`sktime` and `skpro` adopt a framework integration philosophy where third party packages can be easily interfaced and used as components in pipelines etc, we manage dependencies on the level of estimators.
https://www.sktime.net/en/stable/

For instance, `skpro` probabilistic tabular regressors can be used in `sktime` probabilistic forecasting pipelines.

fkiraly:
PS: I will be at Inria in the week of November 6, we could talk in person there.

fkiraly:
I've used this as a test case to stress test the estimator dependency management. PR is here: https://github.com/sktime/skpro/pull/136

thibaultcordier:
Hello @fkiraly, we are curious to learn more about your project and what type of close collaboration you envisage with MAPIE. We'll be able to discuss this with you to get a clearer idea of what you're planning, at the end of this week.

fkiraly:
> We'll be able to discuss this with you to get a clearer idea of what you're planning, at the end of this week.

Excellent! Let me know when is convenient. I don't have your email, but mine is public, so we can do some scheduling there? Will be in Paris this week.

PR #136, opened 2023-10-28
[ENH] interface `MapieRegressor` from `mapie`

fkiraly:
Fixes https://github.com/sktime/skpro/issues/134.

Interfaces `mapie.regression.MapieRegressor`, and introduces `mapie` as an optional soft dependency.

Proof-of-concept for third party package interfaces, similar to `sktime` third party interface handling

Depends on:
* https://github.com/sktime/skpro/pull/138 for manual testing
* https://github.com/sktime/skpro/pull/140 for CI testing
* https://github.com/sktime/skpro/pull/142 for CI testing (bugfix)

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -88,6 +88,17 @@
         "ideas",
         "review"
       ]
+    },
+    {
+      "login": "Ram0nB",
+      "name": "Ramon Bussing",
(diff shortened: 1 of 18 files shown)


fkiraly:
Failures in `run-tests-no-extras` are benign, due to test framework not excluding estimators which require extras. This can be solved by copying the same logic from `sktime` (should probably be in `skbase` at some point).

Issue #132, opened 2023-10-23
[ENH] interface to cyclic boosting package


fkiraly:
It would be nice to interface `cyclic_boosting`, which provides implementations of the cyclic boosting regressors.
https://github.com/Blue-Yonder-OSS/cyclic-boosting/issues/56

They are popular in proba forecasting, but by nature are actually proba supervised regerssion algoritihms so "belong" in `skpro`, from where it can be interfaced by `sktime` for forecasting.

The interface to implement is in this extension template:
https://github.com/sktime/skpro/blob/main/extension_templates/regression.py

FYI @rijkvandermeulen, @FelixWick

fkiraly:
added with https://github.com/sktime/skpro/pull/144

PR #144, opened 2023-11-10
[NEW] Cyclic boosting interface

setoguchi-naoki:
#### Reference Issues/PRs

#132 

J-qpd estimation by quantile regression with Cyclic boosting, A output distribution's parameters are given by j-qpd's loc and scale

cyclic_boosting package

Please check my probability distribution prediction manner is right or not

Not yet

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/regression/cyclic_boosting.py
+++ b/skpro/regression/cyclic_boosting.py
@@ -257,12 +257,12 @@ class CyclicBoosting(BaseProbaRegressor):
             distr_type = Normal
             distr_loc_scale_name = ("mu", "sigma")
             self.var_jqpd = np.sqrt(self.var_jqpd)
-        # TODO: add other distributions
-        # elif distr_type == "Laplace":
-        #     from skpro.distributions.laplace import Laplace
+        elif distr_type == "Laplace":
+            from skpro.distributions.laplace import Laplace
 
-        #     distr_type = Laplace
-        #     distr_loc_scale_name = ("mu", "scale")
+            distr_type = Laplace
+            distr_loc_scale_name = ("mu", "scale")
+            self.var_jppd = np.sqrt(self.var_jqpd) / 2
         # elif distr_type in ["Cauchy", "t"]:
         #     from skpro.distributions.t import TDistribution
 
@@ -322,10 +322,10 @@ class CyclicBoosting(BaseProbaRegressor):
         }
         param1 = {"feature_properties": fp}
         param2 = {"feature_properties": fp, "interaction": [("age", "sex"), ("s1, s3")]}
-        # param3 = {
-        #     "feature_properties": fp,
-        #     "interaction": [("age", "sex"), ("s1, s3")],
-        #     "distr_type": "Laplace",
-        # }
+        param3 = {
+            "feature_properties": fp,
+            "interaction": [("age", "sex"), ("s1, s3")],
+            "distr_type": "Laplace",
+        }
 
-        return [param1, param2]
+        return [param1, param2, param3]
--- a/skpro/regression/cyclic_boosting.py
+++ b/skpro/regression/cyclic_boosting.py
@@ -1,0 +1,351 @@
+"""Cyclic boosting regressors.
+This is a interface for Cyclic boosting, it contains efficient,
+off-the-shelf, general-purpose supervised machine learning methods
+for both regression and classification tasks.
+Please read the official document for its detail
+https://cyclic-boosting.readthedocs.io/en/latest/
+"""
+
+"""License.
+EPL 2.0
+https://github.com/Blue-Yonder-OSS/cyclic-boosting/blob/main/LICENSE
+"""
+
+__author__ = ["setoguchi-naoki"]
+
+import numpy as np
+import pandas as pd
+from scipy.misc import derivative
+from scipy.integrate import quad
+from skpro.regression.base import BaseProbaRegressor
+
+# from cyclic_boosting import common_smoothers, binning
+from cyclic_boosting import (
+    pipeline_CBMultiplicativeQuantileRegressor,
+)
+from cyclic_boosting.quantile_matching import J_QPD_S
+
+
+# todo: change class name and write docstring
+class CyclicBoosting(BaseProbaRegressor):
+    """Cyclic boosting regressor.
+
+    Estimates the parameters of Johnson Quantile-Parameterized Distributions
+    (JQPD) by quantile regression, which is one of the Cyclic boosting's functions
+    this method can more accurately approximate to the distribution of observed data
+
+    Parameters
+    ----------
+    feature_properties : dict
+        name and characteristic of train dataset
+        it is able to set multiple characteristic by OR operator
+        e.g. {sample1: IS_CONTINUOUS | IS_LINEAR, sample2: IS_ORDERED}
+        for basic options, see
+        https://cyclic-boosting.readthedocs.io/en/latest/tutorial.html#set-feature-properties
+    interaction : list[tuple], default=(), optional
+        some combinations of explanatory variables, (interaction term)
+        e.g. [(sample1, sample2), (sample1, sample3)]
+    distr_type: str, default='Normal',
+        probability distribution name which is assumed for observation data's
+        distribution
+
+    Attributes
+    ----------
+    estimators_ : list of of skpro regressors
+        clones of regressor in `estimator` fitted in the ensemble
+    quantiles : list, default=[0.2, 0.5, 0.8]
+        targets of quantile prediction for j-qpd's param
+    quantile_values: list
+        quantile prediction results
+    quantile_est: list
+        estimators, each estimator predicts point in the value of quantiles attribute
+
+    Examples
+    --------
+    >>> from skpro.regression.cyclic_boosting import CyclicBoosting
+    >>> from cyclic_boosting import flags
+    >>> from sklearn.datasets import load_diabetes
+    >>> from sklearn.model_selection import train_test_split
+    >>> X, y = load_diabetes(return_X_y=True, as_frame=True)
+    >>> X_train, X_test, y_train, y_test = train_test_split(X, y)
+    >>>
+    >>> fp = {
+    >>>     'age': flags.IS_CONTINUOUS,
+    >>>     'sex': flags.IS_CONTINUOUS,
+    >>>     'bmi': flags.IS_CONTINUOUS,
+    >>>     'bp':  flags.IS_CONTINUOUS,
+    >>>     's1':  flags.IS_CONTINUOUS,
+    >>>     's2':  flags.IS_CONTINUOUS,
+    >>>     's3':  flags.IS_CONTINUOUS,
+    >>>     's4':  flags.IS_CONTINUOUS,
+    >>>     's5':  flags.IS_CONTINUOUS,
+    >>>     's6':  flags.IS_CONTINUOUS,
+    >>> }
+    >>> reg_proba = CyclicBoosting(feature_properties=fp)
+    >>> reg_proba.fit(X_train, y_train)
+    >>> y_pred = reg_proba.predict_proba(X_test)
+    """
+
+    _tags = {
+        "object_type": "distribution",
+        "estimator_type": "regressor_proba",
+        "capability:multioutput": False,
+        "capability:missing": True,
+        "X_inner_mtype": "pd_DataFrame_Table",
+        "y_inner_mtype": "pd_DataFrame_Table",
+    }
+
+    def __init__(self, feature_properties, interaction=tuple(), distr_type="Normal"):
+        self.feature_properties = feature_properties
+        self.interaction = interaction
+        self.quantiles = [0.2, 0.5, 0.8]
+        self.quantile_values = []
+        self.quantile_est = []
+        self.mean_jqpd = None
+        self.var_jqpd = None
+        self.distr_type = distr_type
+
+        super().__init__()
+
+        # check parameters
+        if not isinstance(feature_properties, dict):
+            raise ValueError("feature_properties must be dict")
+        for i in interaction:
+            if not isinstance(i, tuple):
+                raise ValueError("interaction must be tuple")
+
+        # build estimators
+        features = list(self.feature_properties.keys())
+        for i in interaction:
+            features.append(i)
+
+        for quantile in self.quantiles:
+            self.quantile_est.append(
+                pipeline_CBMultiplicativeQuantileRegressor(
+                    quantile=quantile,
+                    feature_properties=self.feature_properties,
+                    feature_groups=features,
+                    maximal_iterations=100,
+                )
+            )
+
+    def _fit(self, X, y):
+        """Fit regressor to training data.
+
+        Writes to self:
+            Sets fitted model attributes ending in "_".
+
+        Parameters
+        ----------
+        X : pandas DataFrame
+            feature instances to fit regressor to
+        y : pandas DataFrame, must be same length as X
+            labels to fit regressor to
+
+        Returns
+        -------
+        self : reference to self
+        """
+
+        self._y_cols = y.columns
+        y = y.to_numpy().flatten()
+
+        # multiple quantile regression for full probability estimation
+        for est in self.quantile_est:
+            est.fit(X.copy(), y)
+
+        return self
+
+    def _predict(self, X):
+        """Predict labels for data from features.
+
+        State required:
+            Requires state to be "fitted" = self.is_fitted=True
+
+        Accesses in self:
+            Fitted model attributes ending in "_"
+
+        Parameters
+        ----------
+        X : pandas DataFrame, must have same columns as X in `fit`
+            data to predict labels for
+
+        Returns
+        -------
+        y : pandas DataFrame, same length as `X`, same columns as `y` in `fit`
+            labels predicted for `X`
+        """
+
+        index = X.index
+        y_cols = self._y_cols
+        columns = y_cols
+
+        # quantile prediction
+        for est in self.quantile_est:
+            yhat = est.predict(X.copy())
+            self.quantile_values.append(yhat)
+
+        # generate j-qpd on each sample
+        mean, var = [], []
+        n_samples = len(X)
+
+        # NOTE: cdf -> pdf calculation because j-qpd's pdf calculation is bit complex
+        def exp_func(x):
+            pdf = derivative(j_qpd_s.cdf, x, dx=1e-6)
+            return x * pdf
+
+        def var_func(x, mu):
+            pdf = derivative(j_qpd_s.cdf, x, dx=1e-6)
+            return ((x - mu) ** 2) * pdf
+
+        for i in range(n_samples):
+            try:
+                j_qpd_s = J_QPD_S(
+                    0.2,
+                    self.quantile_values[0][i],
+                    self.quantile_values[1][i],
+                    self.quantile_values[2][i],
+                )
+                # TODO: scipy.integrate will be removed in scipy 1.12.0
+                # NOTE: integral range (-inf to inf) return NaN, it should check
+                loc, _ = quad(exp_func, a=0.0, b=np.inf)
+                mean.append(loc)
+                scale, _ = quad(var_func, a=0.0, b=np.inf, args=(loc))
+                var.append(scale)
+            except ValueError:
+                mean.append(np.nan)
+                var.append(np.nan)
+                continue
+
+        # fill nan by mean
+        mean_arr = np.asarray(mean)
+        mean_arr[np.isnan(mean_arr)] = np.nanmean(mean_arr)
+        var_arr = np.asarray(var)
+        var_arr[np.isnan(var_arr)] = np.nanmean(var_arr)
+        self.mean_jqpd = mean_arr.reshape(-1, len(y_cols))
+        self.var_jqpd = var_arr.reshape(-1, len(y_cols))
+
+        y_pred = pd.DataFrame(mean, index=index, columns=columns)
+        return y_pred
+
+    def _predict_proba(self, X):
+        """Predict distribution over labels for data from features.
+
+        State required:
+            Requires state to be "fitted".
+
+        Accesses in self:
+            Fitted model attributes ending in "_"
+
+        Parameters
+        ----------
+        X : pandas DataFrame, must have same columns as X in `fit`
+            data to predict labels for
+
+        Returns
+        -------
+        y_pred : skpro BaseDistribution, same length as `X`
+            labels predicted for `X`
+        """
+
+        # mean, variance calculation from j-qpd
+        _ = self._predict(X)
+
+        if self.distr_type == "Normal":
+            from skpro.distributions.normal import Normal
+
+            distr_type = Normal
+            distr_loc_scale_name = ("mu", "sigma")
+            self.var_jqpd = np.sqrt(self.var_jqpd)
+        # TODO: add other distributions
+        # elif distr_type == "Laplace":
+        #     from skpro.distributions.laplace import Laplace
+
+        #     distr_type = Laplace
+        #     distr_loc_scale_name = ("mu", "scale")
+        # elif distr_type in ["Cauchy", "t"]:
+        #     from skpro.distributions.t import TDistribution
+
+        #     distr_type = TDistribution
+        #     distr_loc_scale_name = ("mu", "sigma")
+        else:
+            raise NotImplementedError(f"{self.distr_type} is not support")
+
+        params = {}
+        # row/column index
+        ix = {"index": X.index, "columns": self._y_cols}
+        params.update(ix)
+        # location and scale
+        loc_scale = {
+            distr_loc_scale_name[0]: self.mean_jqpd,
+            distr_loc_scale_name[1]: self.var_jqpd,
+        }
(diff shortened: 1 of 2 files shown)


fkiraly:
looks good!

Did this get closed by accident?

setoguchi-naoki:
@fkiraly 
Hi, Franz, I'm one of the contributor of Cyclic boosting. I closed the PR manually so that I wanted to get review from another contributor of Cyclic boosting on my forked repository because my code is a draft.

fkiraly:
> I closed the PR manually so that I wanted to get review from another contributor of Cyclic boosting on my forked repository because my code is a draft.

My opinion, it's best to have reviews here - general best practice recommendation to have all discussion in the "host repo" and not in the fork, as the host repo is indexed by various tools, while fork repos are in general not (and can also disappear more easily).

It's marked as draft too, so no danger of getting it merged.

Also feel free to ask any of the skpro contribs for review!

setoguchi-naoki:
@fkiraly 
Thank you for giving me your opnion and follow your comment. I have finished to implement for interface of Cyclic boosting, so Would you review the code for me? I concern weather my implementation fit skpro's manner or not.

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,315 @@
...
+        self._y_cols = y.columns
+        y = y.to_numpy().flatten()
+
+        # multiple quantile regression for full probability estimation
+        for est in self.quantile_est:
+            est.fit(X.copy(), y)
+
+        return self
+
+    def _predict(self, X):
```
FelixWick, line 161: I would suggest to not run the full quantile regression - QPD thing here, but simply use a mean prediction mode (e.g., pipeline_CBPoissonRegressor for a multiplicative model on range 0, inf).

FelixWick: But that would require two different fit modes (as we want to use quantile regression with subsequent QPD for predict_proba), from which the user would have to select via a parameter in the init of the estimator. So, I can understand your rationale of using the mean of QPD. If we wanted to keep a single fit mode, we might consider using just the median (anyway predicted by the middle quantile regression model for QPD) though (avoiding the complexity via the mean of QPD).

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,315 @@
...
+        index = X.index
+        y_cols = self._y_cols
+
+        # quantile prediction
+        for est in self.quantile_est:
+            yhat = est.predict(X.copy())
+            self.quantile_values.append(yhat)
+
+        j_qpd = J_QPD_S(
+            0.2,
```
FelixWick, line 190: I would suggest to make this a parameter of the estimator, 0.2 is an arbitrary choice.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
...
+from scipy.stats import norm, logistic
+from scipy.misc import derivative
+from scipy.integrate import quad
+
+from typing import Optional
+
+from skpro.distributions.base import BaseDistribution
+
+
+class J_QPD_S(BaseDistribution):
```
FelixWick, line 19: We might also add the bounded version J_QPD_B.

skpro/regression/cyclic_boosting.py
FelixWick, line 190: This should be self.quantiles[0], as it needs to fit the lower quantile in the estimator (which is a parameter above).

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
...
+            # NOTE: integral interval should be checked, -inf to inf will be NaN
+            l, _ = quad(var_func, args=(mu, theta, kappa, delta, n), a=0.0, b=np.inf)
+            var.append(l)
+
+        var_arr = np.array(var)
+        return pd.DataFrame(var_arr, index=self.index, columns=self.columns)
+
+    def pdf(self, x):
+        # cdf -> pdf calculation because j-qpd's pdf calculation is bit complex
+        def cdf_func(x, theta, kappa, delta, n):
```
FelixWick, line 198: Call the function below to avoid code duplication here.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
...
+            cdf_arr = self.phi.cdf(
```
FelixWick, line 199: Try to avoid the code duplication here.

FelixWick: And in other places :).

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,315 @@
...
+        y_pred : skpro BaseDistribution, same length as `X`
+            labels predicted for `X`
+        """
+
+        # predict j-qpd
+        _ = self._predict(X)
+
+        if self.distr_type == "QPD":
+            return self.qpd
+        else:
```
FelixWick, line 227: I don't think we should offer this option, matching the QPD mean and variance on Gauss or Laplace distributions is odd at best (and can be severely off).

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,315 @@
...
+            qv_high=self.quantile_values[2],
+            index=index,
+            columns=y_cols,
+        )
+        self.qpd = j_qpd
+
+        y_pred = pd.DataFrame(j_qpd.mean(), index=index, columns=y_cols)
+        return y_pred
+
+    def _predict_proba(self, X):
```
FelixWick, line 202: I would suggest to also use predict_quantiles for the quantile predictions. And here we can then just give the QPD.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
...
+        self.theta[neg] = qv_high[neg] - l
+
+        B_L = self.B - self.L
+        H_B = self.H - self.B
+        _min = np.where(B_L < H_B, B_L, H_B)
+
+        self.delta = 1.0 / self.c * sinh(arccosh((self.H - self.L) / (2 * _min)))
+        self.kappa = 1.0 / (self.delta * self.c) * _min
+
+        # DataFrame
```
FelixWick, line 123: I don't get this index thing here.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
...
+                / delta
+                * sinh(
+                    arcsinh(1.0 / kappa * log((x - self.l) / theta))
+                    - arcsinh(n * self.c * delta)
+                )
+            )
+            return cdf_arr
+
+        pdf = []
+        for i in x.index:
```
FelixWick, line 210: Aren't these values all the same?

fkiraly:
I'm noticing the build failure in readthedocs, `exception: no module named skpro.regression.cyclic_boosting`

This usually indicates some mixup with imports, although I was unable to find any error in a couple minutes searching.

skpro/distributions/qpd.py
fkiraly: yes, this does not look right to me. The index etc usually only "lives" in the distribution itself, not in the parameters.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,245 @@
+"""J-QPD probability distribution."""
```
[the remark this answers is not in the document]
fkiraly: hm, only public functions, though.

skpro/regression/cyclic_boosting.py
setoguchi-naoki: @FelixWick 
How about Having two kind of estimators in Cyclic boosting class? I think it look seemless from user because keeping the single fit mode. Is it bit complicated?

FelixWick: Yes, that is what I meant. But at fit time, the user needs to decide what estimator to train. It would be a waste of runtime to train both always. (That is also true for the quantile regression of three different quantiles for the QPD, if all you want is the median.)

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,315 @@
+"""Cyclic boosting regressors.
+This is a interface for Cyclic boosting, it contains efficient,
+off-the-shelf, general-purpose supervised machine learning methods
+for both regression and classification tasks.
+Please read the official document for its detail
+https://cyclic-boosting.readthedocs.io/en/latest/
+"""
+
+"""License.
+EPL 2.0
```
fkiraly, line 10: The interfaces in sktime/skpro should be released under the BSD-3 license of the repo.
Of course the `cyclic_boosting` package can be subject to a different license.

I would think this should be ok, as there is no substantial logic, only interface logic, in the `skpro` module? Comments appreciated, especially if you think otherwise.

Btw, the docstring of the class is what this appears as in the doc, so I would recommend to give more credit to the upstream package, see for instance the `mapie` regressor.

FelixWick: I agree. As I also commented before, we should not copy code from Cyclic Boosting here, but just call the respective functionality.

skpro/regression/cyclic_boosting.py
fkiraly: is it perhaps helpful to look at the QR model by @Ram0nB here?
https://github.com/sktime/skpro/blob/main/skpro/regression/multiquantile.py

skpro/regression/cyclic_boosting.py
fkiraly: up to you

skpro/regression/cyclic_boosting.py
fkiraly: hm, that's an interesting question. `predict_quantiles` will automatically default to the `quantiles` method of the `predict_proba` return, that ensures consistency. The interface does not enforce exact consistency, so if there's two consistent options, a method/strategy parameter in the `__init__` may be the way to go.

skpro/regression/cyclic_boosting.py
setoguchi-naoki: @FelixWick 
Does alpha always mention that it should be lower quantile, right? the alpha is fixed as 0.2 (20 percentile) in the current code
`self.quantiles = [0.2, 0.5, 0.8]`
and
`j_qpd = QPD_S(
            alpha=0.2,
            qv_low=self.quantile_values[0],
            qv_median=self.quantile_values[1],
            qv_high=self.quantile_values[2],
            index=index,
            columns=y_cols,
        )`

Do you mean the change like this? (and add alpha as a class's parameter)
`self.quantiles = [alpha, 0.5, 1 - alpha]`
and
`j_qpd = QPD_S(
            alpha=alpha,
            qv_low=self.quantile_values[0],
            qv_median=self.quantile_values[1],
            qv_high=self.quantile_values[2],
            index=index,
            columns=y_cols,
        )`

FelixWick: No, no need to introduce a new parameter, you can simply use self.quantiles[0], this corresponds to the alpha in QPD class.

skpro/regression/cyclic_boosting.py
FelixWick: Ok, I see. In that case, let's stay consistent. Let me summarize my suggestion in the other comment thread :).

skpro/regression/cyclic_boosting.py
FelixWick: Ok, I changed my mind. For simplicity, what do you think of the following structure:
- always fit three quantiles: alpha, 0.5 (median), and 1-alpha (alpha to be defined as parameter instead of quantiles triplet). That means just use quantile regression as single estimator. But we should use either multiplicative or additive mode, depending on the target range.
- in predict_proba, give QPD as distribution.
- in predict, give the median as point estimate.
- in predict_quantiles, give the three quantile estimates (alpha, median, 1-alpha), or, if that's not possible from the interface, just the alpha estimate.

setoguchi-naoki: @FelixWick 
I followed your comments. Additionally, predict_interval method was implemented. In predict_quantiles, if alphas are same with the quantiles that the estimator train, the method returns estimated value directlly,  if it is not same, the method returns specified quantile value from QPD.

skpro/distributions/qpd.py
setoguchi-naoki: I have avoided some duplication and relocated

skpro/distributions/qpd.py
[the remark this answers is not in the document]
setoguchi-naoki: removed copying code and import J_QPD_S, J_QPD_B from cyclic_boosting package

skpro/regression/cyclic_boosting.py
setoguchi-naoki: I removed these options

skpro/distributions/qpd.py
setoguchi-naoki: X is pandas Datafreme which include a point of distribution as value and corresponded QPD's index in my code. So,  these have different value and pdf method returns specific pdf value. However, I'm not sure that my implementation is correct. Please point out my understanding or implementation's mistake if it is exist.

setoguchi-naoki:
@fkiraly 
> I'm noticing the build failure in readthedocs, exception: no module named skpro.regression.cyclic_boosting
This usually indicates some mixup with imports, although I was unable to find any error in a couple minutes searching.

I got this error message, I don't know why it is occured. Do you have any ideas?
Cyclic boosting's version is correct.
> ModuleNotFoundError: CyclicBoosting requires package 'cyclic_boosting>=1.2.1' to be present in the python environment, with version >=1.2.1, but incompatible version 1.0 was found

fkiraly:
hmmmm, @setoguchi-naoki, the typical source of the error message would be a combination fo two things:

* the lower bound `"cyclic-boosting>=1.2.1"` that you put in the `python_dependencies` tag
* `cyclic-boosting 1.0` being actually present in the python env.

Which is strange, because the `pyproject.toml` contains `cyclic-boosting>=1.2.1` as well.

Are you getting this on local tests? If so, your environment might be at fault.

fkiraly:
PS: the code-quality check is coming up with a few linting issues, let us know if you want advice on setting up local linting!

This usually runs to prevent syntax errors going through, but code formatting issues will also prevent remote CI from running.

setoguchi-naoki:
@fkiraly 
Sorry for being late. I got the Import error's reason, there is a little mistake in Cyclic boosting package. the problem will be solved soon. However, I got the other import error in part of docstring's test

> sphinx.ext.autosummary.ImportExceptionGroup: no module named skpro.distributions.Empirical

Could you please support for me?

setoguchi-naoki:
@FelixWick 
Thank you for your review, I implemented extended J_QPDs. The problem which I have is only the issue related to build docstring but I don't know how to solve it even though reading its error code. :(

FelixWick:
@setoguchi-naoki Maybe related to this? https://stackoverflow.com/questions/62226506/python-sphinx-autosummary-failed-to-import-module

setoguchi-naoki:
I tried to build docstring on my local machine. it run with some warnings but it looks successful to me
screen_shot 2023-12-26 141536

fkiraly:
Apologies for the delay, I was busy at work and with the `sktime` releases.

The docs failure is actually a genuine test failure, due to how `scikit-base` based packages like `skpro`, `sktime` handle soft dependencies.

In order to not require of every user to install all the soft dependencies, they must be isolated, that is, the package should install and not break with it is not present.

For this, imports of soft dependencies, like `cyclic_boosting`, need to reside inside classes or methods, not at the top of the file.

General instructions here:
https://www.sktime.net/en/latest/developer_guide/dependencies.html#handling-soft-dependencies

But I'll also point out what to do in the PR.

skpro/distributions/qpd.py
```diff
@@ -0,0 +1,705 @@
...
+    "FelixWick",
+    "setoguchi-naoki",
+]  # interface only. Cyclic boosting authors in cyclic_boosting package
+
+import warnings
+from typing import Optional
+
+import numpy as np
+import pandas as pd
+from cyclic_boosting.quantile_matching import (
```
fkiraly, line 14: these imports should move into the distribution classes.

My recommendation is to move them inside the methods where they are needed.

skpro/regression/cyclic_boosting.py
```diff
@@ -0,0 +1,428 @@
...
+
+__author__ = [
+    "setoguchi-naoki"
+]  # interface only. Cyclic boosting authors in cyclic_boosting package
+
+import warnings
+
+import numpy as np
+import pandas as pd
+from cyclic_boosting import (
```
fkiraly, line 19: similar, these imports should move into the class methods

fkiraly (changes requested):
Nice!

Apologies for the hiatus, that was work related (and I had to finish the `sktime` releases).

Looks good to go, I see only an issue with dependency isolation, as mentioned above. That should also solve the docs issue.

fkiraly:
@setoguchi-naoki, @FelixWick, I proceeded pushing some commits which isolate the dependencies - I hope this is ok, if not, please revert.

What I did:

* move imports into classes and methods
* skip the class doctest, move to a pytest test that is only executed when the required dependencies are present

skpro/distributions/qpd.py
fkiraly: done in my commits

skpro/regression/cyclic_boosting.py
fkiraly: done in my commits

fkiraly (changes requested):
The failures now are genuine.

The reason seems to be, `CyclicBoosting` does not construct with all the parameter sets returned by `get_test_params`.

In addition, the docstring examples of the distributions should be isolated, similar to the `CyclicBoosting` one (I skipped it and moved to conditional pytest, in `test_cyclic_boosting`)

(I will stop here with contributions, as most remaining issues touch logic which you are more familiar with)

fkiraly:
@setoguchi-naoki, @FelixWick, let us know if you need a hand!

setoguchi-naoki:
@fkiraly 
Thank you for your cooperation! I solved some bug and add a new test script for qpd.py. then, all check have passed. please check them.

fkiraly (changes requested):
Nice! Looks almost done, failures are two kinds:

* the `cyclic_boosting` package appears not to be python 3.12 compatible, so the `python_version` tag should be set to `<3.12`
* at one point, the inner logic seems to be looking for a column `s1` in `X` that does not exist - seems like a column name treatment bug.

fkiraly (approved):
nice - looks all good now!

You should also add yourself to the `.all-cotributorsrc` file!

fkiraly:
(tests have not run through yet, but I've started them again)

setoguchi-naoki:
@fkiraly 
I have found why that error occued. Test dataset's columns are updated at line 156 in scenarios_regressor_proba.py.
`X.columns = ["age", "sex", "bmi", "bp", 1, 2, 3, 4, 5, 6]` 
It needs to same feature_propety's key names and dataset columns for Cyclic boositing but the former is fixed as a test paramter now. How can I change dynamically test parameters along with changing scenario?

fkiraly (changes requested):
Oh, I see!
Let me rephrase if I understood correctly:

`CyclicBoosting` has a parameter, `feature_properties`, in which keys need to be the same as (all) variable names. That causes a failure in the test, because the same parameter set is tested with all data scenarios, where there are different sets of variable names.

This is an interesting question - optimally, the user does not have to worry about variable specific parameters.

May I suggest to auto-detect the types? A simple solution could be, use `pandas` [`is_float_dtype`](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_float_dtype.html) on `X.dtype`. For each column where this is true, assume continuous, otherewise categorical. I would suggest to do that if `feature_properties` is `None`. This would work as follows - you have an internal `_feature_properties` (with underscore at the start), and if `feature_properties` is given, it is that, and if not, it is the auto-detected one.

That would solve the issue with the tests, and imo would make the algorithm much more convenient for the user too.

Further requests relate to the docstring of `CyclicBoosting`:

* it should give credit to the package, e.g., say sth like "direct interface to `cyclic_boosting classes etcetc`". I would do that in the second line.
* `feature_properties` should be listed in the docstring, or it might be confusing to a user. Or is the specification very long?
* All parameters with a default should have that default stated in the docstring.

FelixWick:
In fact, the user doesn't need to give feature properties, there is a default mechanism (which, admittedly, still leaves some room for improvement ;)). So, your suggestion should work out-of-the-box, I think.

fkiraly:
Nice!
Still one would have to set a default, right now the user is forced to specify.

I would then advice to remove the variable specific parameters from `get_test_params`, and add separate `pyteset` based tests to ensure correctness when they are passed.

FelixWick:
Let me know if you need my help, @setoguchi-naoki

setoguchi-naoki:
@FelixWick @fkiraly 
Thank you guys for your suggestion! I will make simple feature analysis function and add test for it. In fact, I am already developing automatic feature type analysis function for Cyclic boosting package. It can detect but simple data property also more complicate ones (e.g. seasonality). However, I guess it need to have a little time to release, please allow me to take simple approach for now. I will reflect it for skpro after releasing this function.

setoguchi-naoki:
@fkiraly 
As Felix said, feature_properties (and feature_groups) always don't need to set up. So I just updated CyclicBoosting class's paramaters and removed specific test paramaters as a most simple solution. if feature_properties and feature_groups are None, these are set automatically in cyclic boosting model. I guess it should passed all test by this modification.

fkiraly (changes requested):
Excellent, thanks!

Could you perhaps add a test that uses the manual feature typing to `test_cyclic_boosting`, so we have that covered?

Otherwise, I have no additional requests, very nice.

Once you think you are ready, pleas switch this PR to "ready for review".

setoguchi-naoki:
I have finished :)

fkiraly (approved):
Nice!
All green for merge if tests pass.

fkiraly (changes requested):
Fails with `test_cyclic_boosting_with_manual_paramaters - ValueError: The SPT values need to be monotonically increasing.` - error seems genuine.

FelixWick:
The quantile predictions need to be ordered correctly for the QPD to make sense. And as we are dealing with statistical models, it can happen that this is not the case, especially in low-statistics tests. So, I would suggest to just tune the respective test ;).

fkiraly:
that, or adding a coercion?

FelixWick:
Yeah, I have implemented a potential coercion in QPD_RegressorChain in Cyclic Boosting. But a simpler approach would do as well here, I think.

setoguchi-naoki:
I modified SPT's value checking from replacement by mean to sorting.

fkiraly:
Seems lik there are sporadic timeouts of `test_cyclic_boosting_with_manual_paramaters` - a single test run is limited to 10min or it will fail with the assumption that the long runtime is a consequence of a failure.

setoguchi-naoki:
All test have passed :)

fkiraly (approved):
🎉 🎆 🎉 🎆 🎉

fkiraly:
Minor question, non-blockinng, should the QPD not go in the "non-parametric" section of the distirbutions?

They are called "quatile parameterized" but afaik are typically considered non-parametric since they are explicit in terms of quantiles (which are interpolated) rather than intrinsic simple parameters.

fkiraly:
Just looking through open convos at the moment, and I found this by @FelixWick:

> But as I wrote below, I think we should rename the parameter alpha in predict_quantiles, because it is confusing.

Apologies, I overlooked this - are you suggesting that the interface be changed in general for all estimators, with a rename of `alpha`? There might be a case for that, though I cannot find the arguments you are referencing.

FelixWick:
> Minor question, non-blocking, should the QPD not go in the "non-parametric" section of the distributions?
> 
> 
> 
> They are called "quantile parameterized" but afaik are typically considered non-parametric since they are explicit in terms of quantiles (which are interpolated) rather than intrinsic simple parameters.

Yes, that sounds reasonable.

FelixWick:
> Just looking through open convos at the moment, and I found this by @FelixWick:
> 
> 
> 
> > But as I wrote below, I think we should rename the parameter alpha in predict_quantiles, because it is confusing.
> 
> 
> 
> Apologies, I overlooked this - are you suggesting that the interface be changed in general for all estimators, with a rename of `alpha`? There might be a case for that, though I cannot find the arguments you are referencing.

No, that was not a general comment, just for Naoki's implementation. But I think he fixed it. Right, @setoguchi-naoki ?

fkiraly:
>> Minor question, non-blocking, should the QPD not go in the "non-parametric" section of the distributions?

> Yes, that sounds reasonable.

Ok, moving to the non-parametric taxon then.

setoguchi-naoki:
> > Just looking through open convos at the moment, and I found this by @FelixWick:
> > > But as I wrote below, I think we should rename the parameter alpha in predict_quantiles, because it is confusing.
> > 
> > 
> > Apologies, I overlooked this - are you suggesting that the interface be changed in general for all estimators, with a rename of `alpha`? There might be a case for that, though I cannot find the arguments you are referencing.
> 
> No, that was not a general comment, just for Naoki's implementation. But I think he fixed it. Right, @setoguchi-naoki ?

Yes, it is only for my predict_quantiles implementation now.

PR #189, opened 2024-01-29
[BUG] fix lookup for specialized test classes

fkiraly:
This PR fixes lookup for the specialized test classes.

The internal `all_objects` call was passing classes as `object_type` to the `skpro` registry lookup, which it did not recognize - it should be the type string.

As a consequence, the list of classes was coming up empty - instead of containing classes of a given scitype - hence test coverage was reduced since initially when the `skpro` `all_objects` utility was upgraded (which is not long ago but long enough for uncovered failures to creep in). The generic test classes `TestAllEstimators`, `TestAllObjects` were not impacted.

Depends on fixes for silent test failures:

* https://github.com/sktime/skpro/pull/191
* https://github.com/sktime/skpro/pull/192
* temporary skip https://github.com/sktime/skpro/pull/193


fkiraly:
closed by accident?

PR #194, opened 2024-01-30
[BUG] bugfixes to QPD distributions - `QPD_U`, `QPD_S`

### LINKED ISSUE #190: [BUG] QPD distribution and `CyclicBoosting` API non-compliance
@setoguchi-naoki, @FelixWick, I have to apologize in advance for this.

Due to an error in the test logic introduced in an update to class retrieval (see https://github.com/sktime/skpro/pull/189), probability distributions went uncovered for most of your PR's duration.

This is my fault for not noticing, and breaking it in the first place.

Bad news is that the QPD distribution have a few non-compliances:

* `cdf` method
* `mean`, `var`
* subsetting

`CyclicBoosting` as well:

* `_predict_quantiles` arg should be `alpha`, not `quantiles`

Good news is that this has not been released yet, and I was running more tests before the 2.2.0 release, which is what ultimately caught the problem.

I'll simply wait with 2.2.0 until we had time to fix this - happy to help.

For testing locally, you need to depend on the branch #189 until it is merged.

fkiraly:
Re-enables tests for `QPD_U`, `QPD_S` skipped in sktime/skpro#193 and attempts to fix failures in QPD distributions.

it appears that #190 failures are resolved - fixes #190.

Further `QPD_B` seems to fail at mean and var calculations, this is a separate bug and raised as https://github.com/sktime/skpro/issues/257. As a mechanical consequence, `QPD_B` is left skipped in the tests.


fkiraly:
@setoguchi-naoki, @felixwick, I have been trying to fix the open bug after re-enabling the tests, it seems like there are some issues with the internal indexing.

I'm not sure what format of `alpha`, `qc_low` etc the interna code assumes, could you help me? I was assuming, flat numpy array (1D), but that does not seem to work.

FelixWick:
alpha: float,
qv_low: Union[float, np.ndarray],
qv_median: Union[float, np.ndarray],
qv_high: Union[float, np.ndarray],

fkiraly:
I mean, internally. You do this: `params[idx] = np.array([p])`, so `alpha` becomes a 1D array, no?

fkiraly:
Thanks, this seems to have helped.

Now, `QPD_B` is failing internally, it is substituting 0 or 1 into a logit, which produces nans.
Any ideas where or why, @FelixWick, @setoguchi-naoki?

It should not substitute the bounds, because they will always be nan, so where is that coming from?

fkiraly:
FYI, it seems the bugs we were targeting are now resolved.

I have opened a bug issue here, to track that separately: https://github.com/sktime/skpro/issues/257

PR #215, opened 2024-03-19
[ENH] interface to `ngboost`

### LINKED ISSUE #135: [ENH] interface probabilistic regressors from `ngboost` package
Update by @fkiraly  - we should interface `ngboost` regressors as probabilistic supervised (tabular) regressors.
As discussed in the below, this should be an `skpro` regressor, which in turn can be used in `sktime` reduction forecasters (such as `YfromX`). The original request was for using `ngboost` as a forecaster in `sktime`, but this is a tabular proba regressor that needs to go through an additional reduction step (which is now implemented in `sktime`).

It should be straightforward to interface `ngboost` using the probabilistic regressor extension template:
https://github.com/sktime/skpro/blob/main/extension_templates/regression.py
so adding it as a good first issue.

The main techincal concern might be translating the `ngboost` probability distributions into `skpro` probability distributions, but that should also be addressable with a lean adaptation layer (personally, I would add that adaptation in an adapter utility subpackage in `regression.ngboost`).

Original request below.

---

Can we build a Probabilistic Forecaster using Ngboost. A Probabilistic regression like Ngboost method will give us the confidence intervals out of the box unlike others where we need a heuristic(like quantile loss) to get the values of Conf/Pred Ints.

A rough sketch:
```
class ProbForecaster:
     def __init__():
           self.estimator_ = NgboostRegressor()
     def fit(y,X=None):
           y = make_reduction(y)
           self.estimator_.fit(y)
     def predict(fh,X=none,alpha=0.95,conf_int=True):
           mean_preds =  self.estimator_.predict(y)
           distribution =  self.estimator_.pred_dist(y)
           conf_ints = scipy.norm.interval(alpha, distribution.loc, distribution.scale )
           return mean_preds,conf_ints
```

Using Xgboost/LightGBM with quantile loss makes the calculation of prediction intervals inefficient.

**Additional context**
Ngboost Doc- [https://stanfordmlgroup.github.io/projects/ngboost/](https://stanfordmlgroup.github.io/projects/ngboost/)

ShreeshaM07:
#### Reference Issues/PRs

Resolves #135 

This PR has added a way to use `ngboost regressor` in `skpro`.
Tested it using the following piece of code
```python
import pandas as pd
import numpy as np
from sklearn.metrics import mean_squared_error

#Load Boston housing dataset
data_url = "http://lib.stat.cmu.edu/datasets/boston"
raw_df = pd.read_csv(data_url, sep="\s+", skiprows=22, header=None)
X = np.hstack([raw_df.values[::2, :], raw_df.values[1::2, :2]])
Y = raw_df.values[1::2, 2]

X_train, X_test, Y_train, Y_test = train_test_split(X, Y, test_size=0.2)

ngb = NGBoostRegressor()._fit(X_train, Y_train)
Y_preds = ngb._predict(X_test)
Y_dists = ngb._pred_dist(X_test)

# test Mean Squared Error
test_MSE = mean_squared_error(Y_preds, Y_test)
print('Test MSE', test_MSE)

# test Negative Log Likelihood
test_NLL = -Y_dists.logpdf(Y_test).mean()
print('Test NLL', test_NLL)
```

`NGBRegressor` has been added from the `ngboost` package. 

What modifications need to be done and some help on how to add `get_test_params` function and the parameters to be sent via `FITTED_PARAMS_TO_FORWARD`.

No

##### For all contributions

##### For new estimators


skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,167 @@
...
+            validation_fraction=self.validation_fraction,
+            early_stopping_rounds=self.early_stopping_rounds,
+        )
+        self.ngb.fit(X, y)
+        return self
+
+    def _predict(self, X):
+        return self.ngb.predict(X)
+
+    def _pred_dist(self, X):
```
fkiraly, line 140: good start! This is, of course, where the crux of the matter will sit.

In `_predict_proba`, you'll need to convert an `ngboost` distribution int equivalent `skpro` distributions. I recommend using adaptor/converter patterns.

ShreeshaM07: Sorry was a bit busy last couple of days. By converting do you mean like using something like `SklearnProbaReg` adapter from `_sklearn_proba` but for `ngboost` I'd have to make one. Do I understand correctly?

fkiraly: > Do I understand correctly?

Yes - I would do it entirely on the level of distributions first, i.e., transforming the `ngboost` return type into an `skpro` `BaseDistribution`.

My guess would be, you'll not need an extra adapter on the level of regressor, if you call the conversion from `_predict_proba`.

ShreeshaM07:
I have implemented the `_predict_proba` after converting the return type of `ngboost`'s `normal` distribution to `skpro` `Normal` distribution it must now be extended to other distributions that `ngboost` supports and supported by `skpro`. I will also work on implementing `predict_interval` and `predict_quantile` next. In the meantime I will also try to make the conversion of the return types into a function.

The current output upon calling 
Y_pred_proba = ngb._predict_proba(X_test)
print(Y_pred_proba) is 
```
[iter 0] loss=3.6479 val_loss=0.0000 scale=1.0000 norm=6.7178
[iter 100] loss=2.6977 val_loss=0.0000 scale=2.0000 norm=4.9421
[iter 200] loss=2.1456 val_loss=0.0000 scale=2.0000 norm=3.3556
[iter 300] loss=1.8761 val_loss=0.0000 scale=1.0000 norm=1.4400
[iter 400] loss=1.7477 val_loss=0.0000 scale=1.0000 norm=1.3337
<ngboost.distns.normal.Normal object at 0x736de3d40f50>
Normal(columns=RangeIndex(start=0, stop=1, step=1),
       index=RangeIndex(start=0, stop=102, step=1),
       mu=             0
0    16.962767
1    21.572000
2    16.014035
3    30.217625
4    19.818455
..         ...
97   11.209679
98   25.100160
99    7.723768
100  19.720101
101  24.292480

[102 rows x 1 columns],
       sigma=            0
0    1.206229
1    1.079737
2    1.241355
3    1.223454
4    1.187693
..        ...
97   1.292326
98   1.216184
99   1.304442
100  1.106799
101  1.176687

[102 rows x 1 columns])
Test MSE 16.048500328511995
Test NLL 4.412747994731287
```

Please let me know if it looks good or something needs to be changed.

fkiraly (changes requested):
Thanks a lot!

Some comments:
* please look at the extension template, `extension_templates/regression.py` - you should not override `predict_interval` etc, there is no need to copy-paste everything from the base class.
    * here is a longer guide on how-to: https://www.sktime.net/en/latest/developer_guide/add_estimators.html
* please implement `_predict_proba` only, this should start with an underscore.
* in it, it is sufficient to start at the line `# Convert NGBoost distibution to skpro BaseDistribution` and supply the code below. That is all you need to do. No need to do copy-paste anything from the base class.
    * of course, for all distributions that `ngboost` could return. I.e., you need to check which distribution is being returned, and convert to the same `skpro` distribution. All `ngboost` distributions should have a counterpart in `skpro`.

fkiraly:
Also, you should add `ngboost` as a soft dependency to `pyproject.toml`, and via the `python_dependeices` tag to the estimator.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,584 @@
+"""Adapters to ngboost regressors with probabilistic components."""
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["ShreeshaM07"]
+
+import numpy as np
+import pandas as pd
+from ngboost import NGBRegressor
+from ngboost.distns import Normal
```
fkiraly, line 9: imports should happen only in the estimator

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,584 @@
...
+
+
+    Returns
+    -------
+        An NGBRegressor object that can be fit.
+    """
+
+    def __init__(
+        self,
+        Dist=Normal,
```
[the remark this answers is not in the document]
ShreeshaM07: Could you please guide me as to which `sklearn` default learner do I use instead of ngboost `default_tree_ learner`.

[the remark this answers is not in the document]
fkiraly: that's an `sklearn` learner - I would advise to hard-code it directly in `__init__`, i.e., `if estimator is None: self.estimator_ = etc`, with the precise learner spec found here: https://github.com/stanfordmlgroup/ngboost/blob/7da1ca94a8857b865cf39fca71b2564bad0c3e15/ngboost/learners.py

(I would also advise to rename the arg `base` to `estimator`, to be in line with `sklearn` here)

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
     "scikit-base>=0.6.1,<0.8.0",
     "scikit-learn>=0.24.0,<1.5.0",
     "scipy<2.0.0,>=1.2.0",
+    "ngboost",
```
fkiraly, line 50: not to the core dependencies please, to the optional dependencies

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+            Set to None to disable early stopping and validation
+            None enables running over the full data set.
+
+
+    Returns
+    -------
+        An NGBRegressor object that can be fit.
+    """
+
+    _tags = {
```
fkiraly, line 56: this should include other tags as well: `"authors"` (you) and `"maintainers"`

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+        It then creates an object of that particular NGBoost Distribution.
+
+        Returns
+        -------
+        NGBoost Distribution object.
+        """
+        from ngboost.distns import Laplace, LogNormal, Normal, Poisson, T
+
+        dist_ngboost = Normal
+        if self.dist == "Normal":
```
fkiraly, line 118: yes, very nice, that's what I meant.

I would recommend to use a `dict` here, that makes the logic easier to maintain and extend.
Let me know if you want pointers how ot concretely use a `dict`.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+        ----------
+        pred_mean, pred_std and index and columns.
+
+        Returns
+        -------
+        skpro_dist (skpro.distributions.BaseDistribution):
+        Converted skpro distribution object.
+        """
+        skpro_dist = None
+        if self.dist == "Normal":
```
fkiraly, line 193: as above, using a `dict` would help.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+
+        Returns
+        -------
+        y : skpro BaseDistribution, same length as `X`
+            labels predicted for `X`
+        """
+        X = self._check_X(X)
+
+        # Convert NGBoost distribution to skpro BaseDistribution
+        pred_mean = self._predict(X=X)
```
fkiraly, line 250: The general idea makes sense, though I would do it differently:

1. just call `pred_dist` from `ngboost`. This is a full parameteric distirbution.
2. in the `_ngb_dist_to_skpro`, extract the parameters from the `ngboost` distribution and plug them into the `skpro` distribution.

No need to take detours via the mean or `_predict`, especially since some of the distributions are not directly parameterized in terms of the mean.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+            tol=self.tol,
+            random_state=self.random_state,
+            validation_fraction=self.validation_fraction,
+            early_stopping_rounds=self.early_stopping_rounds,
+        )
+        self.ngb.fit(X, y)
+        self._is_fitted = True
+        return self
+
+    def _predict(self, X):
```
fkiraly, line 173: I would recommend, remove `_predict` and rely on the default behaviour entirely.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,337 @@
...
+            col_sample=self.col_sample,
+            verbose=self.verbose,
+            verbose_eval=self.verbose_eval,
+            tol=self.tol,
+            random_state=self.random_state,
+            validation_fraction=self.validation_fraction,
+            early_stopping_rounds=self.early_stopping_rounds,
+        )
+        self.ngb.fit(X, y)
+        self._is_fitted = True
```
fkiraly, line 170: no need to set `_is_fitteed` to `True`, the boilerplate layer in `fit` already does this.

skpro/regression/gradient_boosting/_ngboost.py
ShreeshaM07: I have now taken care as to what the value actually returns from the `ngboost` and converted it to the respective value in `skpro`. It has also been explained using the comments as to what the `distribution` is and what the `parameters` indicate.

skpro/regression/gradient_boosting/_ngboost.py
ShreeshaM07: I think it would be useful to keep `_predict` in case for just predicting the values. I have not used this anymore in `_predict_proba`. Please do let me know if you think otherwise.

ShreeshaM07:
@fkiraly could you please review the updates I have made to the code and if anything else needs to be modified.

skpro/regression/gradient_boosting/_ngboost.py
fkiraly: hm, I suppose it makes sense - not a blocker from my side.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,367 @@
...
+        # Laplace           | loc = mean, scale = scale parameter
+
+        loc_scale_params = ["Normal", "Laplace", "TDistribution"]
+        mu_params = ["Poisson"]
+        s_scale = ["LogNormal"]
+        pred_mean = None
+        pred_std = None
+        # Convert NGBoost distribution to skpro BaseDistribution
+        if self.dist in loc_scale_params:
+            pred_mean = self._pred_dist(X).params["loc"]
```
fkiraly, line 280: this is not a bug, but confusing naming - the two parameters passed are not always mean or std!

You do pass these into the distributions correctly though, so there is nothing to fix except the naming.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,367 @@
...
+            "Normal": Normal,
+            "Laplace": Laplace,
+            "TDistribution": TDistribution,
+            "Poisson": Poisson,
+            "LogNormal": LogNormal,
+        }
+
+        skpro_dist = None
+
+        mu_sigma = ["Normal", "TDistribution", "LogNormal"]
```
fkiraly, line 219: you do this translation to and from pred_mean etc twice, once from and once to.

It would reduce unnecessary complexity of the code if, instead, you would have the parameter names per distribution in a dict, pass the params to this method, and use `**kwargs` to pass it to `ngboost_dists[self.dist]`.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ all_extras = [
     "lifelines<0.29.0",
     "mapie",
     "matplotlib>=3.3.2",
-    "ngboost",
+    "ngboost<=1.5.1",
```
fkiraly, line 59: this should be "strictly smaller than `1.next.0`"

skpro/regression/gradient_boosting/_ngboost.py
ShreeshaM07: I have implemented it in this way now please let me know if it needs to be improved further.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,342 @@
...
+    """
+    Constructor for all NGBoost models.
+
+    This class implements the methods that are common to all NGBoost models.
+    Unless you are implementing a new kind of regression (e.g. interval-censored, etc.),
+    you should probably use one of NGBRegressor, NGBClassifier, or NGBSurvival.
+
+    Parameters
+    ----------
+        Dist              : assumed distributional form of Y|X=x.
```
fkiraly, line 22: this is not in numpydoc format and renders badly. Kindly follow examples in other regressors.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,342 @@
...
+                            A distribution from ngboost.distns, e.g. Normal
+        Score             : rule to compare probabilistic
+                            predictions P̂ to the observed data y.
+                            A score from ngboost.scores, e.g. LogScore
+        Base              : base learner to use in the boosting algorithm.
```
fkiraly, line 27: this should reflect the actual name of the param

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,342 @@
...
+        self.col_idxs = []
+        self.tol = tol
+        self.random_state = check_random_state(random_state)
+        self.best_val_loss_itr = None
+        self.validation_fraction = validation_fraction
+        self.early_stopping_rounds = early_stopping_rounds
+
+        super().__init__()
+
+    def dist_to_ngboost_instance(self):
```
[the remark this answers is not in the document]
fkiraly: I would also make `dist` an argument, to make it explicit what the input is. And not access `self` inside.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if self.dist in dist_params and self.dist in skpro_params:
+            ngboost_params = dist_params[self.dist]
+            skp_params = skpro_params[self.dist]
+            for ngboost_param, skp_param in zip(ngboost_params, skp_params):
+                kwargs[skp_param] = self._pred_dist(X).params[ngboost_param]
+                if self.dist == "LogNormal" and ngboost_param == "scale":
+                    kwargs[skp_param] = np.log(self._pred_dist(X).params[ngboost_param])
+
+                kwargs[skp_param] = self._check_y(y=kwargs[skp_param])
```
fkiraly, line 292: can you explain why we need this line? I would not have put it here, but you probably had a good reason to add it?

fkiraly (changes requested):
Nice, thanks! Really great addition!

I've started the CI tests to see whether they pass - I assume you have checked with `check_estimator` already.

Small change requests only:
* docstring formatting needs to be updated, see above. Also, all parameters should say whether they are optional, and list default value otherwise. Please add, similar to other regressors.
* docstring needs to be updated re dist and score, and list available options
* we should add the regressor to the API reference. This is `docs/source/api_reference/regression.rst`, I would add it directly after MAPIE (given its popularity)

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,342 @@
...
+            "LogScore": LogScore,
+        }
+        score = None
+        if self.score in ngboost_score:
+            score = ngboost_score[self.score]
+
+        self.ngb = NGBRegressor(
+            Dist=dist_ngboost,
+            Score=score,
+            Base=self.estimator,
```
fkiraly, line 174: the estimtaor needs to be cloned - otherwise it will be mutated, and you violate the `sklearn` extension contract (as the failing test tells you)

skpro/regression/gradient_boosting/_ngboost.py
ShreeshaM07: The output parameters returned from the ngboost distribution are `numpy` arrays so if I try to access its index and columns in `kwargs["mu"].index` and `kwargs["mu"].columns` it will give an
 ```
AttributeError: 'numpy.ndarray' object has no attribute 'index'
```
 So I am converting it into a pandas `DataFrame` which is also my return type as mentioned in the docstring for the function using `_check_y()`.

File: skpro/regression/gradient_boosting/_ngboost.py [hunk 19]
```diff
@@ -0,0 +1,342 @@
...
+
+        ngboost_dists = {
+            "Normal": Normal,
+            "Laplace": Laplace,
+            "TDistribution": TDistribution,
+            "Poisson": Poisson,
+            "LogNormal": LogNormal,
+        }
+
+        kwargs["index"] = kwargs["mu"].index
```
fkiraly, line 223: I think this should be from `X.index` (from `predict_proba`), and `y.index` (from `fit`), not the index that ngboost returns.

ShreeshaM07: Yes that makes sense. I have now made it `X.index` inplace of `kwargs["mu"].index` however I am still unable to remove the 
```
                kwargs[skp_param] = self._check_y(y=kwargs[skp_param])
                # returns a tuple so taking only first index of the tuple
                kwargs[skp_param] = kwargs[skp_param][0]
```
As it is causing a shape mismatch error upon removing it.
`ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (111,) and arg 3 with shape (10,)`.
Where 111 expected Index but arg3 would be the number of columns which seems to be 10.

ShreeshaM07:
> Nice, thanks! Really great addition!
> 
> I've started the CI tests to see whether they pass - I assume you have checked with `check_estimator` already.

Could you please help me with the check_estimator. Upon running 
```
from sklearn.utils.estimator_checks import check_estimator
check_estimator(NGBoostRegressor())
```

It is throwing a `TypeError: input must be a pd.DataFrame`.

skpro/regression/gradient_boosting/_ngboost.py
fkiraly: do you have an explanation why this happens?

skpro/regression/gradient_boosting/_ngboost.py
fkiraly: but do you still need the index, given that you removed it now?
The index (for `y_pred`) should be taken from `X` and `y`, instead.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,353 @@
...
+    """
+    Constructor for all NGBoost models.
+
+    This class implements the methods that are common to all NGBoost models.
+    Unless you are implementing a new kind of regression (e.g. interval-censored, etc.),
+    you should probably use one of NGBRegressor, NGBClassifier, or NGBSurvival.
+
+    Parameters
+    ----------
+        dist              : string , default = "Normal"
```
fkiraly, line 22: thanks for updating the docstring. Please check how the docs render in the readthedocs CI job.

Indentation matters! And this will render funnily, so kindly look at the other regressors and indent similarly.

You can check the result, how it looks, in the readthedocs CI job (click "details").

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,353 @@
...
+
+import numpy as np
+from sklearn.utils import check_random_state
+
+from skpro.regression.base import BaseProbaRegressor
+
+
+class NGBoostRegressor(BaseProbaRegressor):
+    """
+    Constructor for all NGBoost models.
```
fkiraly, line 14: Here you probably want to describe the regressor instead, i.e., it is an interface to `NGBRegressor`.

ShreeshaM07:
I have resolved the shape mismatch error using  `ravel()` which the `check_estimator` suggested to apply on `y`. I have also changed the `kwargs["columns"]` to be `y.columns` now . 

But some tests are failing in the `check_estimator` although it is running it without any errors.
```
FAILED: test_constructor[NGBoostRegressor]
FAILED: test_set_params_sklearn[NGBoostRegressor]
FAILED: test_fit_does_not_overwrite_hyper_params[NGBoostRegressor-ProbaRegressorBasic]
FAILED: test_fit_does_not_overwrite_hyper_params[NGBoostRegressor-ProbaRegressorXcolMixIxYnp]
FAILED: test_fit_does_not_overwrite_hyper_params[NGBoostRegressor-ProbaRegressorSurvival]
FAILED: test_input_output_contract[NGBoostRegressor]
```
Could you please help out as to why these are failing.

fkiraly:
> Could you please help out as to why these are failing.

Of course, will look at it later today.

Can you also describe what you have found out in debugging these failures? You can use `raise_exceptions=True` to get an error raise for debugging.

skpro/regression/gradient_boosting/_ngboost.py
fkiraly: @ShreeshaM07, I think the failure relates to this comment which you have not addressed?

ShreeshaM07:
> 
> Can you also describe what you have found out in debugging these failures? You can use `raise_exceptions=True` to get an error raise for debugging.

I have now resolved all the errors except `test_fit_does_not_overwrite_hyper_params`.
And I have found it is being caused due to initialisation of the base learner/estimator to `None` in `init` but  i am replacing it with `DecisionTreeRegressor` later on if estimator is default `None`.

Can you please suggest what i can do to solve this issue. I tried doing the initialisation to DecisionTreeRegressor which is being done in `_fit` in `init` itself if it is `None` but it then gives `test_constructor` and `test_sklearn_params` asssertion error.

skpro/regression/gradient_boosting/_ngboost.py
ShreeshaM07: I have addressed the clone issue it is not related to that. It is with the handling of the `estimator` which is `DecisionTreeRegressor` but `init` was `None`.

fkiraly:
See above - look at the error messages, they should tell you precisely what the problem is.
The parameters written to `self` must always remain unmutated and identical to what was passed in `__init__`.

ShreeshaM07:
> See above - look at the error messages, they should tell you precisely what the problem is. The parameters written to `self` must always remain unmutated and identical to what was passed in `__init__`.

Yes I have now made it `self.estimator_ ` to be passed to `NGBRegressor` so it is keeping the `__init__` params intact and `check_estimator` is giving all tests passed!

fkiraly:
Great! Looks excellent, I will give it a review once the tests pass. From superficial inspection, looks ready to merge!

fkiraly:
Hm, it looks like `check_estimator` tests are indeed passing! Very nice!

Well, with exception of the sixth (index 5) parameter set - do you have an idea why that parameter set specifically is failing? Seems nans are being produced.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,367 @@
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {"dist": "Normal"}
+        params2 = {
```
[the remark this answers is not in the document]
fkiraly: @ShreeshaM07, while we are checking, could you kindly move the empty param set to the first position? Not strictly needed, but easier to read.

skpro/regression/gradient_boosting/_ngboost.py
```diff
@@ -0,0 +1,381 @@
...
+        }
+        params5 = {
+            "dist": "LogNormal",
+            "learning_rate": 0.001,
+            "validation_fraction": 0.2,
+        }
+
+        params6 = {
+            "dist": "TDistribution",
+            "natural_gradient": False,
```
fkiraly, line 377: this one seems to be the failing example. Any ideas?
I cannot spot any issue. Is this perhaps a bug in `ngboost` itself?

fkiraly: t-distribution does have non-degenerate natrual gradient afaik, so that might not be it?
But perhaps it's worth trying a different combination here, e.g., normal with natural gradient, and tdistibution plain.

ShreeshaM07: I tried running with the `ngboost`'s regressor itself it still gave `Singular Matrix` error for the `load_diabetes` dataset for `T-Distribution`. For all other distributions it works as expected. So just in case it was actually due to some mathematical working involving the determinant of the matrix to not be equal to 0 in case of T-Distribution, I tried with `load_breast_cancer` dataset ,even for this it is giving Singular Matrix error. So it might actually be the case that there's a bug in `ngboost` T-Distribution itself.

fkiraly: what we could do, remove the parameter set in this PR, but file a bug report on `sktime`, with code that uses these parameters. Unless of course there is another explanation or a fix.

fkiraly (approved):
Nice, thanks!

This is ready to merge, could you open the issue on the `skpro` tracker to investigate the `TDistribution` failure? With code if you have it, otherwise just a quick note.

ShreeshaM07:
> Nice, thanks!
> 
> This is ready to merge, could you open the issue on the `skpro` tracker to investigate the `TDistribution` failure? With code if you have it, otherwise just a quick note.

I have now opened #291 along with the code causing the issue.

PR #213, opened 2024-03-14
[ENH] interfacing Poisson regressor from sklearn  

nilesh05apr:
#### Reference Issues/PRs
https://github.com/sktime/skpro/issues/7

#### What does this implement/fix? Explain your changes.
Added interface for Poisson Regressor

#### Does your contribution introduce a new dependency? If yes, which one?
None

#### Did you add any tests for the change?
None

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
  How to: add yourself to the [all-contributors file](https://github.com/sktime/skpro/blob/main/.all-contributorsrc) in the `skpro` root directory (not the `CONTRIBUTORS.md`). Common badges: `code` - fixing a bug, or adding code logic. `doc` - writing or improving documentation or docstrings. `bug` - reporting or diagnosing a bug (get this plus `code` if you also fixed the bug in the PR).`maintenance` - CI, test framework, release.
  See here for [full badge reference](https://allcontributors.org/docs/en/emoji-key)
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
- [ ] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [ ] I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant `Examples` section.
- [ ] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured
  dependency isolation, see the [estimator dependencies guide](https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

Issue #7, opened 2020-04-12
(wish)list of probabilistic regressors to implement or to interface 


fkiraly:
A wishlist for probabilistic regression methods to implement or interface.
This is copied from the R counterpart https://github.com/mlr-org/mlr3proba/issues/32 .
Number of stars at the end is estimated difficulty or time investment.

GLM
- [ ] generalized linear model(s) with regression link, e.g., Gaussian *
- [ ] generalized linear model(s) with count link, e.g., Poisson *
- [ ] heteroscedastic linear regression ***
- [ ] Bayesian GLM where conjugate priors are available, e.g., GLM with Gaussian link ***

KRR aka Gaussian process regression
- [ ] vanilla kernel ridge regression with fixed kernel parameters and variance *
- [ ] kernel ridge regression with MLE for kernel parameters and regularization parameter **
- [ ] heteroscedastic KRR or Gaussian processes ***

CDE
- [ ] variants of conditional density estimation (Nadaraya-Watson type) **
- [ ] reduction to density estimation by binning of input variables, then apply unconditional density estimation **

Tree-based
- [ ] probabilistic regression trees **

Neural networks
- [ ] interface tensorflow probability - some hard-coded NN architectures **
- [ ] generic tensorflow probability interface - some hard-coded NN architectures ***

Bayesian toolboxes
- [ ] generic pymc3 interface ***
- [ ] generic pyro interface ****
- [ ] generic Stan interface ****
- [ ] generic JAGS interface ****
- [ ] generic BUGS interface ****
- [ ] generic Bayesian interface - prior-valued hyperparameters *****

Pipeline elements for target transformation
- [ ] distr fixed target transformation **
- [ ] distr predictive target calibration **

Composite techniques, reduction to deterministic regression
- [ ] stick mean, sd, from a deterministic regressor which already has these as return types into some location/scale distr family (Gaussian, Laplace) *
- [ ] use model 1 for the mean, model 2 fit to residuals (squared, absolute, or log), put this in some location/scale distr family (Gaussian, Laplace) **
- [ ] upper/lower thresholder for a regression prediction, to use as a pipeline element for a forced lower variance bound **
- [ ] generic parameter prediction by elicitation, output being plugged into parameters of a distr object not necessarily scale/location ****
- [ ] reduction via bootstrapped sampling of a determinstic regressor **

Ensembling type pipeline elements and compositors
- [ ] simple bagging, averaging of pdf/cdf **
[...]

PR #222, opened 2024-03-25
[ENH] `GaussianRegressor` using statsmodels `GLM` with gaussian link

julian-fong:
#### Reference Issues/PRs

Implements a GLM model for the gaussian link in #7 , looking for feedback on the code changes

The Gaussian Regressor is a direct interface to the statsmodel package's GLM model, and _fit, _predict, and _predict_proba have all been defined as noted in the regression extension template (have not started on the class method get_test_params)

Yes, this contribution introduces a new dependency to the statsmodel package. A link to the package page can be found here: https://www.statsmodels.org/stable/index.html

This is just a rough draft (and there are probably loads of mistakes!) of the implementation. I am looking for feedback on the areas where I could improve/fix on code wise (on the constructor, fit, and predict methods), or even design wise.

Something I noticed in the GLM statsmodels package is that the fitted GLM model is returned as a separate object, as it is returned as part of a ```GLMResultsWrapper``` object.

Datasets X, y are to be defined in the constructor of the model, and not inside the fit function, so I opted to include the params of the GLM model's fit function inside the _fit function in ```Gaussian Regressor```. (link to the fit function: https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.fit.html#statsmodels.genmod.generalized_linear_model.GLM.fit)

I was considering whether or not to define the statsmodel GLM object inside the fit function instead of the constructor (similar to the ```MapieRegressor``` to follow the template to the letter, but later decided against. I am wondering if that is the recommended option to implement the ```GaussianRegressor``` as well.

Appreciate the feedback!

##### For all contributions

### DIFF
--- a/skpro/regression/linear/_glm.py
+++ b/skpro/regression/linear/_glm.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 from statsmodels.genmod.families.family import Gaussian
 from statsmodels.genmod.generalized_linear_model import GLM
+from statsmodels.tools import add_constant
 
 from skpro.regression.base import BaseProbaRegressor
 
@@ -25,18 +26,6 @@ class GaussianRegressor(BaseProbaRegressor):
 
     Parameters
     ----------
-    y : pandas DataFrame
-        1d array of the endogenous response variable. This array can be 1d or
-        2d. Binomial family models accept a 2d array  with two columns.
-        If supplied each observation is expected to  be [success, failure].
-        Equivalent to statsmodel's (endog).
-
-    X : pandas DataFrame
-        A (n,k) array where n is the number of observations and k is the number
-        of regressors. An intercept is not included by default and should be
-        added by the user (models specified using a formula include an
-        intercept by default). Equivalent to statsmodel's (exog).
-
     family : family class instance
         To specify the binomial distribution family = sm.family.Binomial() Each
         family can take a link instance as an argument.
@@ -65,6 +54,64 @@ class GaussianRegressor(BaseProbaRegressor):
         Available options are 'none', 'drop' and 'raise'. If 'none', no nan
         checking is done. If 'drop', any observations with nans are dropped.
         If 'raise', an error is raised. Default = 'none'
+
+    start_params : array_like (optional)
+        Initial guess of the solution for the loglikelihood maximization.
+        The default is family-specific and is given by the
+        family.starting_mu(endog). If start_params is given then the initial
+        mean will be calculated as np.dot(exog, start_params).
+        This parameter is used inside the GLM fit() function.
+
+    maxiter : int
+        Number of iterations. This parameter is used inside the GLM fit() function.
+
+    method : str
+        Default is 'IRLS' for iteratively re-weighted least squares.
+        This parameter is used inside the GLM fit() function.
+
+    tol : float
+        Convergence tolerance. Default is 1e-8. This parameter is
+        used inside the GLM fit() function.
+
+    scale : str/float
+        scale can be 'X2', 'dev', or a float. The default value is None,
+        which uses X2 for gamma, gaussian and inverse gaussian. X2 is
+        Pearson's chi-square divided by df_resid. The default is 1 for
+        the Bionmial and Poisson families. dev is the deviance divided
+        by df_resid. This parameter is used inside the GLM fit() function.
+
+    cov_type : str
+        The type of parameter estimate covariance matrix to compute.
+        This parameter is used inside the GLM fit() function.
+
+    cov_kwds : dict-like
+        Extra arguments for calculating the covariance of the
+        parameter estimates. This parameter is used inside the GLM fit() function.
+
+    use_t : bool
+        if True, the Student t-distribution if used for inference.
+        This parameter is used inside the GLM fit() function.
+
+    full_output : bool
+        Set to True to have all available output in the Results object’s
+        mle_retvals attribute. The output is dependent on the solver. See
+        LikelihoodModelResults notes section for more information. Not used
+        if methhod is IRLS. This parameter is used inside the GLM fit() function.
+
+    disp : bool
+        Set to True to print convergence messages. Not used if method
+        is IRLS. This parameter is used inside the GLM fit() function.
+
+    max_start_irls : int
+        The number of IRLS iterations used to obtain starting values for
+        gradient optimization. Only relevenat if method is set to something
+        other than "IRLS". This parameter is used inside the GLM fit() function.
+
+    add_constant : bool
+        statsmodels does not include an intercept by default. Specify this as
+        True if you would like to add an intercept (floats of 1s) to the
+        dataset X. Default = False. Note that when the input is a pandas
+        Series or DataFrame, the added column's name is 'const'.
 
     Attributes
     ----------
@@ -139,58 +186,6 @@ class GaussianRegressor(BaseProbaRegressor):
         The value of the weights after the last iteration of fit.
         Only available after fit is called. See statsmodels.families.family
         for the specific distribution weighting functions.
-
-    start_params : array_like (optional)
-        Initial guess of the solution for the loglikelihood maximization.
-        The default is family-specific and is given by the
-        family.starting_mu(endog). If start_params is given then the initial
-        mean will be calculated as np.dot(exog, start_params).
-        This parameter is used inside the GLM fit() function.
-
-    maxiter : int
-        Number of iterations. This parameter is used inside the GLM fit() function.
-
-    method : str
-        Default is 'IRLS' for iteratively re-weighted least squares.
-        This parameter is used inside the GLM fit() function.
-
-    tol : float
-        Convergence tolerance. Default is 1e-8. This parameter is
-        used inside the GLM fit() function.
-
-    scale : str/float
-        scale can be 'X2', 'dev', or a float. The default value is None,
-        which uses X2 for gamma, gaussian and inverse gaussian. X2 is
-        Pearson's chi-square divided by df_resid. The default is 1 for
-        the Bionmial and Poisson families. dev is the deviance divided
-        by df_resid. This parameter is used inside the GLM fit() function.
-
-    cov_type : str
-        The type of parameter estimate covariance matrix to compute.
-        This parameter is used inside the GLM fit() function.
-
-    cov_kwds : dict-like
-        Extra arguments for calculating the covariance of the
-        parameter estimates. This parameter is used inside the GLM fit() function.
-
-    use_t : bool
-        if True, the Student t-distribution if used for inference.
-        This parameter is used inside the GLM fit() function.
-
-    full_output : bool
-        Set to True to have all available output in the Results object’s
-        mle_retvals attribute. The output is dependent on the solver. See
-        LikelihoodModelResults notes section for more information. Not used
-        if methhod is IRLS. This parameter is used inside the GLM fit() function.
-
-    disp : bool
-        Set to True to print convergence messages. Not used if method
-        is IRLS. This parameter is used inside the GLM fit() function.
-
-    max_start_irls : int
-        The number of IRLS iterations used to obtain starting values for
-        gradient optimization. Only relevenat if method is set to something
-        other than "IRLS". This parameter is used inside the GLM fit() function.
     """
 
     _tags = {
@@ -223,6 +218,7 @@ class GaussianRegressor(BaseProbaRegressor):
         full_output=True,
         disp=False,
         max_start_irls=3,
+        add_constant=False,
     ):
         self.family = Gaussian()
         self.offset = offset
@@ -241,6 +237,7 @@ class GaussianRegressor(BaseProbaRegressor):
         self.full_output = full_output
         self.disp = disp
         self.max_start_irls = max_start_irls
+        self.add_constant = add_constant
 
         super().__init__()
 
@@ -253,14 +250,26 @@ class GaussianRegressor(BaseProbaRegressor):
         Parameters
         ----------
         X : pandas DataFrame
-            feature instances to fit regressor to
-        y : pandas DataFrame, must be same length as X
-            labels to fit regressor to
+            A (n,k) array where n is the number of observations and k is the number
+            of regressors. An intercept is not included by default and should be
+            added by the user (models specified using a formula include an
+            intercept by default). Equivalent to statsmodel's (exog).
+
+        y : pandas DataFrame
+            1d array of the endogenous response variable. This array can be 1d or
+            2d. Binomial family models accept a 2d array  with two columns.
+            If supplied each observation is expected to  be [success, failure].
+            Equivalent to statsmodel's (endog).
 
         Returns
         -------
         self : reference to self
         """
+        if self.add_constant:
+            X = add_constant(X)
+
+        y_col = y.columns
+
         glm_estimator = GLM(
             endog=y,
             exog=X,
@@ -290,10 +299,11 @@ class GaussianRegressor(BaseProbaRegressor):
             self.max_start_irls,
         )
 
-        FITTED_PARAMS_TO_FORWARD = ["glm_estimator_"]
+        # forward some parameters to self
+        FITTED_PARAMS_TO_FORWARD = {"glm_estimator_": fitted_glm_model, "y_col": y_col}
 
-        for param in FITTED_PARAMS_TO_FORWARD:
-            setattr(self, param, fitted_glm_model)
+        for k, v in FITTED_PARAMS_TO_FORWARD.items():
+            setattr(self, k, v)
 
         return self
 
@@ -315,7 +325,7 @@ class GaussianRegressor(BaseProbaRegressor):
         -------
         y : pandas DataFrame, same length as `X`, with same columns as y in fit
         """
-        y_column = self.endog.columns
+        y_column = self.y_col
         y_pred_series = self.glm_estimator_.predict(X)
         y_pred = pd.DataFrame(y_pred_series, columns=[y_column])
 
@@ -343,7 +353,7 @@ class GaussianRegressor(BaseProbaRegressor):
         from skpro.distributions.normal import Normal
 
         y_pred_series = self.glm_estimator_.predict(X)
-        y_mu = pd.DataFrame(y_pred_series, columns=[self.endog.columns])
+        y_mu = pd.DataFrame(y_pred_series, columns=[self.y_col])
         y_sigma = np.std(y_mu.values)
         params = {
             "mu": y_mu,
@@ -353,3 +363,26 @@ class GaussianRegressor(BaseProbaRegressor):
         }
         y_pred = Normal(**params)
         return y_pred
+
+    @classmethod
+    def get_test_params(cls, parameter_set="default"):
+        """Return testing parameter settings for the estimator.
+
+        Parameters
+        ----------
+        parameter_set : str, default="default"
+            Name of the set of test parameters to return, for use in tests. If no
+            special parameters are defined for a value, will return `"default"` set.
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {}
+        params2 = {"add_constant": True}
+
+        return [params1, params2]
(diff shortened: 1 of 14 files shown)


fkiraly (changes requested):
Thanks, this is great!
Looks like we're almost done.

What is important is that you follow the "scikit-learn-like" interface precisely, and that means:

* no data should be passed in `__init__`. The data is `X` and `y`. If this needs to be renamed or converged (e.g., to `endog` etc), do that in `_fit` or `_predict`
* `_fit` should have the signature `def _fit(self, X, y):`, no additional arguments. If there are additional arguments, they need to be in `__init__` and read from `self.argname`

Also see the `regression.py` extension template.

Minor things, but blocking (since it's code formatting):

* ensure you follow code formatting guidelines, here's the guide on setting it up locally: https://www.sktime.net/en/stable/developer_guide/coding_standards.html
* you accidentally added a newline in `_sklearn.py`, kindly revert that

Let us know if you need any help or have any questions about the above (e.g., on discord).

fkiraly:
> I was considering whether or not to define the statsmodel `GLM` object inside the `fit` function instead of the constructor (similar to the `MapieRegressor` to follow the template to the letter, but later decided against. I am wondering if that is the recommended option to implement the `GaussianRegressor` as well.

Yes, I think that's the right way to do it. Because `GLM` requires access to the data, and the earliest you see it is `_fit`. So your initial intuition was the right one.

julian-fong:
One more thing, should I include the ability to add any family to be present i.e to include all the distribution families? I know in #7 it only said to implement the gaussian link

fkiraly:
> should I include the ability to add any family to be present i.e to include all the distribution families?

That would be great, though:

* I'm not sure if all families have a corresponding distribution object already implemented
* I'd suggest we make sure first this passes all `check_estimator` tests. Adding more families before that runs might complicate work on this, since it risks introducing a back/forth between parts of logic.

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,387 @@
...
+            family=self.family,
+            offset=self.offset,
+            exposure=self.exposure,
+            freq_weights=self.freq_weights,
+            var_weights=self.var_weights,
+            missing=self.missing,
+        )
+
+        self._estimator = (
+            glm_estimator  # does this need to be cloned using some clone method?
```
fkiraly, line 284: no need to clone, since it is not a parameter of `__init__`.`

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,387 @@
...
+        )
+
+        fitted_glm_model = self._estimator.fit(
```
fkiraly, line 287: I would write this to `glm_fitted_`, and describe in the `Attributes` section of the docstring.

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,387 @@
...
+        X : pandas DataFrame, must have same columns as X in `fit`
+            data to predict labels for
+
+        Returns
+        -------
+        y : pandas DataFrame, same length as `X`, with same columns as y in fit
+        """
+        y_column = self.y_col
+        y_pred_series = self.glm_estimator_.predict(X)
+        y_pred = pd.DataFrame(y_pred_series, columns=[y_column])
```
fkiraly, line 329: I think you also need to make sure that `index` is same as of `X`

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,387 @@
...
+            self.cov_kwds,
+            self.use_t,
+            self.full_output,
+            self.disp,
+            self.max_start_irls,
+        )
+
+        # forward some parameters to self
+        FITTED_PARAMS_TO_FORWARD = {"glm_estimator_": fitted_glm_model, "y_col": y_col}
+
```
fkiraly, line 303: I would actually forward the attributes of `glm_estimator`, too.
See here for a list:
https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLMResults.html#statsmodels.genmod.generalized_linear_model.GLMResults

In `skpro`/`sklearn`, these should end with an underscore, while `statsmodels` does not have that convention.

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,387 @@
...
+            data to predict labels for
+
+        Returns
+        -------
+        y_pred : skpro BaseDistribution, same length as `X`
+            labels predicted for `X`
+        """
+        from skpro.distributions.normal import Normal
+
+        y_pred_series = self.glm_estimator_.predict(X)
```
fkiraly, line 354: That's a sensible variance estimate, but you need to get it from the GLM model - it produces its own varianve predictions!

The method is `get_prediction`, you can get mean and variance from it:
https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLMResults.get_prediction.html#statsmodels.genmod.generalized_linear_model.GLMResults.get_prediction

fkiraly (changes requested):
Great!

I left some comments above, most important is to get `predict_proba` right.

Also, you should test your model using `skpro.utils.check_estimator`, that utility allows you to quickly run relevant tests locally.

skpro/regression/linear/_glm.py
julian-fong: Thanks for that! I wasn't sure how to derive the mu and sigma and I realized I probably should have mentioned that in the main PR req

fkiraly: no worries!

That's actually one of the key points that `skpro` means to help the user and deveoper with - where or how exactly the distributions are represented differs wildly by package, and even within packages ilke `sklearn` and `statsmodels`. There is no clear standard, or unified API for distribution predictors out there

skpro/regression/linear/_glm.py
julian-fong: Should I include forward both the glm estimator ```_estimator``` and the glm fitted estimator (now named ```glm_fit_```) and their corresponding attributes? I will include everything inside the docstring as well.

skpro/regression/linear/_glm.py
julian-fong: Should I include forward both the glm estimator _estimator and the glm fitted estimator (now named glm_fit_) and their corresponding attributes? I will include everything inside the docstring as well.

fkiraly: Yes, I think that would be useful. The user can then choose what they look at (e.g., via `get_fitted_params`).

skpro/regression/linear/_glm.py
fkiraly: yes, see below

julian-fong:
Hey @fkiraly,

I made some code updates, and I ran ```check_estimators``` from ```skpro.utils.estimator_checks``` locally and found one re-occuring error: NotImplementedError('no conversion defined from type pd_DataFrame_Table to pd_DataFrame'). I am wondering how I can fix this issue

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,443 @@
...
+    """
+
+    _tags = {
+        "authors": ["julian-fong"],
+        "maintainers": ["julian-fong"],
+        "python_version": None,
+        "python_dependencies": None,
+        "capability:multioutput": False,
+        "capability:missing": False,
+        "X_inner_mtype": "pd_DataFrame",
```
fkiraly, line 218: this should be `pd_DataFrame_Table` both, that will fix your issue.

fkiraly: (reason: `pd_DataFrame` is not a valid data type string, none of the other distributions has that either)

fkiraly: I fixed it for you, since it looks like everything should run, if not for very well hidden bug.s

fkiraly:
The failure in no-extras is because `statsmodels` is a soft dependency. These should be imported only in methods.

However, `all-extras` tests pass, which seems to indicate that your estimator works! Nice!

fkiraly (changes requested):
Besides closing discussions above, the only outstanding item I can see is moving the `statsmodels` imports into the methods.

julian-fong:
> Besides closing discussions above, the only outstanding item I can see is moving the `statsmodels` imports into the methods.

Ive put the GLM and add_constant imports into the _fit method, but what about the Guassian family import that is defined inside the __init__ method. Should I leave that at the top of the file or move that somewhere else

fkiraly:
> but what about the Guassian family import as that is defined inside the init method. through `self.family`. Should I leave that at the top of the file or move that somewhere else

What you need to think about, if someone imports the class, and does not have `statsmodels` installed, no exceptions should be triggered.

All imports of `statsmodels` should happen within the methods, and if anything happens in `__init__`, it should be after the super call.

julian-fong:
I've resolved the discussions and moved the imports, thanks for all your help and guidance!

fkiraly (changes requested):
Apologies for this last minute request - for clarity, could we rename the estimator to `GLMRegressor` or `GLM`?

`GaussianRegressor` is a bit confusing, and we may like to add further link functions to this estimator.

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,444 @@
...
+
+    stand_errors_ : ndarray
+        The standard errors of the fitted GLM.
+    """
+
+    _tags = {
+        "authors": ["julian-fong"],
+        "maintainers": ["julian-fong"],
+        "python_version": None,
+        "python_dependencies": None,
```
fkiraly, line 212: this tag should have `statsmodels`, so `skpro` knows this requires the dependency.

julian-fong: I added the statsmodel dependency and renamed the module to GLMRegressor. I can refactor some of the code to include all of the available links inside the statsmodels module, but I think another member is currently working on the glm with count link

fkiraly: I think that's a different estimator, it's `sklearn`'s Poisson regressor, that's different from `statsmodels` `GLM` with Poisson link: https://github.com/sktime/skpro/pull/213/files

Further, I'd expect more link functions to be added to this class.

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,444 @@
...
+            self.scale,
+            self.cov_type,
+            self.cov_kwds,
+            self.use_t,
+            self.full_output,
+            self.disp,
+            self.max_start_irls,
+        )
+
+        PARAMS_TO_FORWARD = {
```
fkiraly, line 320: I don't think that the data relatd vars (endog to var_weights) are too interesting.

Either way, you should not overwrite any of the params set in `__init__`.

fkiraly (changes requested):
Fails since you overwrite one of the `__init__` params in `_fit`, please fix.

skpro/regression/linear/_glm.py
julian-fong: When i ran the tests, i got this result ``` 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-0-ProbaRegressorBasic]': 'PASSED',
 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-0-ProbaRegressorXcolMixIxYnp]': 'PASSED',
 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-0-ProbaRegressorSurvival]': 'PASSED',
 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-1-ProbaRegressorBasic]': 'PASSED',
 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-1-ProbaRegressorXcolMixIxYnp]': 'PASSED',
 'test_fit_does_not_overwrite_hyper_params[GLMRegressor-1-ProbaRegressorSurvival]': 'PASSED',```
 
its likely they passed because i have two family parameters: ```self.family```, and ```self.family_```. Regardless, i have removed ```self.family_```.

fkiraly: hm, maybe they end up identical?

fkiraly (changes requested):
Failures look genuine:

* one looke like an array size mismatch?
* the other one with the column names is likely due to you not setting the `columns` of the distribution correctly?

julian-fong:
Hi Franz, it seems like GLMRegressor-1 is failing for test_input_output_contract, and test_pred_quantiles_intervail. Can you give me any insight as to what GLMRegressor-1 is and how it differs from GLMRegressor-0? It seems all of the test cases for GLMRegressor-0 passes when i run ```check_estimator``` locally

fkiraly:
> Can you give me any insight as to what GLMRegressor-1 is and how it differs from GLMRegressor-0?

In `get_test_params`, you return a list of `dict`. This list, in the same order, is used to construct test instances, the first of which becomes `GLMRegressor-0`, the second `GLMRegressr-1` (in python we start counting at 0).

These instances are then used in the tests from `TestAllRegressors`, substituted to the `estimator_instance` args.

So, if `test_input_output_contract[GLMRegressor-1]` fails, it means that `estimator_instance = GLMRegressor(**params)`, where `params` is the second element in the list returned by `get_test_params` fails the test `test_input_output_contract`.

For more details, see

https://www.sktime.net/en/latest/developer_guide/testing_framework.html

julian-fong:
Thanks Franz! i believed i  figured out what the source of the array mismatch was. In statsmodels theres an option to add a constant if a user specifies it (the second parameter set specifies ```add_constant = True```. In that case, it'll create an extra column called 'const' inside the pandas dataframe during training. When the tests call _predict_proba, since there is an extra column from _fit, it'll create an array mismatch

fkiraly:
> When the tests call _predict_proba, since there is an extra column from _fit, it'll create an array mismatch because X_test does not account for that extra column

That sounds like a bug in `statsmodels`, in `get_predictions` specifically, no?

Either way, if we know what the issue is, can we compensate for it by adding the column ourselves? (i assume it is an all-ones column)

We should also report the bug in `statsmodels`, if you can verify that it's indeed coming from `get_predictions`, and we should open a tracking issue that links to the `statsmodels` issue.

julian-fong:
Yes, if the user specifies ```add_constant = True``` in initialization, we can compensate by adding it to the column ourselves if it does not exist already (this affects both the _predict and _predict_proba function). One thing to note is that in our docstring for both _predict and _predict_proba: its assumed that ```X``` has the same columns as ```X``` in `fit`

julian-fong:
i did some testing with a toy example as follows:

```import pandas as pd
import statsmodels.api as sm

# Create a sample dataset with two predictor variables (X1 and X2) and a response variable (y)
data = {
    'X1': [1, 2, 3, 4, 5],
    'X2': [2, 3, 4, 5, 6],
}

data2 = {
    'y': [3, 4, 5, 6, 7]
}
X = pd.DataFrame(data)
print(X)

# Add constant for intercept term
X = sm.add_constant(X)
print(X)
# Fit a GLM model

y = pd.DataFrame(data2)
model = sm.GLM(y, X, family=Gaussian())
result = model.fit()

# Print summary of the fitted model
# print(result.summary())

# Create a new observation for prediction
X_test = pd.DataFrame({'const': [1,1,1], 'X1': [6,2,4], 'X2': [7,2,5]})  # New observations with constant
X_test_ = pd.DataFrame({'X1': [6,2,4], 'X2': [7,2,5]}) #new observations without constant automatically set
# Predict with the model

print("-------predictions-------")
#will work
prediction = result.predict(X_test)
pred = result.get_prediction(X_test).summary_frame()
print(prediction)
print(pred)

#won't work
#prediction = result.predict(X_test_)
pred = result.get_prediction(X_test_).summary_frame()

# print("Prediction for new observation:", prediction)

y_test = pd.DataFrame(prediction, columns = y.columns)```

A value error of ```ValueError: shapes (3,2) and (3,) not aligned: 2 (dim 1) != 3 (dim 0)``` is then returned

**Comment by fkiraly** [comment]
> One thing to note is that in our docstring for both `_predict` and `_predict_proba`: its assumed that X_test has the same columns as `X_train` in `fit`

I mean, we do this only internally, to compensate for the bug in `statmodels` until it is fixed.
That is, if `add_constant=True`, then make a copy of the df where a column is added, and pass that to `statsmodels`. The `skpro` user doesn't have to do that.

Does this make sense?

At the same time, I would (i.e., I'd suggest you do that) report the bug in `statsmodels`, with a reproducible example. If you want a review before that, you can post the bug report here first.

**Comment by julian-fong** [comment]
> > One thing to note is that in our docstring for both `_predict` and `_predict_proba`: its assumed that X_test has the same columns as `X_train` in `fit`
> 
> I mean, we do this only internally in `_predict`, to compensate for the bug in `statsmodels` until it is fixed. That is, if `add_constant=True`, then make a copy of the df `X` where a column is added, and pass that (with the additional column) to `statsmodels`. The `skpro` user doesn't have to do that.

Understood, but would we not want to leverage the same idea for `_predict_proba` since that method also requires adding a column 'const' if it is missing?

**Comment by fkiraly** [comment]
> Understood, but would we not want to leverage the same idea for `_predict_proba` since that method also requires adding a column 'const' if it is missing?

Yes, of course - I would move the logic into a separate method, e.g., `_prep_X`, to avoid copy-pasting too much code.

File: skpro/regression/linear/_glm.py [hunk 9]
```diff
@@ -0,0 +1,443 @@
...
+            Equivalent to statsmodel's (endog).
+
+        Returns
+        -------
+        self : reference to self
+        """
+        from statsmodels.genmod.generalized_linear_model import GLM
+        from statsmodels.tools import add_constant
+
+        if self.add_constant:
```
fkiraly, line 266: do we need to do this in `_fit`, too?

skpro/regression/linear/_glm.py
```diff
@@ -0,0 +1,443 @@
...
+            data to predict labels for
+
+        Returns
+        -------
+        y_pred : skpro BaseDistribution, same length as `X`
+            labels predicted for `X`
+        """
+        from skpro.distributions.normal import Normal
+
+        if self.add_constant:
```
fkiraly, line 381: still a lot of copy-paste. I would move the if/else to the function too

skpro/regression/linear/_glm.py
julian-fong: yes, i'll change the code so its consistent with the other methods

fkiraly: I thought `statsmodels` does it in `fit`?

julian-fong: As per my knowledge, the ```GLM``` model on `statsmodels` does not automatically include a constant column at any point during any fitting or prediction process

fkiraly: I see, so you added this feature?

fkiraly:
FYI, if you want to indicate that you're done with making changes and addressing the review comments, you can click here:
image
(the circling arrows)

Because, to me it looks like this is ready?

fkiraly (approved):
Nice, no blocking comments.

Some comments which might still be worth addressing:

* `family` can be picked by the user, but only Gaussian works at the moment, other choices will cause the estimator to break.
* we should try to cover as many parameters as we can in `get_test_params`.
* the docstring should say, for every parameter, what possible values are. E.g., what are possible values for `cov_type`, `method`, what are expected sizes for `start_params`, etc.

julian-fong:
Sounds good! I'll keep my eye on this page and improve the code so that it is more robust when including other families. Thanks again for all the guidance!

fkiraly:
let me know if you want to do it in this PR, or separately.

julian-fong:
I'll make a seperate pr and include this pr so they are linked together (so this one doesn't get too congested with comments) once my GSoC proposal is submitted.

fkiraly:
ok - I've removed the parameters that are broken or not useable. You can handle them later if you want.

Pratham-ja:
Hi @julian-fong @fkiraly @iwitaly @JonathanBechtel  
 I have gone through the implementation of the code in skpro/regression/linear/_glm.py. As mentioned  in the https://github.com/sktime/skpro/issues/7 the wishlist is to add more regression links like inverse gamma and tweedie. As far as I understand your PR merge has successfully added GLM regressors with gaussian link. 
And now the[ skpro/regression/linear/_glm.py](https://github.com/sktime/skpro/blob/main/skpro/regression/linear/_glm.py) has other families of distributions also. Now  I am trying to add the statsmodel inverse gaussian family of regressor which uses inverse squared maily as their link function.
But I have noticed that[ statsmodel ](https://www.statsmodels.org/stable/glm.html#examples )and the inverse gaussian inversion distribtuion in skrpo/distribution/inversegaussian uses the[ scipy implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.invgauss.html#scipy.stats.invgauss). which are different. Statsmodel is basically calculating Var(Y) = phi * mu^3 and scipy uses Var(Y) = mu^3 * scale^2. So the value ._scale which we get from fitting our statsmodel is basically different from the way we use it after in our scipy for predictions and modeling the distribution. 
Can you please help me with my doubt. I am actually doing some of the wishlishts and trying to add other GLM families from statsmodels. The clarification will help me to code the right and clear approach. 
Note: (Tagging everyone so that anyone of you if can help me, kindly do reply to this message)
Thank you !!

Issue #188, opened 2024-01-28
[BUG] cyclic boosting - sporadic test failures due to convergence failurse


fkiraly:
The recently added cyclic boosting estimator sporadically fails tests due to failed convergence of the loss, e.g.,:

```
FAILED skpro/regression/tests/test_cyclic_boosting.py::test_cyclic_boosting_with_manual_paramaters - cyclic_boosting.utils.ConvergenceError: Your cyclic boosting training seems to be diverging. In the 9. iteration the current loss: 52.52700124396056, is greater than the trivial loss with just mean predictions: 20.816666666666666.
```

FYI @setoguchi-naoki, @felixwick

FelixWick:
I would suggest to fix the random_state when calling sklearn's train_test_split to get a reproducible result. Best to use one which converges for our test case ;).

fkiraly:
that fixes the failing test, though it might be worth checking if there is a user facing problem here.

FelixWick:
I don't think so. Usually, Cyclic Boosting is quite robust and convergence issues rare. The data set used in the test might not be the best choice for quantile regression.

fkiraly:
> The data set used in the test might not be the best choice for quantile regression.

Should we then replace it with a better choice?

FelixWick:
Maybe California housing?

fkiraly:
Hm, there are also still timeouts appearing in the tests - seems like 10% or cases or so, it runs forever. Example:

```
FAILED skpro/tests/test_all_estimators.py::TestAllEstimators::test_fit_does_not_overwrite_hyper_params[CyclicBoosting-ProbaRegressorBasic] - Failed: Timeout >600.0s
```

Would be great if you could diagnose what is going on and try to make sure the algorithm always terminates. You can run the tests locally by using `skpro.utils.check_estimator`.

Not a big issue as we can manage with the tests - it is also odd since the data sets used in testing are fairly small and similar to toy datasets that a typical user could start playing with (and scare them off if it takes ages to run, even if it's an accident).

FelixWick:
Hm, that's weird, because we have a hard stop at 10 iterations (if I'm not missing anything).

fkiraly:
Maybe they are just very long iterations then? 600 sec is 10min!

FelixWick:
That should not happen with small data sets. Maybe something is messed up. Let me have a look these days.

fkiraly:
[...]

PR #232, opened 2024-04-03
[ENH] update Johnson QPDistributions with bugfixes and vectorization (cyclic-boosting ver.1.4.0)

setoguchi-naoki:
#### Reference Issues/PRs

Fexes #190 and #188

- Modfied QPD's methods following interface of vectorized QPD
- Bug fix tests for QPD
- replace scipy.misc.derivative to findiff because it will be removed in scypy 1.12.0

yes, I have extras dependency 
findiff: https://findiff.readthedocs.io/en/latest/

Please check if there are any problems with test

No

#### Any other comments?

So sorry for my late contribution

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks!

Before I go into this in more detail, could you explain why we are removing `QPD_U`?

setoguchi-naoki:
> Thanks!
> 
> Before I go into this in more detail, could you explain why we are removing `QPD_U`?

Because un-bounded mode (J_QPD_extended_U) is not vectorized in latest cyclic-boosting package. It is not difficult to set boundaries from user in many cases, so I guess it no problem but it is better to ask Felix the reason for more detail.

fkiraly (changes requested):
What are your further plans with this?

* to fix bugs with the distribution, and in prep for refactor of parameter broadcasting in https://github.com/sktime/skpro/issues/251, I moved the parameter broadcasting logic int a single function - this has been duplicated three times, now it is in a single place. This will, unfortunately, clash with this PR.
* `QPD_B` `mean` and `var` do not seem to run, see #257, I would recommend to address that first, in isolation, so we can ensure that any changes are covered by tests.
could you kindly explain the removal of `QPD_U`? I thought that was a very useful distribution to have.
* re `findiff` soft dependency: if it is a core dependency of `cyclic-bootsting`, it does not need to be added explicitly.
* for changing parameters in estimators or objects, we need to follow deprecation policy: https://www.sktime.net/en/stable/developer_guide/deprecation.html - I recommend to change things in the regressor in a separate PR after this.
* to reduce coupling between the regressor and the multiple distribution classes, I suggest to introduce an intermediate layer that dispatches to all variants of the Johnson QPD, here: https://github.com/sktime/skpro/pull/254 - please let me know what you think. Happy to help integrating this.

fkiraly:
need any help, @felixwick, @setoguchi-naoki?

setoguchi-naoki:
@fkiraly 
Sorry, I have been concentrating on the other task lately. But I will review the discussion and make some improvements this week or this weekend.

setoguchi-naoki:
@fkiraly 
Allow me to check about QPD_U

> QPD_B mean and var do not seem to run, see https://github.com/sktime/skpro/issues/257, I would recommend to address that first, in isolation, so we can ensure that any changes are covered by tests. could you kindly explain the removal of QPD_U? I thought that was a very useful distribution to have.

As Felix said, It seems to be caused by extended QPD. Vectorized QPD which is included this PR will solve that issue. Regarding to QPD_U, cyclic boosting package doesn't have vectorized QPD with unbounded mode but extended QPD_U. And These methods have different interface. (extended QPD_U only accepts a single sample.)
It means that we need to implement different estimation logic at the same time. And QPD_U cannot benefit from higher speeds. I think that's a bit strange and a code that leads to complexity, but of course it is possible to implement them. Should I proceed?

> this has been duplicated three times, now it is in a single place. This will, unfortunately, clash with this PR.

Understood. I would like to import the modified code and remove the conflicts.

> to reduce coupling between the regressor and the multiple distribution classes, I suggest to introduce an intermediate layer that dispatches to all variants of the Johnson QPD, here: https://github.com/sktime/skpro/pull/254 - please let me know what you think. Happy to help integrating this.

I would like to try this after the above modifications. Please give me time to understand. It seems to me that how QPD_U is implemented is also relevant for this implementation to proceed.

fkiraly:
@setoguchi-naoki, thanks for the explanation!

In this case, I would leave `QPD_U` in, with the loop that currently exists. This can be replaced later if or when a vectorized interface becomes available.

Perhaps I might also help with suggesting a way to vectorize it. Do you have a reference to the full parametric form of it, with math?

> Understood. I would like to import the modified code and remove the conflicts.

Great!

I hope the changes make interfacing easier.
I also wrote an extension template for probability distributions, using hte new extension contract:
https://github.com/sktime/skpro/blob/main/extension_templates/distributions.py

The idea is that, if you implement `_pdf`, `_cdf`, etc, all you need to provide internally is an interfacec that uses already broadcast 2D `np.ndarray`. That should be much simpler than the current contract, and closer to what you were doing with the vectorized QPD implementation from 1.4.0. Hopefully.

Let me know if you have any suggestions on updating, changing, improving the extension contract - this is possible without major overhead as it has not been released yet.

> I would like to try this after the above modifications. Please give me time to understand. It seems to me that how QPD_U is implemented is also relevant for this implementation to proceed.

Yes, that makes sense - it is the appropriate order in my opinion, too.

setoguchi-naoki:
@fkiraly 
I added these improvements. 
1. Reimplementation of QPD_U
2. Bug fix caused by using Extentend QPD class
3. Resolving conflicts with Main branch
4. Integrate QPD_Johonson as an interface

And, QPDs does not support the log_pdf method, so I deleted the tag. Is there a problem?

fkiraly:
Thanks! Could you kindly address the code formatting failures? I will also post a longer review.

fkiraly (changes requested):
Thanks!

High-level comments:

* could you kindly ensure this aligns with the new API for distributions? I.e., `mean` should become `_mean` and get params from `self`, etc. Have a look at the new extension template, or the `Normal` distribution as an example.
    * happy to help if it does not work directly. I do hope that it helps a lot by removing boilerplate that you would otherwise have to write!
* we can't change init signatures of objects easily, especially not of `CyclicBoostingRegressor`. I think it is a good idea to change as you suggested, but we ought to go through deprecation. The procedure is described here: https://www.sktime.net/en/stable/developer_guide/deprecation.html - basically we need to leave the signature but raise warning messages to the user on how to move to the new interface.
    * I would suggest to do this as suggested, but in a separate PR to make the complexity of review and back/forth more manageable, because you need to add warnings, notes for the release manager, etc. You can already open that PR if you branch off this one.
    * let me know if you need assistance on this. As said, the signatures suggested by you in this PR seem sensible to me.
* regarding removal of the `log_pdf` capability: you can leave the tag, since if not provided, `log_pdf` will juts take the log of `pdf`, and that is numerically exact if `pdf` is.

fkiraly (changes requested):
The test failures are due to a simple reason - after the refactor, what you get in `_cdf` etc is a `np.ndarray` (I hoped that would make things simpler).

You can get parameters broadcast to self.shape from `self._bc_params["upper"]`, in each method, so you also do not need to make the params arguments.

setoguchi-naoki:
@fkiraly 

> The test failures are due to a simple reason - after the refactor, what you get in _cdf etc is a np.ndarray (I hoped that would make things simpler).

Yes, I noticed that later. 

> You can get parameters broadcast to self.shape from self._bc_params["upper"], in each method, so you also do not need to make the params arguments.

In this case, lower and upper parameters don't need to broadcast because these has same range with each samples. Hence, May I turn off broadcast_init? 

The reason I set the lower and upper parameters for each method such as mean was because I needed to set the range when performing differential calculations for pdf using findiff. On the other hand, there was no need to specify it when estimating a probability distribution using QPDs. This specification may have been difficult to understand due to its seeming contradiction. Therefore, lower and upper are unified as parameters of the class.

fkiraly:
> In this case, lower and upper parameters don't need to broadcast because these has same range with each samples. Hence, May I turn off broadcast_init?

I see! Though, you stlil broadcast `qv_low` etc, no? So you broadcast some params while not others.

In this case, it might be best if you set the `broadcast_params`, tag, a list of `str` with the names of params you broadcast. You can access those then, internally, and do not need to broadcast manually.

setoguchi-naoki:
> I see! Though, you stlil broadcast qv_low etc, no? So you broadcast some params while not others.

Thank you for helpful information. In fact, these parameters are broadcasted in QPD class of cyclic boosting. So, I think that's operation does not need in skpro.

fkiraly:
I see - in this case, I'd still set the tag, because that also ensure `index` and `columns` are handled, but you do not need to use the dict with the parameters.

fkiraly:
should I help with the refactor? 2.3.0 will release soon, it would be nice if all distributions had moved over.

Also, minor unrelated query - would you be interested to present the cyclic boosting package or algorithm in one of the `sktime` meet-ups? Fridays, 1pm. We can chat on the discord: https://discord.com/invite/54ACzaFsn7

setoguchi-naoki:
I'm very sorry for being late. It took a long time to debug. If the test passes successfully, I think the tasks to be carried out in this PR will be completed. A request was made to #320  for a modification related to changing the default parameters. Since I have no experience in implementing the deprivation process, it would be helpful if you could confirm that I have not omitted any necessary corrections.

> Also, minor unrelated query - would you be interested to present the cyclic boosting package or algorithm in one of the sktime meet-ups? Fridays, 1pm. We can chat on the discord: https://discord.com/invite/54ACzaFsn7

Thank you for your suggestion. This is a good opportunity to increase the presence of Cyclic Boosting, but unfortunately due to schedule constraints, it may be difficult for me to make any presentations.

FelixWick:
I can do a Cyclic Boosting presentation at your meetup.

fkiraly (approved):
Extellent, thanks.

Almost ready! I will wrap this up together with the deprecation instructions so it can go in the next release.

Issue #235, opened 2024-04-04
[ENH] Consolidate quantile parameterized distributions in few classes


fkiraly:
@felixwick, @setoguchi-naoki, @Ram0nB I think we should consolidate the three curent quantile parameterized distribution (QPD) classes into a single one, and move the distribution logic in the quantile regressor in another QPD related class.

This idea is based on two observations:

* the quantile-to-distribution logic in current `MultipleQuantileRegressor` is one way to construct a QPD with arbitrary number of input quantile predictions/values. PR showing how this could look like here: https://github.com/sktime/skpro/pull/236
* the Johnson QPD taking lower, middle, upper quantile predictions/values, output by `CyclicBoosting`, are spread across three classes which seem a bit redundant. They are also being reworked in #232, related to the `cyclic_boosting 1.4.0` release.

I would suggest the following end state:

* the logic in `MultipleQuantileRegressor` moves into a QPD, to which we can later add different interpolation modes and basis functions.

* I would suggest replacing the three `QPD_X` distributions by a single class for Johnson QPD, `QPD_Johnson`, or similar. This can have a parameter `bounding` which can be `"bounded"`, `"unbounded"`, `"semi"`, mapping onto the logic in the three current classes.

Do you have any thoughts about this suggestion? It might simplify the interface(s) of the connected regressors as well.

FelixWick:
I think that makes sense.
Also, instead of importing the J-QPD implementation from Cyclic Boosting, we might consider to create an independent class directly here in the new QPD module. (I know I suggested differently before, but your description changed my mind.)

fkiraly:
> Also, instead of importing the J-QPD implementation from Cyclic Boosting, we might consider to create an independent class directly here in the new QPD module. (I know I suggested differently before, but your description changed my mind.)

I think the "optimal state" is either that, or a lightweight interface, as for instance to `scipy` with `Poisson`.

fkiraly:
Done in https://github.com/sktime/skpro/pull/232

PR #236, opened 2024-04-04
[ENH] empirical quantile parameterized distribution

fkiraly:
Towards #235.

This PR adds an empirical quantile parameterized distribution `QPD_Empirical`, inheriting from `Empirical`, but parameterized by quantile points/predictions.

This PR also factors out the `predict_proba` logic from `MultipleQuantileRegressor` as a parameter change between `QPD_Empirical` and `Empirical`, and uses `QPD_Empirical` directly to simplify the logic. This part is a pure refactor and does not change the internal logic.

As it changes the type of the return object, we may consider a deprecation cycle.

For discussion at the moment.

Alternative approaches for consideration:

* adding an argument to `Empirical` which, if passed, does the reparametrization.
* not inheriting form `Empirical`, but from a joint mixin or parent.

FYI @Ram0nB, @felixwick, @setoguchi-naoki - I would be interested to hear what you think.


FelixWick:
The refactoring seems ok to me.
General comment: I think you shouldn't call the empirical quantile fitting QPD, since it is not really parameterized by the quantiles. (If I understand your Empirical mode correctly as kind of spline fit.)

fkiraly:
> General comment: I think you shouldn't call the empirical quantile fitting QPD, since it is not really parameterized by the quantiles. 

Could you explain this? I think it is fully parameterized by the quantiles.

The argument `quantiles` containts:

* quantile points in the index, e.g., `[0.2, 0.5, 0.8]`
* the corresponding quantiles (quantile values), e.g., `[40, 42, 47]`

FelixWick:
But the quantiles are used as inputs to a spline that is fitted (i.e., parameters of the spline are estimated). The quantiles are not the parameters of the spline function. That means you have to do a fit for each sample. A QPD on the other hand does not do a fit, instead the quantiles are directly used as parameters of the function (what is much faster, of course).

fkiraly:
> But the quantiles are used as inputs to a spline that is fitted

Oh, I see.

Actually, no spline is fitted here. The distribution is just a weighted mixture supported at the quantile points, so in fact they are parameters quite directly, no?

FelixWick:
Ok, I see. Yeah, in that case I think it's ok to call it QPD 👍. Thanks for the explanation.

Ram0nB:
The refactoring looks good to me.

One thing that comes to mind is whether we also want to add the logic of converting a quantile prediction to a distribution estimate to the BaseProbaRegressor. Currently, the implementation of BaseProbaRegressor's _predict_proba uses the var and mean prediction to return a normal distribution. Maybe we can enhance BaseProbaRegressor's _predict_proba such that it uses the QPD_Empirical if _predict_quantiles/_predict_interval are available, and else the current logic. This way, we don't assume a normal distribution if multiple quantiles are available. What are your thoughts on this @fkiraly ?

fkiraly:
moved discussion here:
https://github.com/sktime/skpro/issues/246

PR #243, opened 2024-04-07
[MNT] Fix broken api source links in latest docs version

duydl:

A docs bug discussed on Discord. The source link is pointing to the non-existent latest branch instead of main. I think this would be the simplest solution, though you would have to merge to see if it works. A relevant part in sktime docs for more clarification  https://github.com/sktime/sktime/blob/9838adfed3ebd5fcf42805ddd6448606c2743585/docs/source/conf.py#L33


fkiraly (approved):
Thanks!

PR #247, opened 2024-04-16
[ENH] adapter to `lifelines`, most distributional survival regressors interfaced

fkiraly:
Adds an adapter to `lifelines`, exposing array-like survival functions as `Empirical` distributions in `predict_proba`.

Adds all models from `lifelines` which are capable of full distributional predictions:

* `AalenAdditiveFitter`
* `CoxPHFitter`
* `WeibullAFTFitter`

Remaining AFT fitters require distributions not yet merged.

### DIFF
--- a/docs/source/api_reference/survival.rst
+++ b/docs/source/api_reference/survival.rst
@@ -90,6 +90,17 @@ Proportional hazards models
     CoxPHSkSurv
     CoxNet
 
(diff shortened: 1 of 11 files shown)


fkiraly:
`AalenAdditiveFitter` fails as it produces improper survival functions.
A potential `skpro` sided fix is clipping to proper.

PR #254, opened 2024-04-18
[ENH] simplified interface to Johnson QPD

fkiraly:
This PR adds a simplified interface to the three Johnson QPD (bounded, lower semi-bounded, unbounded) to the module.

The interface uses the delegator introduced in https://github.com/sktime/skpro/pull/252, and is a simple parameter based dispatcher:

* the umbrella class has parameters `lower` and `upper`, and the other parameters which are common to `QPD_U`, `QPD_S`, `QPD_B`
* if `lower` and `upper` are not provided, it dispatches to `QPD_U`
* if `lower` is provided but not `upper`, it dispatches to `QPD_S`
* otherwise, it dispatches to `QPD_B`

As previously discussed in #232 - moving to an umbrella interface might also make the work carried out there easier.

What do you think, @setoguchi-naoki, @felixwick?


fkiraly:
supserseded by https://github.com/sktime/skpro/pull/232 which contains this

PR #255, opened 2024-04-18
[DOC] docstring with mathematical description for `QPD_Empirical`

fkiraly:
This PR adds a mathematical description in the docstring of `QPD_Empirical`.

The docstring also explains why the distribution is quantile parameterized, and its relation to `Empirical`.

A review would be appreciated, @Ram0nB, @setoguchi-naoki, @felixwick - I also hope (and would appreciate a check) for typos in the math.

FYI @vascosch92, as this is an example where two formal objects:

* are equivalent if considered without parameterization
* but not equivalent if considered together with the parameterization

and the more general question what the class/design principle should be here. Compare `Fibonacci` vs `OEIS(n)` object. Also relates to the "multiple names" discussion in https://github.com/VascoSch92/sequentium/issues/49, I am currently of the opinion that different class names should imply difference in parametric object.

### DIFF
--- a/skpro/distributions/qpd_empirical.py
+++ b/skpro/distributions/qpd_empirical.py
@@ -12,9 +12,38 @@ from skpro.distributions.empirical import Empirical
 class QPD_Empirical(Empirical):
     """Empirical quantile parametrized distribution.
 
-    This distribution is parameterized by a set of quantile points.
+    This distribution is parameterized by a set of quantile points and quantiles,
+    quantiles :math:`q_1, q_2, \dots, q_N`
+    at quantile points :math:`p_1, p_2, \dots, p_N`,
+    with :math:`0 \le p_1 < p_2 < \dots < p_N \le 1`.
 
-    todo: add docstr
+    It represents a distribution with piecewise constant CDF and quantile function,
+    the unique distribution satisfying:
+
+    * the support is :math:`[q_1, q_N]`
+    * for any quantile point :math:`p \in [p_i, p_{i+1}]`, it holds that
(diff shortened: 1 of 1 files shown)


skpro/distributions/qpd_empirical.py
```diff
@@ -10,11 +10,40 @@
...
+
+    In vernacular terms, the quantile function agrees with the quantiles prescribed by
+    :math:`q_i` at the quantile points :math:`p_i`, and for other quantile points
+    agrees with the value at the nearest quantile point.
+
+    In explicit terms, the distribution is an empirical distribution (sum-of-diracs),
+    supported at the quantiles :math:`q_1, q_2, \dots, q_N`,
+    with weights :math:`w_1, w_2, \dots, w_N`
+    such that :math:`w_i = (p_{i+1} - p_{i-1})/2` for :math:`1 = 1, \dots, N`,
+    where we define :math:`p_0 = -p_1` and :math:`p_{N+1} = 2 - p_N`.
```
VascoSch92, line 37: Just one question:
you say: `such that w_i = (p_{i+1} - p_{i-1})/2  for 1 = 1, \dots, N ` and then you define `p_0` and `p_{N+1}`. Why you define them? How are these two values playing a role here?

VascoSch92: Ah sorry now I see ;-)

thanks for the explanation. I didn't check that there it was a `p_{i-1})` and we have to define `p_0` (same for `p_{i+1}`

fkiraly: the $w_i$ should sum to 1, hopefully

Issue #23, opened 2023-08-23
[DOC] extension template for probability distributions


fkiraly:
We should write an extension template for probability distributions - possibly we may like to do this after some of the improvements in https://github.com/sktime/skpro/issues/21
which will change the extender contract slightly.

This should look similar to the `sktime` extension templates:
https://github.com/sktime/sktime/tree/main/extension_templates

From @alex-jg3 in #21, it is important to highlight the interplay between the tags (what is exact/approximate?) and the option to not implement certain methods: There are four cases:

* method must be implemented, or one out of k methods must be implemented
* method need not be implemented as there is a default, but can be, to replace a less accuate or less efficient default

combined with:

* method results in numerically exact return
* method return is approximate

Issue #251, opened 2024-04-17
[ENH] improved boilerplate for distibution objects


fkiraly:
Notes on improving boilerplate for distribution objects:

* creating boilerplate layer - this should encompass broadcasting, index/column subsetting at the start of methods.
* we should think about dynamic docstrings to contain explicit formulae for different distribution methods, where available. Possibly a dict with math formulae, and the same approach as `set_config` in `sktime`.
* it would be good to also support parameter/index cases for scalar and multivariate single-instance. In the first case, the behaviour should return scalars, in the second, 2D tables.

Also for discussion.

fkiraly:
Covered by https://github.com/sktime/skpro/pull/265, except dynamic docstrings

PR #265, opened 2024-04-21
[ENH] probability distributions - boilerplate refactor

fkiraly:
Work in progress.

Towards #251.

This PR refactors the probability distributions, inheriting from `BaseDistribution`:

* adds a boilerplate layer dealing with input coercion, formatting, indices
* extension points move from public to private under boilerplate, e.g., to `_pdf` from `pdf`
    * an extension template is added, detailing the extension contract, fixes https://github.com/sktime/skpro/issues/23
* distributions are now polymorphic - if all parameters are scalar, and no index is given, behaves as a scalar. Otherwise behaves as a 2D array distribution, as before.

Also refactors all distributions with "non-complex" signature, and also with exception of QPD (which is subject to multiple other PRs)


fkiraly:
Note to self: refactor is downwards compatible, but default broadcasting cannot deal with the empirical distribution, where parameters are multi-indexed DataFrame.

fkiraly:
FYI @Alex-JG3, a while ago you worked on refactoring and simplifying the broadcasting logic. It would be great if you could look at this PR, especially the end state of `Normal` and `Laplace`.

PR #282, opened 2024-04-25
[ENH] mean-scale family of distributions, composite

fkiraly:
This PR adds a composite distribution, the mean-scale family of distributions.

While inherently useful, this is also used as a development case for composite distribution where parameters are broadcast alongside component distributions, and the test framework around this case.


fkiraly:
Work in progress:

* add tests for general case
* ensure broadcasting works
* test mean-scale of standard normal against normal with mu, sigma

Issue #22, opened 2023-08-23
[ENH] (wish)list of probability distributions to implement


fkiraly:
It would be great to have a basic set of probability distributions implemented.

Umbrella issue for implementing `sktime` probability distributions.

High priority:

- [ ] laplace
- [ ] empirical incl delta
- [ ] multivariate normal

mid priority:

- [ ] t-distribution
- [ ] Cauchy (special case of Cauchy)
- [ ] mixture composition
- [ ] truncated, compositor

low priority:

- [ ] uniform continuous
- [ ] beta
- [ ] chi-squared
- [ ] logistic
- [ ] log-normal
- [ ] Weibull

Should extend `BaseDistribution`, a good template is `Normal`.

Mirrors https://github.com/sktime/sktime/issues/4518
(for high and mid priority)

bhavikar04:
Hi, I'm interested in taking this up. Would you say priority of the distributions aligns with the difficulty in implementation? I'd like to do either multivariate normal or uniform continuous.

fkiraly:
Hmmm, I'd say it is currently actually the opposite way. That is, the remaining low priority ones are easier to get started with, than the remaining high priority ones - simply since the easy higher priority ones are already done.

So, uniform continuous then? Parameterized by lower and upper.

an20805:
Hey @fkiraly, I have implemented uniform continuous distribution in my local branch. How do I proceed further? 
I would also love to implement other distributions.

fkiraly:
@an20805, nice! Let's not duplicate then, @bhavikar04 - how about beta?

The next step would be making a pull request to this repository, and a review cycle, then merge.

fkiraly:
Re energy, for `X, Y\sim Unif(a, b)`, I get:

$\mathbb{E}[X - y] = |y - \frac{b+a}{2} |$ if $y$ lies outside $[a, b]$,
and $\mathbb{E}[X - y] = \frac{(b-y)^2}{2(b-a)}+ \frac{(a-y)^2}{2(b-a)}$ if inside,

and

$\mathbb{E}[X - Y] = \frac{1}{3} (b-a)$ - double checking appreciated.

bhavikar04:
In that case I'll take up log normal distribution then.

fkiraly:
Pinging @Alex-JG3 and @ivarzap who most recently implemented distributions, in case you have any general starter advice.

bhavikar04:
Hey,

[...]

PR #287, opened 2024-04-29
[ENH] Adapter for Scipy Distributions

### LINKED ISSUE #227: [ENH] adapter for `scipy` distributions
Adapting `scipy` distributions is very formulaic and could easily be dealt with by an adapter class. This also avoids duplication in implementation efforts, as `scipy` is a core dependency.

This may require some abstraction around methods, but it seems mostly like a delegator class, as `scipy` does its own broadcasting.

Further discrepancies to be mindful of:

* `energy` and similar integrals are not implemented in `scipy`
* the class parameterization is different, `scipy` uses class methods whereas `skpro` uses `__init__` based parameterization

Good first issue with a design/architecture flavour, can be leveraged to cover a lot of ground in https://github.com/sktime/skpro/issues/22.

malikrafsan:
#### Reference Issues/PRs

Fixes #227 

- Adapter for Scipy distributions
- Fisk Distribution using Scipy Adapter
- Poisson Distribution using Scipy Adapter

no

- [ ] Adapter structure

- I try to install this library locally using `pip install --editable .[dev,test]` and create a simple driver program to use this new distribution
- I try to compare several methods of the existing Fisk distribution with proposed one and they produce the same result 

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks!

This was extremely quick and exactly what I meant! Really nice!

Some comments:

* good idea to have separate discrete/continuous. Currently, the interface indeed is different. There is also the issue of not having pmf in the base class.
    * for the time being, I think we can leave the handling of `pmf` and `log_pmf` as you suggest. However, you have certainly noted the missing handling of `pmf` in the base class, and that's why you tried to work around it wiht the second class. I would recommend, could you open an issue on implementing the `pmf` in the base class? Not necessarily sth to work on right now, but to keep track of what we still need to do. You could also pick it up later in a separate PR, if you'd like.
* what I would do is *replace entirely* the current `Fisk` and `Poisson` distribution implementations with the scipy based one. There is no difference in the logic, only the code is more streamlined through the adapter.
    * don't worry about delete/replace, the classes were written expressly with the idea in mind to later refactor them to a single adapter. What is important to check that this does not change anything, so is a proper refactor - see below.
* what could be helpful is to add some tests for the adapter class, e.g., take `Fisk` and check its outputs against direct outputs from `scipy`. That way we test that the adapter class is really adapting 1:1, and we can check that "before and after" are the same, for this refactor.
* finally, may I suggest to pick one more distribution from `scipy` and add it as an `skpro` object? How about a beta distribution, given that it is on the priority list in #22?
    * with the adapter, it should now be easy to add dozens of distributions with minimal work overhead, by using `scipy`. I would leave this to separate PR. And of course the more exotic properties like energy will not be provided by `scipy`.

malikrafsan:
> I would recommend, could you open an issue on implementing the pmf in the base class? 

Sure! I would love to do it

malikrafsan:
> don't worry about delete/replace, the classes were written expressly with the idea in mind to later refactor them to a single adapter.

Ahh I see, previously, I was a bit afraid to delete them 😅 Okay, I will replace them!

malikrafsan:
> what could be helpful is to add some tests for the adapter class, e.g., take Fisk and check its outputs against direct outputs from scipy. That way we test that the adapter class is really adapting 1:1, and we can check that "before and after" are the same, for this refactor.

I tried to run the test, but it turns out that, after changes, Fisk and Poisson failed the test. I will try to understand the test cases and pytest first as this is also my first time having hands-on on pytest.

Also for confirmation, from what I understand, it is better for me to make new test file to compare the result of the scipy adapter (let's say `Fisk`) vs result directly from scipy right? I will try to make it after fixing the test first 

Screenshot from 2024-05-01 17-33-02

fkiraly:
(I fixed the typo in the `_ppf` arg, hope that is ok)

malikrafsan:
> (I fixed the typo in the _ppf arg, hope that is ok)
Sure! Absolutely no problem for me! Thank you for fixing it 😄

> All looks good, unfortunately the failures are on 3.8, from type annotations, square brackets are not supported. Would you be so kind to remove these? These are internal annotations anyway, in private methods.
Thank you so much for pointing this out! I have removed the type annotations square brackets and the test work fine after that! 😄

> what could be helpful is to add some tests for the adapter class
I have also added the pytest for scipy adapter class, please let me know if you have any feedback for the test!

fkiraly (approved):
Thanks - this is great!

In a separate step, we may want to move the "return 0" logic into the `BaseDistribution` class, and base it on the tag `distribution_type`.

fkiraly:
What would be great is adding a new distibution, to check whether it is indeed easy - how about beta? You can do this in another PR based on this one.

malikrafsan:
> What would be great is adding a new distribution, to check whether it is indeed easy to use this adapter - how about beta? You can do this in another PR based on this one.

Sure, I would like to do that!

Issue #244, opened 2024-04-08
[ENH] inspectable set-valued domains for distributions


fkiraly:
It would be useful if distributions had some degree of inspectability with respect to their domains (sets), use cases include:

* approximation defaults, e.g., integer domain distributions know their domain, so expectations can be computed as sums
* typing for distributions, e.g., discrete, continuous, mixed
* inspecting the support of the discrete (deltas) part for mixed distributions, related to representing continuous and discrete parts separately: https://github.com/sktime/skpro/issues/229
* generating useful defaults in plotting routines
* testing distributions; sets may also be useful in specifying parameter ranges in estimators

Some discussion has already taken place here: https://github.com/VascoSch92/sequentium/issues/46
also regarding possible ways to implement this.

Options discussed:

* using `scipy` `Sets`, possibly also `stats`
* de-novo implementation, following `BaseObject`
* something similar to `sklearn.utils` parameter checking

Some issues from `skpro` architecture which may not be obvious how to cover:

* distributions are "tabular" (matrix distributions with pandas-like row and column index). Domains may vary over entries of the table.
* we may need parameteric sets, though that is not certain. Afaik only `BaseObject` supports parametric objects? Composites are supported by all three options above.

fkiraly:
FYI @VascoSch92

VascoSch92:
I will start to work on a first version of a module for symbolic representation of sets.

The idea is to extend from `BaseObject`. 

I still don't get 100% what it would be the application in `skpro.distributions`, but I will try to mimic the API of [set6](https://joss.theoj.org/papers/10.21105/joss.02598)

I will open a draft PR as soon as I have something interesting. In this way we can discuss the code.

fkiraly:
Great!

So you think it's better to inherit from `BaseObject` than using the existinglogic in `sympy`?

If I may ask, what are your pros/cons and weighting? Just curious.

VascoSch92:
Actually I was playing a little bit with the set implementation of `sympy` and I have to admits that it is pretty nice. 

One could think to use that and extend it to implement the measure of a set and integral computations. 

However, adding `sympy` to the dependencies can be over over-killing. Do we need so much power for the purpose of the project? Can we just install the specific module which takes care of sets? 

[...]

PR #292, opened 2024-05-03
[ENH] inspectable set-valued domains for distributions

VascoSch92:
#### Reference Issues/PRs
First version fro implementation of #244 

The PR implements a first version for set-valued domains for distributions.

There are no tests for the changes. This is a point in the PR checklist. For instance, this is just a. draft PR to discuss a first approach distributions domains.

#### Any other comments?

As said, this is a draft PR to start a discussion about domains for distributions. 

For instance, the most important domains for distributions that `skpro` needs are
-  interval of the real line (class `Interval`)
- finite sets of the real line (class `Finite`)
- direct product of the above domains (class `Product`)

Each of these class implements the methods:
- `__contains__` which can be used to elegantly check if an element is in the domain of an interval
- `boundary` which can be used for integration 

I found no other method that can be useful, for now. Perhaps, a couple of example could clarify the situation.

About Infinite sets of point. They can be of two kinds:
- discrete
- continuous (if they have an accumulation point, e.g., $` \{ \frac{1}{n} : n \in \mathbb{N} \}`$

Which of them do we need? There are examples of distributions which needs these sets?

Finally, I did not implement tags, because of two reasons:
- I'm not 100% sure how tags works (the are tags of subclasses override tags of parent classes or are added?)
- I didn't know which tags could be useful in this context. 

#### PR checklist

- [ ] Implement tests for the new classes
- [ ] Document the various classes
- [ ] Add `_tags`


skpro/domains/domains.py
```diff
@@ -0,0 +1,165 @@
...
+        super().__init__()
+
+    def _validate_values(self, values: Tuple[float, ...]) -> List[float]:
+        """The private method checks if a tuple of numbers is elegible to be a finite set."""
+        for value in values:
+            if not isinstance(value, (float, int)):
+                raise TypeError(f"Expected `float`or `int`, but got {type(value)}.")
+            if value in [-float("inf"), float("inf")]:
+                raise ValueError(f"Value {value} not accepted in finite set.")
+        return sorted(values)
```
fkiraly, line 117: hm, how about `list(set(values))`? Tha talso makes them unique, and might help us later with intersection.

fkiraly (commented):
Thanks a lot!

I have now had time to review and think about this design.
I think it's great, of course still work in progress.

I want to add three things to think about - these are not change requests but thinking points.

1. How do we use this from `BaseDistribution`? I was thinking about two attributes - `domain` and `support`. The `support` is the locus of non-zero mass/density, domain is the full range. For now, `domain` is perhaps redundant, but it will get useful for categorical rv.
    * a technical point, how would this look like? I would use a property and return a clone each time.
2. an open question that I have no good answer for - how do we deal with the a dataframe/array case of distribution? Suppose we have the `Uniform` distribution, it has lower and upper interval. But this can be different for each index. So what is `support` in this case? Some kind of array set? You have `Product`, but I was wondering whether there should be a more direct support of the array case - building it into the core API rather than by composition, like with `BaseDistribution`.
3. I was also thinking about the Lebesgue decomposition, i.e., continuous part plus discrete part. We need to express both domains, somehow, and make them queriable.
4. finally, this is more of a crazy idea: should we decorate all functions with their range and domain? You could think of `getattr(my_distr, "pdf")` returning not a method but an object, like `loc` and `iloc` in `pandas` or `BaseDistribution`. This could have attributes, so doing sth like `my_dist.pdf.domain` and `my_dist.pdf.range` will return a set object. What do you think of it? Maybe it is overengineered, I would not do it right away anyway.

skpro/domains/domains.py
VascoSch92: You are right., because if we define a finite set like Finite((1, 1, 2, 2)), then we need to know just the points (1, 2). Moreover, I was not checking for repeated elements.

I update that.

VascoSch92:
> How do we use this from BaseDistribution? I was thinking about two attributes - domain and support. The support is the locus of non-zero mass/density, domain is the full range. For now, domain is perhaps redundant, but it will get useful for categorical rv.
a technical point, how would this look like? I would use a property and return a clone each time.

I think using `domain` and `support` is a good idea, and cloning each time is clean.

> an open question that I have no good answer for - how do we deal with the a dataframe/array case of distribution? Suppose we have the Uniform distribution, it has lower and upper interval. But this can be different for each index. So what is support in this case? Some kind of array set? You have Product, but I was wondering whether there should be a more direct support of the array case - building it into the core API rather than by composition, like with BaseDistribution.

I don't have an answer for that. What can I suggest is that we start to introducing the domain of a distribution for easy distribution (Normal for example) and then we progress with more challenging one. In this way we can understand what we have to change to serve at best our final scope.

> finally, this is more of a crazy idea: should we decorate all functions with their range and domain? You could think of getattr(my_distr, "pdf") returning not a method but an object, like loc and iloc in pandas or BaseDistribution. This could have attributes, so doing sth like my_dist.pdf.domain and my_dist.pdf.range will return a set object. What do you think of it? Maybe it is overengineered, I would not do it right away anyway.

It is an interesting Idea. I'm thinking, probability density function, logarithmic probability function and cumulative density function have aways the same domain right? Therefore, this decorator it is not helpful to define domains for function associate to distributions (perhaps I'm wrong and there is some case where it can be used for that). On the other side, this could be a nice API for the final user :-) 


Should I finalise the class and then we can work on a branch where we add the domain/support for distribution already coded? What do you think?

fkiraly:
> I don't have an answer for that. What can I suggest is that we start to introducing the domain of a distribution for easy distribution (Normal for example) and then we progress with more challenging one. In this way we can understand what we have to change to serve at best our final scope.

Hm, ok. I suppose that works, incremental design is better than overdesign.

> It is an interesting Idea. I'm thinking, probability density function, logarithmic probability function and cumulative density function have aways the same domain right?

For the full domain, yes (the reals), but supports for the discrete part would be most interesting here, since that is not directly inspectable right now.

VascoSch92:
Hey @fkiraly 

I worked a little on the PR request and I think it is ready for a review.

Let me know what do you think.

Some remarks:
- the way I coded the unit tests can do better.
- what do you think about merging the module not in `main` but in another branch, such that we can experimenting? Once we have a stable API, we can merge in main.
- For instance, just finite, interval and product sets are implemented.

fkiraly:
Yes, I agree with developing this off `main` to a point where the functionality is integrated with the `BaseDistribution` class sufficiently.

I would suggest, let's make a branch on `sktime` called `sets`, I will base it off this branch, and we continue making PR to that branch?

Also, it would be nice to have a short write-up of your plans for the next steps, perhaps in the issue #244. I'm happy to halp with that or set up a meeting.

fkiraly:
done, there is now a branch called `sets` on `skpro`.

fkiraly:
> Should I close this PR and checkout the new branch?

I would say, let's leave this branch and PR as is, as it "freezes" the current state. Perhaps turn it into a draft.
The new PR will point to a branch that will change, so even if they are now identical they are not redundant in function.

Further, I think the tests are failing due to lack of `get_test_params` - perhaps you want to fix that to ensure `BaseObject` API compliance before we move on.

VascoSch92:
> > Should I close this PR and checkout the new branch?
> 
> I would say, let's leave this branch and PR as is, as it "freezes" the current state. Perhaps turn it into a draft. The new PR will point to a branch that will change, so even if they are now identical they are not redundant in function.
> 
> Further, I think the tests are failing due to lack of `get_test_params` - perhaps you want to fix that to ensure `BaseObject` API compliance before we move on.

ok perfect. This PR is draft now and I will open another PR into the other branch to fix the tests.

The tests were passing locally. I should give a look on how you implement the tests in the other modules.

fkiraly:
> The tests were passing locally. I should give a look on how you implement the tests in the other modules.

What were you trying for this? Have you tried `check_estimator` on the new objects?

VascoSch92:
> What were you trying for this? Have you tried `check_estimator` on the new objects?

I implemented just unit tests. I was just running them using pytest directly. Should I use the machinery developed for the `BaseObject`?

fkiraly:
> Should I use the machinery developed for the `BaseObject`?

It's automatically applied by the CI, and you can mimick that by using `check_estimator` on the objects. For the specific classes, you'd need implemented `get_test_params` to pass - or, defaults.

VascoSch92:
> > Should I use the machinery developed for the `BaseObject`?
> 
> It's automatically applied by the CI, and you can mimick that by using `check_estimator` on the objects. For the specific classes, you'd need implemented `get_test_params` to pass - or, defaults.

Sorry i really don't get 100% how it works. Do you have an example? I was trying to understand the tests for distribuitions but with not so much success.

fkiraly:
Sure! here is some explanation on the test classes:

https://www.sktime.net/en/latest/developer_guide/testing_framework.html#an-illustrative-example

Instances of objects are constructed from the parameter settings returned in `get_test_params`. If the latter contains no settings, and there are non-default params in `__init__`, the framework will not know how to construct an instance.

VascoSch92:
Is it mandatory to use this topology of testing if we are extending from `BaseObject` ?

fkiraly:
> Is it mandatory to use this topology of testing if we are extending from `BaseObject`?

Yes, in the sense that `BaseObject` descendants should be `BaseObject` API compliant, otherwise you lose the composability etc.
The requirement weight from this is typically minimal, such as specifying test parameters and not overriding methods.

VascoSch92:
> > Is it mandatory to use this topology of testing if we are extending from `BaseObject`?
> 
> 
> 
> Yes, in the sense that `BaseObject` descendants should be `BaseObject` API compliant, otherwise you lose the composability etc.
> 
> The requirement weight from this is typically minimal, such as specifying test parameters and not overriding methods.
> 
> 

Last question: there is a way to trigger the tests in the exact same way they are triggered in the workflow? Such that i can check that everything works before opening a Pr

fkiraly:
> Last question: there is a way to trigger the tests in the exact same way they are triggered in the workflow? Such that i can check that everything works before opening a Pr

Yes, running `pytest` on your local repository with no soft dependencies installed should trigger `run-tests-no-extras`.

Installing `skpro[all_extras]` and then triggering the tests should have the same effect as `run-tests-all-extras`.

Issue #229, opened 2024-03-31
[ENH] design discussion - `pdf` and `pmf` in distrubtions, discrete, continuous, and mixed


fkiraly:
This is a design discussion on how to handle `pdf` and `pmf` in distrubtions, which can be discrete, continuous (short for "absolutely continuous"), and mixed. Assuming domain on the real numbers, and distributions without singular component.

`scipy` handles these as follows:

* `pmf` is present and `pdf` is not present, for discrete distributions.
* `pdf` is present and `pmf` is not present, for continuous distributions.
* no support for mixed distributions.

I think it would be more consisten with composition a la `sklearn` if all distributions had all these methods, and they correspond to the measures in the Lebesgue decomposition. That is,

* `pmf` and `pdf` are present in all distributions
* the sum of measures implied by `pmf` and `pdf` is a probability measure

In particular, this would mean:

* for discrete distributions, `pmf` sums to one, and `pdf` is always zero
* for continuous distributions, `pdf` integrates to one, and `pmf` is always zero
* for mixed distribution, integral of `pdf` and sum of `pmf` sum to one. In general, the `pdf` integral, or `pmf` sum are not equal to one.

ShreeshaM07:
> Being faithful to the Lebesgue decomposition also has an advantage in mixtures: the `pdf` and `pmf` of a `m = Mixture([d1, d2], [w1, w2])` has `m.pdf = w1 * d1.pdf + w2 * d2.pdf`, and `m.pmf = w1 * d1.pmf + w2 * d2.pmf`, irrespective of components `d1`, `d2` being continuous, discrete, or mixed. (assuming `w1 + w2 == 1`).
> In a sense, this seems to be the convention that treats all edge cases consistently.

Yes, that is correct it will handle all edge cases irrespective of `d1`, `d2` being continuous, discrete or mixed as whenever the distribution becomes discrete the `pdf` integrates to 0 in that interval only the `pmf` will contribute in that interval.
And whenever the distribution becomes continuous in an interval the `pmf` sum will be 0 and only the `pdf` will contribute in that interval.
So in case of mixed interval `m.pdf = w1 * d1.pdf + w2 * d2.pdf`, and `m.pmf = w1 * d1.pmf + w2 * d2.pmf` will still be true.
And `m.pdf + m.pmf == 1` will also be true when we consider the whole interval ie (-inf, inf).

PR #295, opened 2024-05-03
[ENH] distributions: `pmf` and `log_pmf` method

### LINKED ISSUE #289: [ENH] Add Probability Mass Function (`pmf`) and Its Log (`log_pmf`) Methods on Base Distribution Class

All distribution classes inherit from the `BaseDistribution` class, both discrete and continuous distributions. Currently, the class `BaseDistribition` doesn't have `pmf` and `log_pmf` methods to calculate the value of the probability mass function and its logarithmic, but it has `pdf` and `log_pdf` methods. Probability Mass Function (PMF) and Probability Density Function (PDF) are a bit similar as they are both used to describe probability distributions. But, PMF is for discrete distribution whereas PDF is for continuous distribution.

In this situation, all discrete distribution classes inherit the PDF functionality which they shouldn't have and there are no base methods for PMF

The solution that I propose is that 
- We create a base class for each discrete and continuous distribution, that inherits from the `BaseDistribution` class. 
- Move the PDF functionality on the base continuous distribution class, so that discrete distributions don't need to implement it
- Add the PMF functionality on the base discrete distributions to accommodate this functionality for all discrete distribution classes

Another alternative that we can try is to directly add PMF functionality on the `BaseDistribution` class without adding base discrete and continuous distributions. This solution offers fewer changes but mixes the discrete and continuous functionalities on the base class

fkiraly:
This PR adds a `pmf` and `log_pmf` method to the base interface. Fixes #289

In accordance with #229, these return 0 resp `-np.inf` if the distribution is continuous.

Also makes the following, connected changes:
* `pdf` return 0 for discrete distributions
* removes the discrete/continuous handling logic from the `scipy` adapter, as this is now in the base class - FYI @malikrafsan, review appreciated!

I've also changed the way in which `TestScipyAdapter` queries the distributions - by inheritance, not by tag. This is since the tag is "mechanical" (for internal testing only) and it might confuse users to see a value in `object_type` which is not related to an external API property.
Would appreciate a comment on this too, e.g., if you think I am making the code or logic harder to understand or introducing unnecessary case distinctions.


fkiraly:
@malikrafsan, I think the remaining failures are due to checks in the case where `pdf` or `pmf` is identical zero. In this case, imo logarithm of 0 = `-np.inf` should be returned, the tests check for 0 though.

I think it's the tests that should change, not the distribution - no?

malikrafsan:
> In this case, imo logarithm of 0 = -np.inf should be returned, the tests check for 0 though.

Ahh yes, I think this is the correct one. Should I change the test directly on this PR or in another PR?

fkiraly:
I've changed it in this PR, if would be appreciated if you could have a look if it makes sense.

fkiraly:
(btw, are you on our discord? here: https://discord.com/invite/54ACzaFsn7
We are going to set up a workstream around `skpro` over the summer, with the mentoring programme, it would be great to have you around for that - and let us know if interested in joining the mentoring https://github.com/sktime/mentoring)

malikrafsan:
> are you on our discord?

Yes, but I haven't been active for quite a long time

> it would be great to have you around for that - and let us know if interested in joining the mentoring

Sure! It would be very beneficial for me to join this mentoring. I will try to catch up for a bit in the discord!

skpro/distributions/base/_base.py
```diff
@@ -529,6 +529,10 @@ def pdf(self, x):
         ``pd.DataFrame`` with same columns and index as ``self``
             containing :math:`p_{X_{ij}}(x_{ij})`, as above
         """
+        distr_type = self.get_tag("distr:measuretype", "mixed", raise_error=False)
```
malikrafsan, line 532: I am not really sure about this, but is it okay if we assume that the distribution is mixed (continuous and discrete)?

fkiraly: that happens only if the tag is not set.

Since the "exceptional" behaviour happens for continuous/discrete ("return 0" etc), this choice ensures that no exceptional behaviour happens, and we try to call the inner `_pdf` etc.

So, the purpose is not assuming that the distribution is mixed, but just for the purpose of calling the inner methods in all pdf/pmf related functions, since there is not enough information.

PR #298, opened 2024-05-04
[ENH] Beta Distribution

### LINKED ISSUE #22: [ENH] (wish)list of probability distributions to implement

malikrafsan:
#### Reference Issues/PRs

Towards #22

This PR implements Beta distribution based on Scipy Adapter

- [x] Whether the Scipy Adapter makes it easier to implement new distribution

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -42,6 +42,7 @@ Continuous support
     Normal
     TDistribution
     Weibull
+    Beta
 
 
 Integer support
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -35,6 +35,7 @@ Continuous support
     :toctree: auto_generated/
     :template: class.rst
 
+    Beta
     ChiSquared
     Fisk
     Laplace
@@ -42,7 +43,6 @@ Continuous support
     Normal
     TDistribution
     Weibull
-    Beta
 
 
 Integer support
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -3,6 +3,7 @@
 # adapted from sktime
 
 __all__ = [
+    "Beta",
     "ChiSquared",
     "Empirical",
     "Fisk",
@@ -19,7 +20,6 @@ __all__ = [
     "TDistribution",
     "Uniform",
     "Weibull",
-    "Beta",
 ]
 
 from skpro.distributions.beta import Beta
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -19,8 +19,10 @@ __all__ = [
     "TDistribution",
     "Uniform",
     "Weibull",
+    "Beta",
 ]
 
+from skpro.distributions.beta import Beta
 from skpro.distributions.chi_squared import ChiSquared
 from skpro.distributions.empirical import Empirical
 from skpro.distributions.fisk import Fisk
--- a/skpro/distributions/beta.py
+++ b/skpro/distributions/beta.py
@@ -1,0 +1,65 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Beta probability distribution."""
+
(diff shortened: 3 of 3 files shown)


docs/source/api_reference/distributions.rst
```diff
@@ -42,6 +42,7 @@ Continuous support
     Normal
     TDistribution
     Weibull
+    Beta
```
fkiraly, line 45: let's order these by alphabet

skpro/distributions/__init__.py
```diff
@@ -19,8 +19,10 @@
     "TDistribution",
     "Uniform",
     "Weibull",
+    "Beta",
```
fkiraly, line 22: let's order these by alphabet

skpro/distributions/beta.py
```diff
@@ -0,0 +1,65 @@
...
+
+import pandas as pd
+from scipy.stats import beta, rv_continuous
+
+from skpro.distributions.adapters.scipy import _ScipyAdapter
+
+
+class Beta(_ScipyAdapter):
+    r"""Beta distribution.
+
```
fkiraly, line 14: could you add a short note about parameterization? Have a look at the other distributions.

The reason is, many distributions have different parameterizations, e.g., normal can be parameterized by sigma or sigma-squared. It suffices to spell out one of the distribution defining functions, usually I choose the "simplest" for users to understand.

Let me know if you are unsure, then I will add it later.

fkiraly (changes requested):
Excellent - it looks like we have minimized the effort to add more distributions with the adapter!

I suppose this takes 5 or 10 min per distribution now?

I left a few comments.

fkiraly:
> Whether the Scipy Adapter makes it easier to implement new distribution

Well, you tell me - you implemented it, was it easier?

docs/source/api_reference/distributions.rst
malikrafsan: Fixed, thank you for pointing this out!

skpro/distributions/__init__.py
malikrafsan: Fixed, thank you for pointing this out!

skpro/distributions/beta.py
malikrafsan: Recently, I have added the parameterization note about the beta distribution. I had a look at the other distributions, and a lot of them are noted by PDF or CDF formula. That's why I tried to add the PDF formula as I think it is easier to understand, in the context of Beta Distribution. I took the formula from the [Wikipedia](https://en.wikipedia.org/wiki/Beta_distribution). I added the PDF formula as given

$$f(x) = \frac{x^{\alpha-1}(1-x)^{\beta-1}}{B(\alpha, \beta)}$$

Please let me know if there is something I missed! Also, I am totally fine as well if you want to modify the PR as well

malikrafsan:
> > Whether the Scipy Adapter makes it easier to implement new distribution
> 
> Well, you tell me - you implemented it, was it easier?

Sure! It was much easier compared to when I tried to implement the Weibull / LogLogistic distributions 😅

skpro/distributions/beta.py
fkiraly: yes, that works, especially with the explanation that B is for normalization.

fkiraly (approved):
All addressed!

Issue #290, opened 2024-05-02
[ENH] `ngboost` survival prediction model


fkiraly:
The PR https://github.com/sktime/skpro/pull/215 has added `ngboost` as an `skpro` regressor.

The package can also be invoked for survival prediction, i.e., for using censoring information `C` as in the `skpro` survival models.
We should also add a version of the `ngboost` algorithm for survival prediction, hence.

To avoid user confusion, it might be better to add this as a separate estimator in the `survival` module, although the two estimators may end up sharing a lot of similarities and hence, possibly, an inheritance tree.

ShreeshaM07:
Should I start working on this issue?

fkiraly:
If you find it interesting, of course!

Given that you've interfaced `ngboost` first time round, it should be less work for you than anyone.

PR #301, opened 2024-05-07
[ENH] Interface `NGBSurvival` in skpro

ShreeshaM07:
#### Reference Issues/PRs

Implements #290 

This interfaces the `NGBSurvival` which is part of the `ngboost` package for survival analysis.
No

If this is sufficient for all types of survival analysis that are planned. As mentioned in <a href=https://github.com/stanfordmlgroup/ngboost/blob/master/ngboost/ngboost.py>NGBoost</a> it does not support `interval-censored` survival analysis.
Yes

##### For new estimators

### DIFF
--- a/skpro/survival/gradient_boosting_surv/_ngboost_surv.py
+++ b/skpro/survival/gradient_boosting_surv/_ngboost_surv.py
@@ -1,0 +1,339 @@
+"""class for NGBoost probabilistic survival regression."""
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["ShreeshaM07"]
+
+import numpy as np
+
+from skpro.survival.base import BaseSurvReg
+
+
+class NGBoostSurvival(BaseSurvReg):
+    """Interface of NGBSurvival of ngboost in skpro.
+
+    NGBSurvival is a wrapper for the generic NGBoost class that
+    facilitates survival analysis.
+    Use this class if you want to predict an outcome that
+    could take an infinite number of
+    (ordered) values, but right-censoring is present in the observed data.
+
+    Parameters
+    ----------
+    dist : string , default = "LogNormal"
+        assumed distributional form of Y|X=x.
+        A distribution from ngboost.distns, e.g. LogNormal
+        Available distribution types
+        1. "LogNormal"
+    score : string , default = "LogScore"
+        rule to compare probabilistic predictions P̂ to the observed data y.
+        A score from ngboost.scores, e.g. LogScore
+    estimator : default learner/estimator: DecisionTreeRegressor()
+        base learner to use in the boosting algorithm.
+        Any instantiated sklearn regressor.
+    natural_gradient : boolean , default = True
+        whether natural gradient must be used or not.
+    n_estimators : int , default = 500
+        the number of boosting iterations to fit
+    learning_rate : float , default = 0.01
+        the learning rate
+    minibatch_frac : float, default = 1.0
+        the percent subsample of rows to
+        use in each boosting iteration
+    verbose : boolean, default=True
+        flag indicating whether output
+        should be printed during fitting
+    verbose_eval : int ,default=100
+        increment (in boosting iterations) at
+        which output should be printed
+    tol : float, default = 1e-4
+        numerical tolerance to be used in optimization
+    random_state : int, RandomState instance or None, optional (default=None)
+
+    Returns
+    -------
+        An NGBSurvival object that can be fit.
+    """
+
+    _tags = {
+        "authors": ["ShreeshaM07"],
+        "maintainers": ["ShreeshaM07"],
+        "python_dependencies": "ngboost",
+    }
+
+    def __init__(
+        self,
+        dist="LogNormal",
+        score="LogScore",
+        estimator=None,
+        natural_gradient=True,
+        n_estimators=500,
+        learning_rate=0.01,
+        minibatch_frac=1.0,
+        col_sample=1.0,
+        verbose=True,
+        verbose_eval=100,
+        tol=1e-4,
+        random_state=None,
+    ):
+        self.dist = dist
+        self.score = score
+        self.estimator = estimator
+        self.natural_gradient = natural_gradient
+        self.n_estimators = n_estimators
+        self.learning_rate = learning_rate
+        self.minibatch_frac = minibatch_frac
+        self.col_sample = col_sample
+        self.verbose = verbose
+        self.verbose_eval = verbose_eval
+        self.tol = tol
+        self.random_state = random_state
+
+        super().__init__()
+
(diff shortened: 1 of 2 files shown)


ShreeshaM07:
When I tried to clone the regressor like I did for NGBRegressor it started giving this
```
RuntimeError: Cannot clone object NGBSurvival(Dist=<class 'ngboost.distns.utils.SurvivalDistnClass.<locals>.SurvivalDistn'>, random_state=RandomState(MT19937) at 0x737553A8CB40), as the constructor either does not set or modifies parameter Dist
```
But when I ran the `check_estimator()` it passed all checks which was not the case in the other one.  @fkiraly  Could you please enlighten me as to why I did not have to clone it in this case.

fkiraly:
That is weird. All tests seem to pass.

Can you write down a full code block that causes the error, assuming this branch for imports etc?

skpro/survival/gradient_boosting_surv/_ngboost_surv.py
```diff
@@ -0,0 +1,339 @@
...
+        ----------
+        X : pandas DataFrame
+            feature instances to fit regressor to
+        y : pd.DataFrame, must be same length as X
+            labels to fit regressor to
+        C : pd.DataFrame, optional (default=None)
+            censoring information for survival analysis,
+            same length as X and y
+            should have entries 0 and 1 (float or int)
+            1 = uncensored, 0 = (right) censored
```
fkiraly, line 136: I think this is switched around.

The `skpro` interface has 1 = censored. So you need to do "one minus" if `C` is not `None`.

fkiraly: reason is, we need a unified API, censored always needs to be represented by the same number. Different packages do it differently, but from the `skpro` side we always need to do it the same way.

skpro/survival/gradient_boosting_surv/_ngboost_surv.py
```diff
@@ -0,0 +1,339 @@
...
+        """
+        import pandas as pd
+        from ngboost import NGBSurvival
+        from ngboost.scores import LogScore
+        from sklearn.tree import DecisionTreeRegressor
+
+        # C is a DataFrame with 1s if C is None
+        # uncensored = 1, (right) censored = 0
+        if C is None:
+            C = pd.DataFrame(np.ones(len(y)), index=y.index, columns=y.columns)
```
fkiraly, line 151: so, `else: C = 1 - C`

fkiraly: and a comment on how `ngboost` represents censoring, compared to `skpro`

fkiraly (changes requested):
Nice!

Left some comments above, some high-level comments:

* censoring is wrong way round, we should fix it
* there is substantial duplication with the ngboost regressor, I suggest thinking about how to deduplicate. Personally, I'd move common private methods into a mixin class (not a full adapter) in `regression.adapters.ngboost`, then both classes can inherit from that, and `BaseDistribution`.
* I would move the estimator either into `ensemble` or a separate `ngboost` folder. `ensemble` would be consistent with `sklearn` - let's do that, I will also move the `regression` estimators together there.

ShreeshaM07:
> * there is substantial duplication with the ngboost regressor, I suggest thinking about how to deduplicate. Personally, I'd move common private methods into a mixin class (not a full adapter) in `regression.adapters.ngboost`, then both classes can inherit from that, and `BaseDistribution`.

If I move the `_dist_to_ngboost_instance` and the `_ngb_dist_to_skpro` to a mixin class but in case of the regressors many of the distributions are allowed and the default is `Normal` whereas in case of Survival only `LogNormal` and `exponential` is done in ngboost so the default here would be `LogNormal` as `exponential` has not yet been implemented in skpro.

I am handling this default assignment in these methods itself so how do I distinguish as to whether it was called from the survival class or the regressor if I am going to do it this way.

skpro/survival/gradient_boosting_surv/_ngboost_surv.py
```diff
@@ -0,0 +1,339 @@
...
+            col_sample=self.col_sample,
+            verbose=self.verbose,
+            verbose_eval=self.verbose_eval,
+            tol=self.tol,
+            random_state=self.random_state,
+        )
+
+        # from sklearn.base import clone
+
+        # self.ngbsurv_ = clone(self.ngbsurv)
```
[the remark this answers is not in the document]
ShreeshaM07: Yes if i uncomment this line(and the import clone line) and rename `self.ngbsurv_ = NGBSurvival` to `self.ngbsurv = NGBSurvival` it gives the runtime error : cannot clone object.

It seems a bit odd indeed that `NGBRegressor` satisfies the `sklearn` contract whereas `NGBSurvival` does not.

fkiraly:
> I am handling this default assignment in these methods itself so how do I distinguish as to whether it was called from the survival class or the regressor if I am going to do it this way.

I see - I'd say this is possibly a case of the "generalize" pattern in refactoring then. To remove significant duplication you need to generalize the method slightly so it allows you to cover both "subcases" to run.
I have a very concrete idea how one could do this by adding one additional argument, but I do not want to prescribe a solution, since I would like to see whether you come up with the same (or a better one).

ShreeshaM07:
Please let me know if this would be sufficient I have also modified the NGBoostRegressor but not added it in this commit I will first pull the changes (shifted to ensemble) after this is merged and then make another PR for NGBoostRegressor.

ShreeshaM07:
> This looks good!
> 
> Though, should the regressor not also inherit from the new `NGBoostAdapter`? Jus checking, if you want to do this in a separate PR or if you have forgotten to commit changes.

I have only added the `NGBoostSurvival` here for the `NGBoostRegressor` since there were some changes made yesterday in #302 . I thought it would be better to pull the most recent updates and then modify `NGBoostRegressor` in a separate PR after this is merged. Do you want me to add it now itself ?

fkiraly (approved):
Ah, that makes sense.

Good call to avoid merge conflicts, I did not spot that.

Let's finalize the refactor then in another PR.

fkiraly:
PS, if you use "closing keywords" to link the issue at the start, see here - https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
then GitHub automatically links the issue to the PR, this is useful to close the issue with reference to PR automatically if the PR gets merged.

ShreeshaM07:
I have created a new PR #309 for the `NGBoostRegressor`.

Perhaps next we must implement or interface `exponential` distribution in `skpro` which can be used in `NGBoostSurvival` and ofcourse `NGBoostRegressor` too.

PR #302, opened 2024-05-07
[ENH] moving ensemble regressors to `regression.ensemble`

fkiraly:
This PR moves ensemble regressors into a new folder `regression.ensemble` - `BaggingRegressor` which previously was in a file of the same name, and `NGBoostRegressor` which was in `gradient_boosting`.

Requires no deprecation, since `NGBoostRegressor` was added since the last releas, and the import path for `BaggingRegressor` does not change.

Also adds a missing API ref link to `BaggingRegressor`.


fkiraly:
FYI @malikrafsan

fkiraly:
FYI @ShreeshaM07, for review - this is trying to mirror what is in `sklearn`'s `ensemble` module. Let me know if you disagree or have a better suggestion.

ShreeshaM07:
> FYI @ShreeshaM07, for review - this is trying to mirror what is in `sklearn`'s `ensemble` module. Let me know if you disagree or have a better suggestion.

No objections from my side please go ahead.

PR #303, opened 2024-05-08
[DOC] tutorial notebook for probability distributions

fkiraly:
Adds a tutorial notebook for probability distributions.

Section 3 is WiP, I plan adding it later.


VascoSch92:
Nice :-) Now it is super clear how distributions work.

Could you run the notebook? Such that it is also possible to see the output?

fkiraly:
> Could you run the notebook? Such that it is also possible to see the output?

Yes, I'll do this at the end though, since I am still working on it, and don't want to commit 100 versions of each picture.

fkiraly:
Btw, if you know how one could modify `plot` to allow plotting of multiple distributions in one place, I would be interested to hear.

Right now, if I reuse the `ax` object, it does not plot.

VascoSch92:
> Btw, if you know how one could modify `plot` to allow plotting of multiple distributions in one place, I would be interested to hear.
> 
> Right now, if I reuse the `ax` object, it does not plot.

I will give a look

fkiraly:
@VascoSch92, update:

* I ran the notebook now, it has images. I think I'm finished for now.
* I tried again plotting mulitple times in the `ax`, it seems to work. Bit odd, not sure why it did not worked before.

VascoSch92 (approved):

PR #320, opened 2024-05-12
[MNT] Deprecation message for `CyclicBoosting` changes

setoguchi-naoki:
#### Reference Issues/PRs

#232

- Add deprication warning
- Docstring
No

No
I'm not familiar with the detail for deprecation process and which future version we should set for removing default parameters.

No

##### For all contributions

##### For new estimators


skpro/distributions/qpd.py
```diff
@@ -160,6 +178,13 @@ def mean(self, lower=0.0, upper=np.inf):
         pd.DataFrame with same rows, columns as `self`
         expected value of distribution (entry-wise)
         """
+        #  TODO <X.X.X>: remove warning
+        comment = "Parameter 'lower' and 'upper' will be moved to \
+                   class parameters in version X.X.X. \
+                   To retain prior behaviour, set 'lower' to 0.0 \
+                   and 'upper' to large number explicitly"
```
fkiraly, line 185: should this not be `None` instead of a large number?

fkiraly (changes requested):
I think this is good, given the changes.

I would improve it as follows:

* the message in `__init__` needs only be raised if the changing parameters are invoked with a non-default, since that is the only case where behaviour changes or can break.

The parameters in the methods imo can be removed without deprecation message, as they were not API compliant

skpro/distributions/qpd.py
setoguchi-naoki: It should be a large number here because 'upper' is used as the range of the differential. However, I made a mistake in the docstring for the default value of 'upper ', it is not None but 1e3 in the future version. In this case too, the large number should be set to the 'upper' parameter to cover the range of the user's dataset.

setoguchi-naoki:
> the message in __init__ needs only be raised if the changing parameters are invoked with a non-default, since that is the only case where behaviour changes or can break.

Thanks, I modified as following your comment. the default values of 'lower' and 'upper' will be removed in the future version. I made a wrong deprecation message. It is also modified. If the default value is removed, a deprecation message is always raised, as this can break the user's code.

> The parameters in the methods imo can be removed without deprecation message, as they were not API compliant

Sorry, I'm not sure I understand your comment. Which method did you mention?

setoguchi-naoki:
In addition, what future version should we specify in deprecation message?

fkiraly:
2.4.0 - the second next MINOR release.

I've opened a new PR here that covers only the regressor class: https://github.com/sktime/skpro/pull/329
The distribution should be covered by this: #254

fkiraly:
If we want to make further changes, these will have to wait until 2.5.0, as we will now proceed to the 2.3.0 release shortly.

The changes in #329 or #327 should not prevent us from making further changes to the distribution interface.

setoguchi-naoki:
Thank you for your contribution. As there are no tasks left for me to do, I have no problem closing this PR if it is unnecessary.

fkiraly:
Well, I dealt only with the `CyclicBoosting` class for deprecation.

We may stlil have to decide what to do with the distributions, in deprecation. For instance:

* all QPD_X distributions have parameters `lower` and `upper`. Are these needed? Or should we deprecate them? I think @felixwick mentioned this in his review of #327, and I replied that I left these because they were in the initial implementation.
* what about `dist_shape`?
* do we want to allow arbitrary distributions, incl `scipy` and `skpro` distributions as base distributions in QPD?

My advice would hence be, let's be clear about the target interfaces, from that it will become clear whether we need any further change/deprecation management, or not.

setoguchi-naoki:
> all QPD_X distributions have parameters lower and upper. Are these needed? Or should we deprecate them? I think @FelixWick mentioned this in his review of https://github.com/sktime/skpro/pull/327, and I replied that I left these because they were in the initial implementation.

As mentioned in #327 , the lower and upper parameters are not necessarily needed in all QPD classes if findiff is not used. Therefore, the upper parameter can be removed in QPD_S and the lower and upper parameters in QPD_U.

> what about dist_shape?

The dist_shape parameter is an option for fine-tuning the QPD_U. Therefore, it can be removed as a deprecated parameter if simplification of the QPD class interface is a priority. How do you think?

> do we want to allow arbitrary distributions, incl scipy and skpro distributions as base distributions in QPD?

It would be good to be able to specify scipy and skpro distributions, as you have implemented in #327. In this case, I don't think some support is needed in the deprecated process.

Let me check the situation. Should I add a deplication message to this PR to meet the #327 specification? Is there anything else I should consider?

FelixWick:
The shape parameter is only used for the extended QPD versions (Which include QPD_U. That means only QPD_S and QPD_B are "original" ones.), which I suggest to remove here.

fkiraly:
> As mentioned in https://github.com/sktime/skpro/pull/327 , the lower and upper parameters are not necessarily needed in all QPD classes if findiff is not used. Therefore, the upper parameter can be removed in QPD_S and the lower and upper parameters in QPD_U.

Ok, that makes sense, I was wondering what these were used for, already.

The need to use `findiff` is completely removed now, since the current native implementation computes `pdf` explicitly (just use the chain rule 😁 ). So let's deprecate the unused parameters then.

> The dist_shape parameter is an option for fine-tuning the QPD_U. Therefore, it can be removed as a deprecated parameter if simplification of the QPD class interface is a priority. How do you think?

I agree it should be removed. I would actually replace it by:

* renaming `version` to a more appropriate name, e.g., `base_dist`
* allowing `base_dist` to be a parametric `skpro` distirbution, which can have its own parameters. If users want to pass `dist_shape`, they can pass it as a parameter to `base_dist`.

fkiraly:
> Let me check the situation. Should I add a deplication message to this PR to meet the https://github.com/sktime/skpro/pull/327 specification? Is there anything else I should consider?

This might require extending the logic, at least addition of a test case that passes an `skpro` distirbution.

I would also strongly suggest to rename the parameter `version`, as it has strong connotations with versioning, e.g., semver.

setoguchi-naoki:
@fkiraly 
I modfied deplication message like

- Removal of uppers and dist_shape from some QPD classes
- Rename version to base_dist

> at least addition of a test case that passes an skpro distirbution.

The `skpro` distribution has not yet been introduced and tested in qpd.py. Should it be done in other branches?

fkiraly:
Great!

I think you're quite out of sync with your branch since the new release - can you merge `main` into your branch, and ensure to resolve all conflicts? Only the files related to `CyclicBootsing` should be changed.

setoguchi-naoki:
I think I have resolved the conflicts in the main branch.

fkiraly:
> I think I have resolved the conflicts in the main branch.

👍 yes, looks like it!

fkiraly (changes requested):
Great!

One minor thing, we should ensure that users can already move to the future behaviour, i.e., have a chance to update their code in advance. That would mean adding the `base_dist` parameter already right now, an dif it is passed, it overrides `version`.

Have a look here how renaming a parameter works, including a suggested warning message: https://www.sktime.net/en/stable/developer_guide/deprecation.html#id2

setoguchi-naoki:
Thank you. Fixes for parameter renaming have been applied.

fkiraly (approved):
Made some minor tweaks around the docstrings (moved to the "final" form), and release manager actions.

PR #325, opened 2024-05-14
[ENH] `Exponential` distribution

ShreeshaM07:

This interfaces the `Exponential` distribution using the `_ScipyAdapter`.

No

Yes

##### For all contributions

##### For new estimators
(https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

### DIFF
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -37,13 +37,13 @@ Continuous support
 
     Beta
     ChiSquared
+    Exponential
     Fisk
     Laplace
     Logistic
     Normal
     TDistribution
     Weibull
-    Exponential
 
 
 Integer support
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -43,6 +43,7 @@ Continuous support
     Normal
     TDistribution
     Weibull
+    Exponential
 
 
 Integer support
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -7,6 +7,7 @@ __all__ = [
     "ChiSquared",
     "Delta",
     "Empirical",
+    "Exponential",
     "Fisk",
     "Laplace",
     "Logistic",
@@ -21,7 +22,6 @@ __all__ = [
     "TDistribution",
     "Uniform",
     "Weibull",
-    "Exponential",
 ]
 
 from skpro.distributions.beta import Beta
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -21,12 +21,14 @@ __all__ = [
     "TDistribution",
     "Uniform",
     "Weibull",
+    "Exponential",
 ]
 
 from skpro.distributions.beta import Beta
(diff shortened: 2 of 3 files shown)


skpro/distributions/exponential.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, mu, scale, index=None, columns=None):
+        self.mu = mu
+        self.scale = scale
+
+        super().__init__(index=index, columns=columns)
+
+    def _get_scipy_object(self) -> rv_continuous:
+        return expon
+
+    def _get_scipy_param(self) -> dict:
```
VascoSch92, line 47: return type here is a `dict`, but then you are returning a `tuple` composed of a `list` and a `dict`

skpro/distributions/exponential.py
```diff
@@ -0,0 +1,76 @@
...
+    scale :math:`b`, such that the pdf is
+
+    .. math:: f(x) = \exp\left(-\frac{x - \mu}{b}\right)
+
+    The mean :math:`\mu` is represented by the parameter ``mu``,
+    and the scale :math:`b` by the parameter ``scale``.
+
+    Parameter
+    ---------
+    mu : float or array of float (1D or 2D)
```
fkiraly, line 25: I would give this a single parameter, `rate`. The `scipy` parameter `scale` is then the inverse of `rate`.

skpro/distributions/exponential.py
ShreeshaM07: Indeed the return type must be changed.

skpro/distributions/exponential.py
ShreeshaM07: I think the exponential distribution takes 2 parameters `loc` and `scale`. If I am not wrong how would having only 1 parameter `rate` work as the distribution is technically `e^-((x-loc)/scale)`.

skpro/distributions/__init__.py
```diff
@@ -21,12 +21,14 @@
     "TDistribution",
     "Uniform",
     "Weibull",
+    "Exponential",
```
fkiraly, line 24: can you put this to be alphabetically ordered?

docs/source/api_reference/distributions.rst
```diff
@@ -43,6 +43,7 @@ Continuous support
     Normal
     TDistribution
     Weibull
+    Exponential
```
fkiraly, line 46: can you put this to be alphabetically ordered?

skpro/distributions/exponential.py
fkiraly: I would suggest, let's use the same parameters as on wikipedia:

https://en.wikipedia.org/wiki/Exponential_distribution

We cannot use `lambda` as it is a reserved python keyword, so `rate` is the next best thing as it is consistent with Poisson process terminology.

ShreeshaM07:
I've made the changes accordingly please let me know if anything else needs to be done. I shall create PR with the updated `ngboost` files as well.

fkiraly (approved):
Looks good!

PR #327, opened 2024-05-15
[ENH] native implementation of Johnson QPD family, explicit pdf

fkiraly:
This PR reworks the family of QPD family of distributions for efficiency and to allow removal of the newly introduced dependency `findiff` in #232.

The dependency `findiff` was introduced for approximation of `pdf`, but in fact it is unnecessary as the `pdf` can be analytically derived by applying the chain rule. True, it has to be applied three or four times, but it's still the chain rule... efficiency and accuracy gains are significant, and it helps us avoid computing numerical derivatives for all entries in a large matrix, together with the now unnecessary `findiff` dependency.

Makes the following changes:

* refactoring of the three QPD distributions tp use `skpro` machinery:
    * use of the `skpro` native parameter broadcasting system instead of ad-hoc broadcasting
    * use of the `skpro` native approximation for `mean`, `var`, instead of three copies of similar (and partially duplicative) approximation inside the distributions
* refactoring between the three QPD distributions with the end of simplification
    * refactoring QPD parameter computation into a single, fully vectorized function, `_prep_qpd_vars`
* clean room reimplementation of `cdf`, `ppf` of the three distributions based on the `cyclic_boosting` reference
* new implementation of `pdf`, as derivative of `cdf`

As side effects of the rework:
* all parameters now broadcast in numpy-like fashion, including `alpha`, `lower`, `upper`, which previously was not possible
* the distributions can be 2D with more than 1 column, which previously was not possible
* `version` (the base distribution) can now be an arbitrary continuous `scipy` distribution
* `pdf` is numerically exact
* the distributions do not have soft dependencies anymore

Regarding the relation to `cyclic_boosting`:

* this is clean room reimplementation and credit is given, so I hope this is fine license-wise - @felixwick?
* this is the result of trying to remove the `findiff` dependency for computing the `pdf` from the `cdf` that was introduced in #232, as well as cleanup before release. I ended up simplifying a lot, ending up here. In this sense, the work of @setoguchi-naoki was crucial in arriving at this point.
* I would have no issue at all with you moving the improved code into `cyclic_boosting`. We can even restore the dependency and maintain the distribution logic in `cyclic_boosting` if that were your preference, e.g., for ownership reasons.

FYI @felixwick, @setoguchi-naoki, review appreciated - I would like to remove the `findiff` dependency before the next release.

### DIFF
--- a/skpro/distributions/qpd.py
+++ b/skpro/distributions/qpd.py
@@ -272,17 +272,17 @@ class QPD_S(BaseDistribution):
         phi = _resolve_phi(version)
         self.phi = phi
 
-        params = _prep_qpd_vars(phi=phi, mode="S", **self._bc_params)
-        self.params = params
+        qpd_params = _prep_qpd_vars(phi=phi, mode="S", **self._bc_params)
+        self._qpd_params = qpd_params
 
     def _ppf(self, p: np.ndarray):
         """Quantile function = percent point function = inverse cdf."""
         lower = self._bc_params["lower"]
-        delta = self.params["delta"]
-        kappa = self.params["kappa"]
-        c = self.params["c"]
-        n = self.params["n"]
-        theta = self.params["theta"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        theta = self._qpd_params["theta"]
 
         phi = self.phi
 
@@ -295,11 +295,11 @@ class QPD_S(BaseDistribution):
     def _pdf(self, x: np.ndarray):
         """Probability density function."""
         lower = self._bc_params["lower"]
-        delta = self.params["delta"]
-        kappa = self.params["kappa"]
-        c = self.params["c"]
-        n = self.params["n"]
-        theta = self.params["theta"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        theta = self._qpd_params["theta"]
 
         phi = self.phi
 
@@ -325,11 +325,11 @@ class QPD_S(BaseDistribution):
     def _cdf(self, x: np.ndarray):
         """Cumulative distribution function."""
         lower = self._bc_params["lower"]
-        delta = self.params["delta"]
-        kappa = self.params["kappa"]
-        c = self.params["c"]
-        n = self.params["n"]
-        theta = self.params["theta"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        theta = self._qpd_params["theta"]
 
         phi = self.phi
 
@@ -455,18 +455,18 @@ class QPD_B(BaseDistribution):
         phi = _resolve_phi(version)
         self.phi = phi
 
-        params = _prep_qpd_vars(phi=phi, mode="B", **self._bc_params)
-        self.params = params
+        qpd_params = _prep_qpd_vars(phi=phi, mode="B", **self._bc_params)
+        self._qpd_params = qpd_params
 
     def _ppf(self, p: np.ndarray):
         """Quantile function = percent point function = inverse cdf."""
         lower = self._bc_params["lower"]
-        rnge = self.params["rnge"]
-        delta = self.params["delta"]
-        kappa = self.params["kappa"]
-        c = self.params["c"]
-        n = self.params["n"]
-        xi = self.params["xi"]
+        rnge = self._qpd_params["rnge"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        xi = self._qpd_params["xi"]
 
         phi = self.phi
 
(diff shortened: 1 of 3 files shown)


FelixWick:
License-wise no concerns from my side.
I'm also fine with keeping the code in this repository.
I will have a closer look at the code tomorrow.

fkiraly:
Looking forward to the review!

This PR should be considered together with https://github.com/sktime/skpro/pull/330, which implements an improved integral heuristic for `mean`, `var`.

I think the heuristic in #330 is better than what you had in the distribution (first take derivative of cdf to obtain pdf, then integrate against x) and the current base class heuristic (Monte Carlo based on `sample`).

fkiraly:
merging for the release to avoid `findiff` dependency - please let me know if we should change anything, we can do so for the next release or, if necessary, in a hotfix.

skpro/distributions/qpd.py
```diff
@@ -263,138 +265,78 @@ def __init__(
...
-        self.qpd = J_QPD_S(
-            alpha=alpha,
-            qv_low=qv_low,
-            qv_median=qv_median,
-            qv_high=qv_high,
-            l=self.lower,
-            version=version,
-        )
-        super().__init__(index=index, columns=columns)
+        in_sinh = np.arcsinh(phi.ppf(p) * delta) + np.arcsinh(n * c * delta)
```
FelixWick, line 288: In most of the cases, you probably want to do a np.outer rather than an element-wise multiplication here. (In the Cyclic Boosting implementation I made that optional.)

skpro/distributions/qpd.py
```diff
@@ -263,138 +265,78 @@ def __init__(
...
+        lower = self._bc_params["lower"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        theta = self._qpd_params["theta"]
+
+        phi = self.phi
+
+        in_arcsinh = np.log((x - lower) / theta) / kappa
```
FelixWick, line 335: In most of the cases, you probably want to do a np.outer rather than an element-wise multiplication here. (In the Cyclic Boosting implementation I made that optional.)

skpro/distributions/qpd.py
```diff
@@ -503,137 +453,80 @@ def __init__(
...
-        upper = params["upper"]
-        index = params["index"]
-        mean = self.mean().values
-        x = np.linspace(lower, upper, num=int(1e3))
-        cdf = self.qpd.cdf(x)
-        if cdf.ndim < 2:
-            cdf = cdf[:, np.newaxis]
-        var = var_func(x, mean, cdf.T, index.shape[0])
-        return var
+        in_cdf = xi + kappa * np.sinh(delta * (phi.ppf(p) + n * c))
```
FelixWick, line 477: In most of the cases, you probably want to do an outer product (e.g., by means of p[:, np.newaxis]) rather than an element-wise multiplication here. (In the Cyclic Boosting implementation I made that optional.)

skpro/distributions/qpd.py
```diff
@@ -503,137 +453,80 @@ def __init__(
...
+        rnge = self._qpd_params["rnge"]
+        delta = self._qpd_params["delta"]
+        kappa = self._qpd_params["kappa"]
+        c = self._qpd_params["c"]
+        n = self._qpd_params["n"]
+        xi = self._qpd_params["xi"]
+
+        phi = self.phi
+
+        phi_ppf = phi.ppf((x - lower) / rnge)
```
FelixWick, line 525: In most of the cases, you probably want to do an outer product (e.g., by means of p[:, np.newaxis]) rather than an element-wise multiplication here. (In the Cyclic Boosting implementation I made that optional.)

FelixWick:
Drop the sinhlogistic version from the docstring, it was only used in the extended QPD versions (which we dropped).

FelixWick:
There should be no upper bound for QPD_S.

FelixWick:
I would suggest to drop the QPD_U, because this is one of the extended QPD versions. I'm not sure it works properly in the given implementation.

skpro/distributions/qpd.py
```diff
@@ -503,137 +453,80 @@ def __init__(
...
-        ppf_all = ppf_func(p_unique, self.qpd)
-        ppf_map = np.tile(p_unique, (qv_low.size, 1)).T
-        ppf = np.zeros((index.shape[0], len(columns)))
-        for r in range(p.shape[0]):
-            for c in range(p.shape[1]):
-                t = np.where(ppf_map[:, c] == p[r][c])
-                ppf_part = ppf_all[t][c]
-                ppf[r][c] = ppf_part
-        return ppf
+        # we work through the chain rule for the entire nested expression in cdf
```
FelixWick, line 493: I didn't check the math, but I assume you have compared the result with the numeric version?

FelixWick (commented):
I didn't check all details of the reimplementation, but I assume you compared results to make sure it is consistent. Looks good in general. A few comments inline.

skpro/distributions/qpd.py
fkiraly: not in most cases imo, but I do see how this would be useful in many cases.

This is, however, a framework level concern, so I moved it into a new issue: https://github.com/sktime/skpro/issues/341

Btw, if `np.ndarray` of dim 3 could be passed, the `numpy` native upcasting would already provide the kind of behaviour that is described in #341.

fkiraly:
> There should be no upper bound for QPD_S.

The upper bound was already there, in the distribution initially implemented. We can of course remove it, as I am not sure where it is used. I left it since I was apprehensive of breakages coming from public access of the interface, we will have to remove via deprecation process.

fkiraly:
> I didn't check all details of the reimplementation, but I assume you compared results to make sure it is consistent. 

Yes, spot checks only though.

There is a test that checks against the case where we expect a uniform distribution.

setoguchi-naoki:
> > There should be no upper bound for QPD_S.
> 
> The upper bound was already there, in the distribution initially implemented. We can of course remove it, as I am not sure where it is used. I left it since I was apprehensive of breakages coming from public access of the interface, we will have to remove via deprecation process.

This was introduced for the use of findiff; it can be removed if findiff is not required.

PR #330, opened 2024-05-15
[ENH] improved defaults for `BaseDistribution` `_mean`, `_var`, and `_energy_x`

fkiraly:
This PR adds improved defaults for `BaseDistribution` `_mean`, `_var`, and `_energy_x`.

The current defaults are Monte Carlo estimates for these quantities, using `sample`.

A faster and more accurate estimtate is obtained from integrating the `ppf`, which can be done only if it is available.

An example is the identity $\mathbb{E}[X] = \int_{0}^1 F_X^{-1}(x) \mbox{ } dx$, which is true for any real random variable `X`. Similar identities can be derived for variance and energy cross-term, which are both expectations in a single copy of `X`.

one could also do sth similar for gridding or triangulating the self-term, but it is not obvious whether that is better than sthe uniform sampling strategy, so it is not included in this PR.

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -1138,21 +1138,6 @@ class BaseDistribution(BaseObject):
         2D np.ndarray, same shape as ``self``
             energy values w.r.t. the given points
         """
-        approx_spl_size = self.get_tag("approx_energy_spl")
-        if x is not None and self._has_implementation_of("_ppf"):
-            approx_method = (
-                "by approximating the energy expectation by the integral "
-                "of the absolute difference of x to the ppf,"
-                f"with {approx_spl_size} equidistant nodes"
-            )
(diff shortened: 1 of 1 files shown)


fkiraly:
Code to test speed and accuracy:

```
from skpro.distributions.tests.test_base_default_methods import _DistrDefaultMethodTester
from time import time

import numpy as np

mu = np.random.normal(size=(1000, 100))
sigma = np.abs(np.random.normal(size=(1000, 100)))

d = _DistrDefaultMethodTester(mu, sigma)
start = time()
mu_hat = d.mean()
duration = time() - start

err = np.mean(np.abs(mu_hat-mu))
```

On this branch: 5.3 sec, 6.4e-16 error.
Current `main`: 9.2 sec, 2.0e-2 error.
(both on my local system)

That is, slight improvement in runtime; massive improvement in accuracy.

PR #332, opened 2024-05-16
[MNT] `Exponential` dist in `NGBoostRegressor`, `NGBoostSurvival`

ShreeshaM07:

This adds exponential distribution to `NGBoostRegressor` and `NGBoostSurvival`.

No

Yes

##### For all contributions

### DIFF
--- a/skpro/regression/adapters/ngboost/_ngboost_proba.py
+++ b/skpro/regression/adapters/ngboost/_ngboost_proba.py
@@ -32,7 +32,7 @@ class NGBoostAdapter:
         -------
         NGBoost Distribution object.
         """
-        from ngboost.distns import Laplace, LogNormal, Normal, Poisson, T
+        from ngboost.distns import Exponential, Laplace, LogNormal, Normal, Poisson, T
 
         ngboost_dists = {
             "Normal": Normal,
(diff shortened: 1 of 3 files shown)


fkiraly (approved):
Nice!

I think we can merge like this, but I am noticing a duplication in `_predict_proba`, namely the logic to translate `ngboost` distirbution parameters to `skpro` distribution parameters is duplicated. How about moving these into the adapter, too, to have only one location?

Not a blocker, let me know if you would like to do it in tihs PR, another one, or not at all.

ShreeshaM07:
I have moved most of the common code to the adapter now I hope that looks fine.

fkiraly:
> I have moved most of the common code to the adapter now I hope that looks fine.

The way I'd do it, I'd remove the dependency on `X` and calls of `self._pred_dict`, and instead have `index`, `columns` as args.

fkiraly:
That's what I meant, excellent!

ShreeshaM07:
> The way I'd do it, I'd remove the dependency on `X` and calls of `self._pred_dict`, and instead have `index`, `columns` as args, as well as the return of `_pred_dict`.

Yeah that makes more sense.

fkiraly (approved):

PR #350, opened 2024-05-21
[ENH] sets module - skbase compatibility

VascoSch92:
I'm sorry. 

I did some mistake with the new and old branch. 

The easiest way was to create a new branch and open a PR. 

I will close the other ones.

### DIFF
--- a/skpro/domains/domains.py
+++ b/skpro/domains/domains.py
@@ -126,6 +126,7 @@ class Interval(Domain):
         """Return the boundary of the interval, i.e., the extremities."""
         return self._left, self._right
 
+    @classmethod
     def get_test_params(cls, parameter_set="default"):
         """Return testing parameter settings for the estimator."""
         return {
@@ -182,6 +183,7 @@ class Finite(Domain):
         """Return the boundary of the finite set, i.e., the finite set itself."""
         return tuple(sorted(self.values))
 
+    @classmethod
     def get_test_params(cls, parameter_set="default"):
         """Return testing parameter settings for the estimator."""
         return {
@@ -260,6 +262,7 @@ class Product(Domain):
         """
         return tuple(element.boundary for element in self.product)
 
+    @classmethod
     def get_test_params(cls, parameter_set="default"):
         """Return testing parameter settings for the estimator."""
         return {
--- a/skpro/domains/domains.py
+++ b/skpro/domains/domains.py
@@ -168,6 +168,8 @@ class Finite(Domain):
                 raise TypeError(f"Expected `float`or `int`, but got {type(value)}.")
             if value in [-float("inf"), float("inf")]:
                 raise ValueError(f"Value {value} not accepted in finite set.")
+        if len(values) != len(set(values)):
+            raise ValueError(f"Detected duplicated values in {values}!")
         return list(set(values))
 
     def __contains__(self, item) -> bool:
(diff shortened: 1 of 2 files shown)


fkiraly:
No need to be sorry.

This is git: https://xkcd.com/1597/

skpro/domains/domains.py
```diff
@@ -126,6 +126,13 @@ def boundary(self) -> Tuple[float, float]:
         """Return the boundary of the interval, i.e., the extremities."""
         return self._left, self._right
 
+    def get_test_params(cls, parameter_set="default"):
```
fkiraly, line 129: I think you are missing the `@classmethod` decorator, then it should work.

VascoSch92: added

fkiraly: restarted tests

fkiraly:
Made a few changes to ensure `skbase` compatibility.

FYI, the API convention is that all classes must behave like dataclasses:

* init params are written to `self` in the constructor
* over the lifetime of the object, they are never mutated, only read

skpro/domains/domains.py
```diff
@@ -126,6 +129,14 @@ def boundary(self) -> Tuple[float, float]:
         """Return the boundary of the interval, i.e., the extremities."""
         return self._left, self._right
 
+    @classmethod
+    def get_test_params(cls, parameter_set="default"):
+        """Return testing parameter settings for the estimator."""
+        return {
+            "values": [1, 2],
+            "parenthesis": "open",
```
VascoSch92, line 137:
```suggestion
            "_parenthesis": "open",
```

VascoSch92:
> Made a few changes to ensure `skbase` compatibility.
> 
> FYI, the API convention is that all classes must behave like dataclasses:
> 
> * init params are written to `self` in the constructor
> * over the lifetime of the object, they are never mutated, only read

Ok.

Now the testing are failing because of this error:
```
ERROR skpro/tests/test_all_estimators.py::TestAllObjects::test_repr[Product] - AttributeError: 'Interval' object has no attribute 'parenthesis'
```
You add the attribute `_parenthesis`. Therefore, i changed in the get_test_param the key `parenthesis` to `_parenthesis`.

I hope i correct :-)

fkiraly:
Ah, that was my mistake! The `parenthesis` arg should be always what was passed to `__init__`, so the "resolve" needs to write to another arg. The `get_test_params` reference still should be `"parenthesis"`  - I fixed it and it should work now.

VascoSch92:
Still an error but i think i know what is. I will give a look

fkiraly:
I see, this raises an interesting question.

Do we want to enforce the dataclass like API from `BaseObject` on the init args that get simplified?

VascoSch92:
Good question.

I don't have a strong opionion.

What are the pros to use it?

fkiraly:
> What are the pros to use it?

Immutability of the blueprint - you can identify the object you are dealing with by the class and params uniquely. The usual case this guards against is change of the object blueprint over its lifetime - but not "simplification" operations, as in this case.

VascoSch92:
If we want to have it, i can change the class such that it thrown an error if you are creating a finite sets with repetition. Let me know what you orefer ;)

I can change it tomorrow and start to work with distribuitions ;)

fkiraly:
Hm, I'd be fine with either solution here. Perhaps raising an error if not unique makes most sense.

fkiraly (approved):
it does!

VascoSch92:
Can we merge such that I can start with domains and distributions ? :-)

PR #355, opened 2024-05-23
[ENH] Gamma Distribution

ShreeshaM07:

Implements Gamma distribution.
No

Yes

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/gamma.py
+++ b/skpro/distributions/gamma.py
@@ -1,0 +1,78 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Exponential probability distribution."""
+
+__author__ = ["ShreeshaM07"]
+
+import pandas as pd
+from scipy.stats import gamma, rv_continuous
+
+from skpro.distributions.adapters.scipy import _ScipyAdapter
+
+
+class Gamma(_ScipyAdapter):
+    r"""Gamma Distribution.
+
+    The Gamma Distribution is parameterized by shape :math:`\alpha` and
+    rate :math:`\beta`, such that the pdf is
+
+    ..math:: f(x) = \frac{x^(\alpha-1)*\exp\left(-\beta*x)*\beta^\alpha}{\tau(\alpha)}
+
+    where `\tau(\alpha)` is the Gamma function.
+    For all positive integers,`\tau(\alpha) = (alpha-1)!`.
+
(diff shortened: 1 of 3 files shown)


skpro/distributions/gamma.py
```diff
@@ -0,0 +1,78 @@
...
+class Gamma(_ScipyAdapter):
+    r"""Gamma Distribution.
+
+    The Gamma Distribution is parameterized by shape :math:`\alpha` and
+    rate :math:`\beta`, such that the pdf is
+
+    ..math:: f(x) = \frac{x^(\alpha-1)*\exp\left(-\beta*x)*\beta^\alpha}{\tau(\alpha)}
+
+    where `\tau(\alpha)` is the Gamma function.
+    For all positive integers,`\tau(\alpha) = (alpha-1)!`.
```
VascoSch92, line 21:
```suggestion
    where :math:`\tau(\alpha)` is the Gamma function.
    For all positive integers, :math:`\tau(\alpha) = (\alpha-1)!`.
```

fkiraly (approved):
Looks good to me.

Typos in docstring should be fixed, though are not blocking.

skpro/distributions/gamma.py
ShreeshaM07: Could you please be so kind to write the expression for the `f(x)` as well in this LaTex format I am not able to do it with trial and error, or maybe lead me to some resource that explains the usage.

fkiraly: but `f` looks correct already - does it not compile to sth readable?

You can always test it by putting it in markdown, like this:

$`f(x) = \frac{x^{\alpha-1}\exp\left(-\beta x\right)\beta^{\alpha}}{\tau(\alpha)}`$

ShreeshaM07: for some reason the expression is working when i use $$ in markdown here but it doesnt convert in readthedocs and instead if i use ..math:: neither here nor there it is readable.

ShreeshaM07:
Finally it works! please let me know if anything else needs to be done.

skpro/distributions/gamma.py
fkiraly: I'm looking at the docs, it is perfectly readable for me:

image

fkiraly:
oh, I see - you fixed it. Great!

PR #356, opened 2024-05-23
[ENH] Alpha distribution

SaiRevanth25:
 Implements Alpha Distribution. Towards https://github.com/sktime/skpro/issues/22

No

 No

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/alpha.py
+++ b/skpro/distributions/alpha.py
@@ -17,6 +17,18 @@ class Alpha(_ScipyAdapter):
     It is often used for modeling data with heavy right tails,
     unlike the Gaussian distribution(which is symmetric and bell-shaped).
 
+    The probability density function (PDF) of the Alpha distribution is given by:
+    .. math::
+
+        f(x) = \frac{1}{x^2 \Phi(a) \sqrt{2\pi}}
+                \exp\left(-\frac{1}{2}\left(\frac{a - 1}{x}\right)^2\right)
+
+    where:
+       - :math:`a` is the shape parameter.
+       - :math:`Phi` is the cumulative distribution function (CDF) of the
+                    standard normal distribution.
+
+
     Parameters
     ----------
     a : float or array of float (1D or 2D), must be positive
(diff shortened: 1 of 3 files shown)


skpro/distributions/alpha.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self, a, index=None, columns=None):
+        self.a = a
+
+        super().__init__(index=index, columns=columns)
+
+    def _get_scipy_object(self) -> rv_continuous:
+        return alpha
+
+    def _get_scipy_param(self) -> dict:
```
VascoSch92, line 45: here there is a `dict` as return, but actually you are returning a tuple of a list and a dictionary

fkiraly (approved):
Looks good!

Not blocking, but we should:

* add an explanation of the parameterization to the docstring - the "alpha" distribution is not a common one, so more explanation on how to obtain it would also be useful (e.g., in relation to Gaussian)
* fix the type annotations, as @VascoSch92 suggests

skpro/distributions/alpha.py
```diff
@@ -0,0 +1,67 @@
...
+import pandas as pd
+from scipy.stats import alpha, rv_continuous
+
+from skpro.distributions.adapters.scipy import _ScipyAdapter
+
+
+class Alpha(_ScipyAdapter):
+    r"""Alpha distribution.
+
+    The alpha distribution is characterized by its shape parameter :math:`\a`,
```
fkiraly, line 15: could you add one of the distribution defining functions here, as a reference? pdf for instance

SaiRevanth25: image

Does this check out?

Screenshot 2024-05-24 132836

fkiraly: Almost:

* I think the inline tex expression might have to be on the same line to compile. Have a look how it render in the doc build.
* I would not make the dependency on `a` explicit on the lfet hand side, for consistency
* the `\` in front of the `a` is unnecessary
* `x` is not a random variable, it's a free (bound) variable. You can omit its explanation.
* you can also omit the argument of `Phi` in the explanation, and say "is" instead of "represents".

SaiRevanth25: made the necessary changes, i think its good to go.

PR #358, opened 2024-05-23
[ENH] Creating a new Bayesian Regressor with PyMC as a backend

meraldoantonio:
#### Reference Issues/PRs
https://github.com/sktime/skpro/issues/7

This WIP PR implements a Bayesian Linear Regressor with PyMC as a backend

Yes - it depends on PyMC family: PyMC itself, XArray and ArviZ

The design of the BayesianLinearRegressor. Especially:
1. The introduction of the priors. For now, the class hardcodes the priors. We need to think about the way in which the users should inject their own priors. 

Not yet

#### Any other comments?
N/A

##### For all contributions
- [ X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
(This is not yet done)


skpro/regression/bayesian.py
```diff
@@ -0,0 +1,391 @@
...
+
+            # Expected value of outcome
+            self.mu = pm.Deterministic("mu", self.intercepts + pm.math.dot(X_data, self.slopes))
+
+            # Likelihood (sampling distribution) of observations
+            Y_obs = pm.Normal("y_obs", mu=self.mu, sigma=self.sigma, observed=y_data, dims =("obs_id"))
+
+            # Sample from the posterior
+            self.trace = pm.sample(
+                draws=2000,
```
fkiraly, line 154: should we make these available in the `__init__`? Can of course leave these to defaults.

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,391 @@
...
+                draws=2000,            
+                tune=1500,             
+                chains=1,              
+                random_seed=42,             
+                target_accept=0.90, # Target acceptance probability; higher value leads to higher accuracy but slower sampling
+                return_inferencedata=True,  # Return an InferenceData object 
+                progressbar=True            
+            )
+
+        self.fitted = True
```
fkiraly, line 163: no need to do this, handled by `fit`

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,348 @@
+"""Probabilistic linear regression by PyMC"""
+
+
+__author__ = ["meraldoantonio"]
+
+from skpro.regression.base import BaseProbaRegressor
+import pymc as pm
```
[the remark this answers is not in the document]
meraldoantonio: Noted; I've rewritten the class and incorporated these changes

skpro/regression/bayesian.py
meraldoantonio: Yep; I've added the options to set these in `__init__`

skpro/regression/bayesian.py
meraldoantonio: Noted; removed in the latest draft

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,349 @@
...
+        with pm.Model(coords={"obs_id": X.index, "pred_id": X.columns}) as self.model:
+
+            # Mutable data containers
+            X_data = pm.Data("X", self._X, dims = ("obs_id", "pred_id"))
+            y_data = pm.Data("y", self._y_vals, dims = ("obs_id"))
+
+            # Priors for unknown model parameters
+            self.intercept = pm.Normal("intercept", mu=self.intercept_mu, sigma=self.intercept_sigma)
+            self.slopes = pm.Normal("slopes", mu=self.slopes_mu, sigma=self.slopes_sigma, shape = self._X.shape[1], dims=("pred_id"))
+            self.noise = pm.HalfNormal("noise", sigma=self.noise_sigma)
```
fkiraly, line 116: would inverse gamma not be more standard here, as it is conjugate to the normal?

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,349 @@
...
+        self._predict_done = False
+
+        with pm.Model(coords={"obs_id": X.index, "pred_id": X.columns}) as self.model:
+
+            # Mutable data containers
+            X_data = pm.Data("X", self._X, dims = ("obs_id", "pred_id"))
+            y_data = pm.Data("y", self._y_vals, dims = ("obs_id"))
+
+            # Priors for unknown model parameters
+            self.intercept = pm.Normal("intercept", mu=self.intercept_mu, sigma=self.intercept_sigma)
```
fkiraly, line 114: design discussion point - is there a way to let the user specify these? Either as `pymc` or as skpro distributions. The latter would be interface consistent but would require translation between distribution objects, and `pymc` distributions.

Maybe in another PR though, this would complicate things - `pymc` distributions are nice because of their autodiff, but they seem to require a model context always and a name, so they do not exist on their own, or are not meant to.

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,349 @@
...
+
+            # Constructing the posterior
+            self.trace = pm.sample(chains = self.chains, draws = self.draws)
+
+            # Constructing the in-sample posterior predictive
+            self.trace.extend(pm.sample_posterior_predictive(self.trace))
+            
+        return self
+
+    def get_prior(self, return_type = "xarray"):
```
fkiraly, line 132: why do we need this function?

skpro/regression/bayesian.py
```diff
@@ -0,0 +1,349 @@
...
+        if self._predict_done:
+            pm.set_data({"X": self._X}, coords={"obs_id": self._X.index, "pred_id": self._X.columns})
+        return az.summary(self.trace.prior, var_names = ["intercept", "slopes", "noise"])
+  
+    def plot_ppc(self, **kwargs):
+        """Plot the prior predictive check"""
+        import arviz as az
+        return az.plot_ppc(self.trace, **kwargs)
+    
+    def get_posterior(self, return_type = "xarray"):
```
fkiraly, line 175: we should consider generic bayesian mixins here - a design question for now.

skpro/regression/bayesian.py
fkiraly: we could consider using `bambi`
https://github.com/bambinos/bambi
but that might lock some design decisions in that are not pleasant for sklearn users...

fkiraly (changes requested):
Large number of code formatting failures - see here for a guide on how to automate good code formatting: https://www.sktime.net/en/stable/developer_guide/coding_standards.html

fkiraly:
@meraldoantonio, code formatting failures still, see above. You should be able to automatically format on save, or be able to type things like `black .` in your console to format.

fkiraly (changes requested):
Nice contribution!

Some high-level points:

* could you split the notebook off into a chained PR, based on the estimator PR? The notebook may require some more review (time), and should not block the estimator
* in the estimator, I would kindly ask you to remove the visualization dependencies from `python_dependencies`, and instead introduce dependency checks in the methods that need them? This way, users do not need to visualisation dependencies when using the model in a deployment pipeline.

fkiraly:
Strange import error - is this related to an upper bound of any of the imports implied by 3.9, e.g., `scipy`?

meraldoantonio:
> Strange import error - is this related to an upper bound of any of the imports implied by 3.9, e.g., `scipy`?

Apparently there is a bug with Arviz 0.17 and scipy>=1.13 ([source 1](https://discourse.pymc.io/t/importerror-cannot-import-name-gaussian-from-scipy-signal/14170/6)) ([source 2](https://forum.cadet-web.de/t/scipy-signal-gaussian-import-from-process-0-8-0-to-0-9-0-broken/926/3)). 

The bug is no longer present in Arviz 0.18 but this requires Python 3.10 and above.

As a temporary solution, I've locked the scipy version in `all-extras` in `project.toml`

fkiraly:
Makes sense.

From a maintenance perspective, applying the version bound in the `pyproject.toml` is not a good solution, since the lock is implied only by a single estimator, and not by `scipy` itself.

Could you add the lock instead in the `python_dependencies` tag of the estimator, and revert the changes to `pyproject`?

meraldoantonio:
> Makes sense.
> 
> From a maintenance perspective, applying the version bound in the `pyproject.toml` is not a good solution, since the lock is implied only by a single estimator, and not by `scipy` itself.
> 
> Could you add the lock instead in the `python_dependencies` tag of the estimator, and revert the changes to `pyproject`?

Makes sense! But I've tried this a couple of times and for some reason, the test framework keeps installing the "wrong" version of `scipy` (version 1.13.1), even after specifying "`scipy<=1.12.0`" in the `python_dependencies` tag... 

It might be that other libraries are pulling in a conflicting version, but I haven't managed to find the exact cause..

Any ideas?

fkiraly:
> Any ideas?

Why are you trying to bound `scipy` instead of arviz? I would simply bound `arviz>=0.18`, based on your statements, as well as `python_version >= 3.10`

fkiraly:
PS: why did you close the notebook PR? That was a nice notebook, and indeed it would be nice as separate PR.

meraldoantonio:
> > Any ideas?
> 
> Why are you trying to bound `scipy` instead of `arviz`? I would simply bound `arviz>=0.18`, based on your statements, as well as `python_version >= 3.10`

Ah ok! I was trying to make it work for python 3.9. I've now locked the python version and arviz version in the latest commit

> PS: why did you close the notebook PR? That was a nice notebook, and indeed it would be nice as separate PR.

Just closed it temporarily as I'm still working on adding the `update` method. I’ll re-open it once that's done!

fkiraly:
is this PR ready for merge?

I would really suggest to chunk PRs in smaller, self-contained additions, so we can merge your contributions more quickly and PR do not get too large. For instance, `update` we can add later and it should not "hurt" (or would it?)

meraldoantonio:
> is this PR ready for merge?
Yes it is!

> I would really suggest to chunk PRs in smaller, self-contained additions, so we can merge your contributions more quickly and PR do not get too large. For instance, `update` we can add later and it should not "hurt" (or would it?)
Sure, noted! Yes, you're right, we can add the `update` method in later PRs

fkiraly:
>> is this PR ready for merge?

> Yes it is!

Then you should adjust the tagging/signposting:

* remove the "WIP" tag from the title
* turn the PR form draft to ready (top right side menu)
* "re-request a review", press circling arrows next to dev pictures who have already reviewed

fkiraly:
PS: there is a conflict in the `pyproject.toml`, as your PR does not seem up to date with the recent release. Fixing is as usual: sync your fork, and merge main into your branch.

meraldoantonio:
PS on the `update` method: Based on my research, there isn't an ideal method for performing update using `PyMC`. There are 2 possible routes, however, both have significant limitations:

1. **Rebuilding the prior from the posterior using marginal 1D empirical approximation**: This approach, described in the [PyMC documentation](https://www.pymc.io/projects/examples/en/latest/howto/updating_priors.html), works fine when dealing with a single variable. However, if we have multiple variables (such as in our case, where we have an intercept and a slope per feature), this method is problematic because it loses the correlation structure between variables. 

2. **Rebuilding the prior using a multivariate normal approximation**: Another approach is to approximate the posterior as a multivariate normal distribution, as discussed in the [PyMC Experimental library](https://www.pymc.io/projects/experimental/en/latest/generated/pymc_experimental.utils.prior.prior_from_idata.html). While this method retains the correlation between variables, it assumes normality, which may not hold in cases where the posterior  is multimodal or skewed. 

Instead of introducing a potentially misleading `update` method, I think it would be better to omit it at this stage

fkiraly (changes requested):
The code looks great!

There are some testing issues:

* the failure on 3.9 still seems there, even though we clearly set the tag. Could you check if you can understand what is going on there?
* `get_test_params` should be populated with at least two examples.

meraldoantonio:
> The code looks great!
> 
> There are some testing issues:
> 
> * the failure on 3.9 still seems there, even though we clearly set the tag. Could you check if you can understand what is going on there?
> * `get_test_params` should be populated with at least two examples.

I'm not sure what's going on with the failing tests with Python 3.9, I've tried playing around with the versions but somehow the test suite ignores this tag :(

I've added examples to `get_test_param`

fkiraly:
Summarizing our collaborative coding session on November 8, we have found out what the problem is in the failures on python 3.9, namely a known issue with the dependency checking framework that is already fixed in `sktime` and `skbase`.

The todo is summarized in https://github.com/sktime/skpro/issues/489, and should happen in a separate pull request.

fkiraly:
oops, looks like this got accidentally closed, I reopened.

The reason for this is, you accidentally used a reserved "closing keyword" in #490, that is, you wrote "fixes #358" in the preamble. If you do this, then the mentioned issue or PR gets closed once the PR is merged! I did not know btw that this also applied to PR, not just to issues.

fkiraly (approved):
Looks good now. I fixed a merge conflict in `pyproject`, somehow the dead deps `tabulate` and `uncertainties` returned.

PR #363, opened 2024-05-25
[ENH] Half Normal Distribution

SaiRevanth25:
 Implements Half Normal Distribution

 No

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/halfnormal.py
+++ b/skpro/distributions/halfnormal.py
@@ -15,7 +15,6 @@ class HalfNormal(_ScipyAdapter):
     This distribution is univariate, without correlation between dimensions
     for the array-valued case.
 
-    The pdf is defined only for :math:`\( x \geq 0 \)`. For :math:`\( x < 0 \), \( f(x) = 0 \)`.
     The distribution is `cut off` at :math:`\( x = 0 \)`. There is no mass assigned to
     negative values; they are entirely excluded from the distribution.
 
@@ -23,7 +22,7 @@ class HalfNormal(_ScipyAdapter):
     :math:`\sigma`, such that the pdf is
 
     .. math:: f(x) = \frac{\sqrt{2}}{\sigma \sqrt{\pi}}
-                    \exp\left(-\frac{x^2}{2\sigma^2}\right)
+                    \exp\left(-\frac{x^2}{2\sigma^2}\right), x>0 otherwise 0
 
     The standard deviation :math:`\sigma` is represented by the parameter ``sigma``.
 
--- a/skpro/distributions/halfnormal.py
+++ b/skpro/distributions/halfnormal.py
@@ -1,0 +1,210 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Half-Normal probability distribution."""
+
+__author__ = ["SaiRevanth25"]
+
+import numpy as np
+import pandas as pd
+from scipy.special import erf, erfinv
+
+from skpro.distributions.base import BaseDistribution
+
+
+class HalfNormal(BaseDistribution):
+    r"""Half-Normal distribution (skpro native).
+
+    This distribution is univariate, without correlation between dimensions
+    for the array-valued case.
+
+    The half-normal distribution is parametrized by the standard deviation
+    :math:`\sigma`, such that the pdf is
+
+    .. math:: f(x) = \frac{\sqrt{2}}{\sigma \sqrt{\pi}}
+                    \exp\left(-\frac{x^2}{2\sigma^2}\right)
+
+    The standard deviation :math:`\sigma` is represented by the parameter ``sigma``.
+
+    Parameters
+    ----------
+    sigma : float or array of float (1D or 2D), must be positive
+        standard deviation of the half-normal distribution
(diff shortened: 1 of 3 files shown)


skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,210 @@
...
+from skpro.distributions.base import BaseDistribution
+
+
+class HalfNormal(BaseDistribution):
+    r"""Half-Normal distribution (skpro native).
+
+    This distribution is univariate, without correlation between dimensions
+    for the array-valued case.
+
+    The half-normal distribution is parametrized by the standard deviation
```
fkiraly, line 19: I think you need to say something about the domain, or that pdf is zero for negative.

Also, please clarify whether this is truncated or clipped, as it might not be clear to the reader: that is, is there a mass on zero?

skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,210 @@
...
+        Private method, to be implemented by subclasses.
+
+        Returns
+        -------
+        2D np.ndarray, same shape as ``self``
+            energy values w.r.t. the given points
+        """
+        sigma = self._bc_params["sigma"]
+
+        energy_arr = sigma * np.sqrt(2 - 2 / np.pi)
```
fkiraly, line 68: can you explain how you have derived these?

skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,210 @@
...
+        -------
+        2D np.ndarray, same shape as ``self``
+            energy values w.r.t. the given points
+        """
+        sigma = self._bc_params["sigma"]
+
+        cdf = self.cdf(x)
+        pdf = self.pdf(x)
+
+        energy_arr = (x) * (2 * cdf - 1) + 2 * sigma**2 * pdf
```
fkiraly, line 96: same, can you kindly explain your derivation?

skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,210 @@
...
+            values to evaluate the log pdf at
+
+        Returns
+        -------
+        2D np.ndarray, same shape as ``self``
+            log pdf values at the given points
+        """
+        sigma = self._bc_params["sigma"]
+        log_pdf_arr = -0.5 * (x / sigma) ** 2
+        log_pdf_arr = log_pdf_arr - np.log(np.sqrt(2 / np.pi) / sigma)
```
fkiraly, line 159: is the "-" correct?

skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,210 @@
...
+__author__ = ["SaiRevanth25"]
+
+import numpy as np
+import pandas as pd
+from scipy.special import erf, erfinv
+
+from skpro.distributions.base import BaseDistribution
+
+
+class HalfNormal(BaseDistribution):
```
fkiraly, line 13: you could use the scipy adapter, scipy has `halfnorm` - although that's not necessary, expressions are straightforward. Just point this out, not a change request

skpro/distributions/halfnormal.py
SaiRevanth25: Updated!

skpro/distributions/halfnormal.py
SaiRevanth25: i think using the scipy adatpter would be better in this case!!

SaiRevanth25: i think using the scipy adapter would be better in this case!!

fkiraly (changes requested):
The scipy adapter does not add explicit energy calculations - could you add those back? It is useful to have these as explicit numerical formulae if you are able to work them out!

skpro/distributions/halfnormal.py
```diff
@@ -0,0 +1,76 @@
...
+
+class HalfNormal(_ScipyAdapter):
+    r"""Half-Normal distribution.
+
+    This distribution is univariate, without correlation between dimensions
+    for the array-valued case.
+
+    The pdf is defined only for \( x \geq 0 \). For \( x < 0 \), \( f(x) = 0 \).
+    The distribution is "cut off" at \( x = 0 \). There is no mass assigned to
+    negative values; they are entirely excluded from the distribution.
```
VascoSch92, line 20:
```suggestion
    The pdf is defined only for :math:`\( x \geq 0 \)`. For :math:`\( x < 0 \), \( f(x) = 0 \)`.
    The distribution is `cut off` at :math:`\( x = 0 \)`. There is no mass assigned to
    negative values; they are entirely excluded from the distribution.
```

otherwise you have
Screenshot 2024-05-26 at 22 01 53

VascoSch92 (changes requested):

skpro/distributions/halfnormal.py
fkiraly: I think the notation is unusual, I would also shorten this a lot. I would just append $, x>0; \mbox{otherwise} 0$ to the formula.

SaiRevanth25: Updated it!! please let me know if anything else is to be done! Thank you.

fkiraly (approved):

Issue #342, opened 2024-05-18
[ENH] improved `polars` support - lazy and native logic


fkiraly:
Issue about next steps in extending `polars` support in `skpro`.

Background: `skpro` already supports both lazy and eager `polars` containers, as `polars_eager_table` and `polars_lazy_table`.

However, key limitations:

* support is currently exclusively via conversion back/forth
* the round trip to `numpy` or `pandas` breaks the lazy chain

So, suggested next steps, using only the eager type:

1. try `polars` input/output with a few estimators, `fit` / `predict` only. Add tests in a dedicated `polars` test file, non-systematic for the start.
2. `sklearn` now supports `polars` - so, we should try to pass on `polars` frames in some of the estimators, extending the `X_inner_mtype` to both pandas and polars (in a list of str, with `polars_eager_table`). Example: `GaussianProcess`
3. if that works well, we should extend other regresssors wrapping `sklearn` in the same way.
4. next I would work on `Pipeline`, a composite. The `Pipeline` is native to `skpro`.

Once the above seems to work, I would look at the lazy` type.
Here, we should use the boilerplate layer to make lazy state changes.
Design will follow when we are there, ideas appreciated.

FYI @julian-fong.

julian-fong:
Linking #6381 as discussion and coordination thread

PR #370, opened 2024-05-30
[ENH] Introducing test file for polars support for estimators

julian-fong:
#### Reference Issues/PRs

#342 is the linked issue that overlays the entire workflow for polars support. Opening this thread mainly for the test file I am planning to implement for polars support in skpro

Implements a test file for testing various fit/predict functions using polars dataframes. Ideas for any tests regarding testing polars support in skpro is much appreciated!

No, it does not introduce a new dependency

I was going through a couple regression estimators glm , and quickly realized that the underlying dependency `statsmodels` currently do not support polars dataframes as inputs for their fit functions. Since that is the case, it seems like conversion between polars to pandas must be necessary in order to allow fitting (and I would assume predicting) to happen.

@fkiraly Would this be the correct approach to deal with estimators that do not potentially have underlying polars support in them?


julian-fong:
a follow up question: what is the expectation if currently some of the estimators are coded to return pandas dataframes. As an example, in `regression.mapie`, the `_predict` function turns `y_pred` into a pandas Dataframe and returns that as an output. 

Should we approach writing this test file under the assumption that all estimators have the capability/functionality to return polars Dataframes?

skpro/tests/test_polars.py
```diff
@@ -0,0 +1,37 @@
+"""Test file for polars dataframes"""
+
+import pytest
+from sktime.utils.validation._dependencies import _check_soft_dependencies
```
fkiraly, line 4: use the skpro version

fkiraly:
> what is the expectation if currently some of the estimators are coded to return pandas dataframes. As an example, in `regression.mapie`, the `_predict` function turns `y_pred` into a pandas Dataframe and returns that as an output. 

afaik, the boilerplate layer, i.e., `BaseProbaRegerssor.predict`, will turn it back into a `polars.DataFrame` if that was passed initially to `fit`. In the tests, I would check that this is indeed the case.

skpro/tests/test_polars.py
```diff
@@ -0,0 +1,121 @@
...
+import pytest
+from sklearn.datasets import load_diabetes
+from sklearn.model_selection import train_test_split
+
+from skpro.datatypes import check_raise
+from skpro.datatypes._table._check import check_polars_table
+from skpro.datatypes._table._convert import convert_pandas_to_polars_eager
+from skpro.utils.validation._dependencies import _check_soft_dependencies
+
+X, y = load_diabetes(return_X_y=True, as_frame=True)
```
fkiraly, line 14: data generation and loading should happen in a method or - optimally - pytest fixture, not on module load.
(we have too many of these in `sktime` and it slows test collection and imports)

skpro/tests/test_polars.py
```diff
@@ -0,0 +1,121 @@
...
+X_train, X_test, y_train, _ = train_test_split(X, y, test_size=0.33, random_state=42)
+
+TEST_ALPHAS = [0.05, 0.1, 0.25]
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["polars", "pyarrow"], severity="none"),
+    reason="skip test if polars/pyarrow is not installed in environment",
+)
+def test_polars_eager_conversion_methods():
```
fkiraly, line 30: possibly redundant - these should already be tested by various tests in the `datatypes` module, have you looked whether this is the case?

skpro/tests/test_polars.py
```diff
@@ -0,0 +1,121 @@
...
+    Parameters
+    ----------
+
+    estimator: a given regression estimator
+
+    """
+    # create a copy of estimator to run further checks
+    estimator_copy = estimator
+
+    X_train_pl = convert_pandas_to_polars_eager(X_train)
```
fkiraly, line 66: this should be part of the fixture generation, it should produce polars X and y

skpro/tests/test_polars.py
julian-fong: Yes I think they are covered inside `datatypes.tests.test_convert.py`?

skpro/tests/test_polars.py
julian-fong: configured pytext fixture `polars_load_diabetes_polars` to load polars X and y

skpro/tests/test_polars.py
julian-fong: configured pytest fixture `polars_load_diabetes_pandas` to load polars X and y

julian-fong:
The CLI failures look weird - some of them are just long empty pages.

julian-fong:
If i am trying to test an estimator say in `test_polars_eager_regressor_in_fit_predict`, what is the best approach to do so?

skpro/tests/test_polars.py
```diff
@@ -0,0 +1,135 @@
...
+    assert check_raise(
+        y_pred_interval, "pred_interval", "Proba", "predict_interval return"
+    )
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies(["polars", "pyarrow"], severity="none"),
+    reason="skip test if polars/pyarrow is not installed in environment",
+)
+def test_polars_eager_regressor_in_predict_quantiles(estimator):
```
fkiraly, line 123: this is failing because there is no `estimator` fixture - you need to add a `mark.parametrize` or a `pytest_generate_fixtures` implementation

julian-fong: Should I specify one at random? Or is there a way to potentially pass all regression models into this test - not 100% sure what is the best approach here when specify estimators

fkiraly: interesting question - for the start, I would say, pick any, as none support polars internally, later we can be more systematic

fkiraly (approved):

PR #381, opened 2024-06-11
[BUG] fix `test_methods_p` logic when `shuffle` is `True`

### LINKED ISSUE #380: [BUG] test_methods_p not handling getattr after shifting
**Expected behavior**

when the `shuffled` is true while the testing happens in the `test_all_distrs.py` and it tests the 
`test_methods_p` it will shuffle the distribution and store the shuffled distribution's object instance in `d`.

The output from `getattr(object_instance,method)(p)` will give the output before the distribution is shuffled but when it is checking the output format if indices and columns are matching it is calling `_check_output_format(res, d, method)`.

**Solution**
`res=getattr(object_instance,method)(p)` to be replaced with `res = getattr(d, method)(p)` as this will account for the object_instance after shifting the distribution. This will ensure the outputs are checked with the shifted distribution's instance.

ShreeshaM07:
#### Reference Issues/PRs

fixes #380 

`object_instance` is the distribution before shuffling.
`d` is the one after shuffling so it needs to be used instead 
`res=getattr(object_instance,method)(p)` to be replaced with `res = getattr(d, method)(p)` as this will account for the object_instance after shifting the distribution. This will ensure the outputs are checked with the shifted distribution's instance.

### DIFF
--- a/skpro/distributions/tests/test_all_distrs.py
+++ b/skpro/distributions/tests/test_all_distrs.py
@@ -167,7 +167,7 @@ class TestAllDistributions(PackageConfig, DistributionFixtureGenerator, QuickTes
         else:
             p = np_unif
 
-        res = getattr(object_instance, method)(p)
+        res = getattr(d, method)(p)
 
         _check_output_format(res, d, method)
 


ShreeshaM07:
@fkiraly All the checks have passed if this is merged then it would be helpful for the Histogram Distribution #382 to be merged as well.

fkiraly (approved):
Indeed, this was a mistake, thanks!

PR #382, opened 2024-06-11
[ENH] Histogram distribution

### LINKED ISSUE #323: [ENH] histogram distribution

ShreeshaM07:
#### Reference Issues/PRs

fixes #323 
PR is a new one with updated main merged with #335  as there were some merge conflicts in `__init__.py` in distributions and also `distributions.rst` in `api_reference`.

Implements the histogram distribution using bins and bin_mass as the parameters.

No

Yes

#### Any comments
All the discussion regarding the Histogram discussion has been discussed in the PR #335.

##### For all contributions

##### For new estimators


fkiraly:
merged #381 into `main`, and then from `main` into this.

fkiraly (changes requested):
Great! Really nice work, I like the design thinking in the new base class.

Before we merge, the current "template" pattern does not allow override of the public methods `pdf` etc, or we diverge from the template pattern, see https://github.com/kamranahmedse/design-patterns-for-humans for the pattern.

Could you think of the best way to avoid the override? Or, let me know if you think this is not possible with the current design.

We do not need to merge the two base classes together btw, just avoid the override.

ShreeshaM07:
I hope this will suffice as I am not overriding any public methods anymore.

ShreeshaM07:
@fkiraly , All the checks have passed after the overriding of public methods were avoided as well. Are we waiting on something for this merge?

fkiraly:
a factor of 5 or 6 is considered substantial, so I would advise we check how much time the conversion of arrays requires. 

I would then proceed based on the outcoe:
* if overall runtimes are comparable, I would advise we merge whatever functions we have currently vectorized, and open an issue. This is because that means, a different parameter representation might eventually make the computations more performant.
* if overall runtime with conversion is *worse*, I would advise we pin the matter in the form of an issue and merge the current version.
* If it still is a substantial speed-up, I would advise vectorizing a few key methods in addition.

ShreeshaM07:
This is including the vectorization time. With 2 cases of vectorization time inclusion case 1 when it is not vectorized in `__init__` but in separate functions and case 4 when it is vectorized in `__init__` and not in separate functions.

- mean
    - vectorization time included in each function
     vec_mean_with_vec_time
    - vectorization time not included but vectorized
      vectorized_mean
    - No vectorization at all
      curr_mean
    - vectorized in init itself not in separate methods
      vec_init_mean


- variance
    - vectorization time included in each function
      vec_var_with_vec_time
    - vectorization time not included but vectorized
      vectorized_var
    - no vectorization at all
      curr_var
   - vectorized in init itself not in separate methods
     vec_init_var

It is bettered when it is vectorized in `__init__` itself but becomes worse when vectorized in each methods as and when required. It might fall under category 3 from what you suggested. I'll try to vectorize a few more methods but it might lead to some changes so ill make it a separate PR so that in case it doesn't workout we can always go with the current PR.

ShreeshaM07:
This is including the vectorization time. With 2 cases of vectorization time inclusion case 1 when it is not vectorized in __init__ but in separate functions and case 4 when it is vectorized in __init__ and not in separate functions.

- mean
    - vectorization time included in each function
     vec_mean_with_vec_time
    - vectorization time not included but vectorized
      vectorized_mean
    - No vectorization at all
      curr_mean
    - vectorized in init itself not in separate methods(i.e sum of time to vectorize + sum of time to run the mean operations after vectorizing)
      vec_init_mean



- variance
    - vectorization time included in each function
      vec_var_with_vec_time
    - vectorization time not included but vectorized
      vectorized_var
    - no vectorization at all
      curr_var
   - vectorized in init itself not in separate methods(i.e sum of time to vectorize + sum of time to run the mean operations after vectorizing)
     vec_init_var

As you can see it actually gets worse when it is vectorized in `__init__`. The time for vectorizing these ragged inputs itself is about `0.0002` secs after which it takes time to calculate mean, var etc. So even though the time to calculate variance etc gets optimised by a factor of `5` or `6`. The initializing itself is taking a lot of time.
This should come under the category 2
> if overall runtime with conversion is worse, I would advise we pin the matter in the form of an issue and merge the current version.

Please do give your take based on your inference of the data.

ShreeshaM07:
> Hm, I would then agree that the payoff is perhaps not worth the effort at the moment.
> 
> I would suggest to summarize the findings and outstanding questions in an issue, and perhaps at a later time point review together:
> 
> * the design of the intermediate base class from this PR
> * the designs and parametrizations of the histogram distribution and two similar distributions: `Empirical`, and `Mixture`

Yeah I will  summarize the findings of this PR in an issue.

fkiraly (approved):

PR #384, opened 2024-06-12
[ENH] `GLM` with multiple `distributions` and `link` function support

### LINKED ISSUE #230: [ENH] further work on `GLMRegressor` interfacing `statsmodels` `GLM`
From #222, remaining work items to interface `statsmodels` `GLM`:

* currently, only gaussian `family` is implemented. Further families should be interfaced from `statsmodels`, in particular Gamma and Tweedie. https://github.com/sktime/skpro/issues/383
* we should try to cover as many parameters as we can in `get_test_params`, currently coverage is low.
* the docstring should say, for every parameter, what possible values are. E.g., what are possible values for `cov_type`, `method`, what are expected sizes for `start_params`, etc.
* some of the parameters of `statsmodels` `GLM` are not exposed to the user, as they require array-like input which is unavailable in `predict`, e.g., `offset`, `exposure`. It should be investigated how these could be interfaced, with a sensible treatment in `predict` where applicable.

### LINKED ISSUE #383: [ENH] Multiple link function support for GLMs

The existing `GLMRegressor` supports only the Gaussian family and in that specifically only the `Identity` link function. There are many other family like Poisson, Binomial, Negative Binomial , Gamma that can be implemented and each of these having a set of safe `link` functions of their own. So the idea would be to support these different family and their `link` function in order to make the `GLMRegressor` of more broader use.

Currently the `_glm.py` in skpro is only using the `Gaussian` family and its default `link` function interfaced from statsmodels. However we can extend this to support multiple link functions and different families by either adding an adapter for this to convert the available families in `statsmodels` to `skpro` BaseDistributions available.

ShreeshaM07:
#### Reference Issues/PRs

fixes #383 and closes #230

This creates an adapter converting the statsmodels GLM families to skpro equivalent giving `GLM`s a broader capability of distributions and link functions.

No

Yes

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/regression/linear/_glm.py
+++ b/skpro/regression/linear/_glm.py
@@ -21,29 +21,40 @@ class GLMRegressor(BaseProbaRegressor):
 
     Parameters
     ----------
-    family : str, default : "Normal"
-        Available options are
-        1.Normal
-        2.Poisson
-        3.Gamma
-    link : str
-        Available safe options are
-        Normal : Log, Identity, InversePower
-        Poisson : Log, Identity, Sqrt
-        Gamma : Log, Identity, InversePower
-    offset_var : pd.Index([str]) or int, default = None
-        If ``pd.Index([str])``, then the exog or ``X`` passed while ``fit``-ting
+    family : string, default : "Normal"
+        The family parameter denotes the type of distribution
+        that will be used.
+        Available family/distributions are
+        1."Normal"
+        2."Poisson"
+        3."Gamma"
+    link : string, default : None
+        This parameter is used to represent the link function to be
+        used with the distribution.
+        If default is None it will internally replace with default of the
+        respective family. The default is the first string
+        against each family below.
+        Available safe options for the respective family are:
+        ``Normal`` : "Identity", "Log", "InversePower";
+        ``Poisson`` : "Log", "Identity", "Sqrt";
+        ``Gamma`` : "InversePower", "Log", "Identity";
+    offset_var : string or int, default = None
+        Pass the column name as a string or column number as an int in X.
+        If string, then the exog or ``X`` passed while ``fit``-ting
         must have an additional column with column name passed through
         ``offset_var`` with any values against each row. When ``predict``ing
         have an additional column with name same as string passed through ``offset_var``
-        in X with all the ``offset_var`` values stored in the column for each row.
+        in X with all the ``offset_var`` values for predicting
+        stored in the column for each row.
         If ``int`` it corresponding column number will be considered.
-    exposure_var : pd.Index([str]) or int, default = None
-        If ```pd.Index([str])``, then the exog or ``X`` passed while ``fit``-ting
+    exposure_var : string or int, default = None
+        Pass the column name as a string or column number as an int in X.
+        If string, then the exog or ``X`` passed while ``fit``-ting
         must have an additional column with column name passed through
         ``exposure_var`` with any values against each row. When ``predict``ing
         have additional column with name same as string passed through ``exposure_var``
-        in X with all the ``exposure_var`` values stored in the column for each row.
+        in X with all the ``exposure_var`` values for predicting
+        stored in the column for each row.
         If ``int`` it corresponding column number will be considered.
     missing : str
         Available options are 'none', 'drop' and 'raise'. If 'none', no nan
@@ -403,16 +414,16 @@ class GLMRegressor(BaseProbaRegressor):
 
         from warnings import warn
 
-        l1 = "Note: in `GLMRegressor`, the sequence of the parameters will change "
-        l2 = "in skpro version 2.4.0. It will be as per the order present in the"
-        l3 = "current docstring with the top one being the first parameter.\n"
-        l4 = "The defaults for the parameters will remain same and "
-        l5 = "there will be no changes.\n"
-        l6 = "Please use the `kwargs` calls instead of positional calls for the"
-        l7 = "parameters until the release of skpro 2.4.0 "
-        l8 = "as this will avoid any discrepancies."
-        warn_msg = l1 + l2 + l3 + l4 + l5 + l6 + l7 + l8
-        warn(warn_msg)
+        warn(
+            "Note: in `GLMRegressor`, the sequence of the parameters will change "
+            "in skpro version 2.4.0. It will be as per the order present in the"
+            "current docstring with the top one being the first parameter.\n"
+            "The defaults for the parameters will remain same and "
+            "there will be no changes.\n"
+            "Please use the `kwargs` calls instead of positional calls for the"
+            "parameters until the release of skpro 2.4.0 "
+            "as this will avoid any discrepancies."
+        )
 
     def _fit(self, X, y):
         """Fit regressor to training data.
@@ -445,20 +456,7 @@ class GLMRegressor(BaseProbaRegressor):
         offset_var = self._offset_var
         exposure_var = self._exposure_var
 
-        if offset_var is not None:
-            if isinstance(offset_var, int):
-                offset_var = pd.Index([X.iloc[:, offset_var].name])
-        if exposure_var is not None:
-            if isinstance(exposure_var, int):
-                exposure_var = pd.Index([X.iloc[:, exposure_var].name])
-        if offset_var is not None and exposure_var is not None:
-            X = X.drop([offset_var[0], exposure_var[0]], axis=1)
-        elif offset_var is not None:
-            X = X.drop(offset_var, axis=1)
-        elif exposure_var is not None:
-            X = X.drop(exposure_var, axis=1)
-
-        X_ = self._prep_x(X)
+        X_ = self._prep_x(X, offset_var, exposure_var, False)
 
         y_col = y.columns
 
@@ -475,19 +473,21 @@ class GLMRegressor(BaseProbaRegressor):
 
         self._estimator = glm_estimator
 
-        fitted_glm_model = glm_estimator.fit(
-            self._start_params,
-            self._maxiter,
-            self._method,
-            self._tol,
-            self._scale,
-            self._cov_type,
-            self._cov_kwds,
-            self._use_t,
-            self._full_output,
-            self._disp,
-            self._max_start_irls,
-        )
+        glm_fit_params = {
+            "start_params": self._start_params,
+            "maxiter": self._maxiter,
+            "method": self._method,
+            "tol": self._tol,
+            "scale": self._scale,
+            "cov_type": self._cov_type,
+            "cov_kwds": self._cov_kwds,
+            "use_t": self._use_t,
+            "full_output": self._full_output,
+            "disp": self._disp,
+            "max_start_irls": self._max_start_irls,
+        }
+
+        fitted_glm_model = glm_estimator.fit(**glm_fit_params)
 
         PARAMS_TO_FORWARD = {
             "df_model_": glm_estimator.df_model,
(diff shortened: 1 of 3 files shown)


ShreeshaM07:
I have thought of 2 possible ways to pass the `offset` and `exposure` parameters while `predict`ing.
<a href=https://github.com/sktime/skpro/pull/384/commits/4bdd71d9634ee926149aca03e11d27be072750c2>Design 1</a> and <a href=https://github.com/sktime/skpro/pull/384/commits/e6b20d182c17dc4e84f2cdf85c2a3580e3f95b28>Design2 </a>.

## Design 1
Idea behind this is that I initialize the object of the class with whether `offset` and `exposure` is going to be passed while predicting and will contain boolean values. Since we cannot have an argument `offset` and `exposure` in the `_predict` function user must pass additional columns with names `offset` and `exposure` which will be removed/dropped while `fit`ting and `predict_proba` but will be converted to `array` and passed to `statsmodels` `predict` thereby keeping consistency in the number of `columns` and `rows`.
`Pros`
- It will allow user to keep changing the `offset` and `exposure` params even after fitting data.
- User doesn't have to keep re initializing the model in case they want to change the `offset`/`exposure` values.

`Cons`
- User will have to create 2 extra `columns` with the exact same name of `offset` and `exposure` each time in the `exog`/`X` variable.
- No way to add a test setting to check this as it cannot add extra columns with the name `offset`/`exposure`.

## Design 2
Idea here is that I pass the `offset`/`exposure` array while initializing the object itself. It must have the same length as the number of rows in the `X` that will be passed while predicting.
`Pros`
- Simple code :)
- User doesn't have to struggle having to modify the `X` while `fit`ting.

`Cons`
- shape of `X` passed to `predict` must be known while initializing/constructor the object itself.
- If the size of prediction `X` needs to be changed the `offset`/`exposure` must be changed for which it will need to be `re-initialized` again.
- `Re-fitting` the model will need to happen each time `X` size changes for `predict`.
- Even for this method there is no way to add a test setting as size of `X` used for predicting is not known prior.

## My Conclusion
Both the designs above are not a foolproof way but they both give correct answers. Since we cannot add a test setting for both these I am not sure I can think of a better way to do it where it can be added to the test setting too.

fkiraly:
A strong design principle in `sklearn`-like designs is separating data from model specification, because only that allows reliable model composition.

As Option 2 violates that principle (offset and exposure are part of the data), as your cons imply, I would have a very strong preference towards option 1.

I would vary the idea a little, by adding parameters `exposure_var` and `offset_var`, which are `pandas` index elements or int. If int, it is assumed `iloc`; otherwise `loc`. There are sensible defaults, I would say `None`, which means no exposure/offset is passed (and I believe `statsmodels` assumes constant exposure and zero offset then).

ShreeshaM07:
> I would vary the idea a little, by adding parameters `exposure_var` and `offset_var`, which are `pandas` index elements or int. If int, it is assumed `iloc`; otherwise `loc`. There are sensible defaults, I would say `None`, which means no exposure/offset is passed (and I believe `statsmodels` assumes constant exposure and zero offset then).

I am little unsure on what this means. Do you mean to add these 2 parameters `exposure_var` and `offset_var` along with the already present `exposure` and `offset` bool parameters as I have done in <a href=https://github.com/sktime/skpro/pull/384/commits/4bdd71d9634ee926149aca03e11d27be072750c2>Design1</a>.
If we are adding along with these then I think there would be no need to have the extra columns in `X` that the user has to pass while `fitting`,`predicting` etc. This would again mean that we are not separating the model specification from the data.

Could you please elaborate the idea a little.

fkiraly:
> I am little unsure on what this means. Do you mean to add these 2 parameters `exposure_var` and `offset_var` along with the already present `exposure` and `offset` bool parameters as I have done in

I was suggesting to replace these with two more informative variables, concretely:

Replace `exposure` by `exposure_var`, type is `pandas` index element (e.g., string). If `exposure_var = None`, it behaves like your `exposure = False`. If `exposure_var = "exposure"`, it behaves like your `exposure = True`. And `exposure_var = "sth_else"` can be used for pointing to another variable.

> Could you please elaborate the idea a little as to what exactly the `exposure_var` and `offset_var` will contain.

The type would be a single `pandas` index element. For instance, `str`. If `str`, then it would point ot the column `loc`. I also suggest to use integers `iloc` rather than `loc`.

skpro/regression/linear/_glm.py
```diff
@@ -167,8 +179,44 @@ class GLMRegressor(BaseProbaRegressor):
...
+                return sm_fmly[family](link_function)
+            except Exception:
+                msg = "Invalid link for family, default link will be used"
+                warn(msg)
+
+        return sm_fmly[family]()
+
     def __init__(
         self,
+        family="Normal",
```
fkiraly, line 218: it would have been great to have these as starting params from the get-go, but right now we cannot add these as starting parameters due to the deprecation policy - it would make user code break.

We have to add the new params at the end, and I would suggest following deprecation policy to move them to the start eventually, see https://www.sktime.net/en/latest/developer_guide/deprecation.html

ShreeshaM07:
> Replace `exposure` by `exposure_var`, type is `pandas` index element (e.g., string). If `exposure_var = None`, it behaves like your `exposure = False`. If `exposure_var = "exposure"`, it behaves like your `exposure = True`. And `exposure_var = "sth_else"` can be used for pointing to another variable.

Yea this makes more sense and I have completed implementing it that way and also re ordered the new params to the end. Next we will have to work on adding test setting for it.

fkiraly:
great! btw, if you want to move the position of the parameters later, we should follow the "move parameter position" recipe - we can make the change right away.

skpro/regression/linear/_glm.py
```diff
@@ -18,11 +21,34 @@ class GLMRegressor(BaseProbaRegressor):
 
     Parameters
     ----------
+    family : str, default : "Normal"
+        Available options are
+        1.Normal
```
fkiraly, line 26: should be strings, and explain what these mean

skpro/regression/linear/_glm.py
```diff
@@ -18,11 +21,34 @@ class GLMRegressor(BaseProbaRegressor):
...
+        2.Poisson
+        3.Gamma
+    link : str
```
fkiraly, line 29: add and explain the default

skpro/regression/linear/_glm.py
```diff
@@ -18,11 +21,34 @@ class GLMRegressor(BaseProbaRegressor):
...
+        Available safe options are
+        Normal : Log, Identity, InversePower
+        Poisson : Log, Identity, Sqrt
+        Gamma : Log, Identity, InversePower
+    offset_var : pd.Index([str]) or int, default = None
```
fkiraly, line 34: why is this a pd.index? I was thinking index element, there is no benefit in passing this wrapped in an index

skpro/regression/linear/_glm.py
```diff
@@ -201,6 +332,88 @@ def __init__(
         self.max_start_irls = max_start_irls
         self.add_constant = add_constant
 
+        if family == "0":
```
fkiraly, line 335: this looks excessive - have you looked at the deprecation handling? It has examples at the end.

I think we should avoid setting new defaults for unaffected parameters.

fkiraly (changes requested):
Nice! Great extension.

Small blocking change requests relate to:

* the docstring should be clearer on how the new parameters are handled
* the change pattern is a bit excessive, so there is risk of error. Have you checked that these are indeed the right defaults?
    * there is a lot of repetition - if this were "final" code I would say this should be made more DRY

skpro/regression/linear/_glm.py
```diff
@@ -201,6 +332,88 @@ def __init__(
...
+        else:
+            self._max_start_irls = max_start_irls
+        if add_constant == "16":
+            self._add_constant = False
+        else:
+            self._add_constant = add_constant
+
+        from warnings import warn
+
+        l1 = "Note: in `GLMRegressor`, the sequence of the parameters will change "
```
fkiraly, line 406: not good style - do
```python
warn_msg = (
    "line 1"
    "line 2"
    "line 3"
)

File: skpro/regression/linear/_glm.py [hunk 7]
```diff
@@ -227,31 +440,53 @@ def _fit(self, X, y):
...
+
+        if offset_var is not None:
+            if isinstance(offset_var, int):
+                offset_var = pd.Index([X.iloc[:, offset_var].name])
+        if exposure_var is not None:
+            if isinstance(exposure_var, int):
+                exposure_var = pd.Index([X.iloc[:, exposure_var].name])
+        if offset_var is not None and exposure_var is not None:
+            X = X.drop([offset_var[0], exposure_var[0]], axis=1)
+        elif offset_var is not None:
```
fkiraly, line 456: this can be simplified, just drop them sequentially

skpro/regression/linear/_glm.py
```diff
@@ -227,31 +440,53 @@ def _fit(self, X, y):
...
-            self.method,
-            self.tol,
-            self.scale,
-            self.cov_type,
-            self.cov_kwds,
-            self.use_t,
-            self.full_output,
-            self.disp,
-            self.max_start_irls,
+            self._start_params,
```
fkiraly, line 479: I would say, prepare a dictionary for this and use kwargs, not args. Using so many args is quite risky.

skpro/regression/linear/_glm.py
```diff
@@ -304,15 +539,82 @@ def _predict(self, X):
         -------
         y : pandas DataFrame, same length as `X`, with same columns as y in fit
         """
+        offset_var = self._offset_var
+        exposure_var = self._exposure_var
+        offset_arr = None
+        exposure_arr = None
+
+        if offset_var is not None:
```
fkiraly, line 547: duplication, not DRY. You could stick it into `_prep_x`.

skpro/regression/linear/_glm.py
```diff
@@ -304,15 +539,82 @@ def _predict(self, X):
...
+            y_sigma = y_predictions_df["mean_se"].rename("sigma").to_frame()
+            params["mu"] = y_mu
+            params["sigma"] = y_sigma
+        elif skp_dist == Poisson:
+            y_mu = y_predictions_df["mean"].rename("mu").to_frame()
+            params["mu"] = y_mu
+        elif skp_dist == Gamma:
+            y_mean = y_predictions_df["mean"]
+            y_sd = y_predictions_df["mean_se"]
+            y_alpha = (y_mean / y_sd) ** 2
```
fkiraly, line 606: kudos, this is the right idea and not obvious.

Though, it should be `y_mean / (y_sd ** 2)`, no?

skpro/regression/linear/_glm.py
```diff
@@ -304,15 +539,82 @@ def _predict(self, X):
...
+            y_beta = (y_alpha / y_mean).rename("beta").to_frame()
```
fkiraly, line 607: slightly numerically more stable would be `y_mean / (y_sd **2)` since you avoid error propagation due to cancellation, but this is very minor impact imo

skpro/regression/linear/_glm.py
```diff
@@ -182,11 +314,21 @@ def __init__(
         disp=False,
         max_start_irls=3,
         add_constant=False,
+        family="0",
```
fkiraly, line 317: there is actually no condition that triggers the warning. The correct condition would be, "user passes a positional argument to `__init__`", but that is sth we cannot easily detect.

Unless you have an idea how to do that, I would suggest we use the final defaults already and not `"0"` etc, and we always raise the warning.

skpro/regression/linear/_glm.py
```diff
@@ -201,6 +343,49 @@ def __init__(
...
+        self._full_output = self.full_output
+        self._disp = self.disp
+        self._max_start_irls = self.max_start_irls
+        self._add_constant = self.add_constant
+
+        from warnings import warn
+
+        warn(
+            "Note: in `GLMRegressor`, the sequence of the parameters will change "
+            "in skpro version 2.4.0. It will be as per the order present in the"
```
fkiraly, line 380: the current version is 2.3.1, so the earliest we can do this is 2.5.0.

fkiraly (approved):

PR #386, opened 2024-06-13
[BUG] - bugfix when `None` was specified for `max_iter` parameter in sklearn regressors

### LINKED ISSUE #385: [BUG] `sklearn.utils._param_validation.InvalidParameterError` thrown when using max_iter parameter in various sklearn regressors
**Describe the bug**
```
sklearn.utils._param_validation.InvalidParameterError: The 'max_iter' parameter of BayesianRidge must be an int in the range [1, inf). Got None instead.
```
This error occurs when using the `max_iter` parameter in various regression methods. It seems like `None` is being set as the default and sklearn introduced a new validation to disallow `None` as a possible value

**To Reproduce**

To reproduce:
use the check_estimator functionality on a regressor, either `BayesianRidge` or `ARDRegression`. Note that this requires the latest version of scikit-learn
```
from skpro.regression.linear import BayesianRidge
from skpro.utils.estimator_checks import check_estimator

check_estimator(BayesianRidge)
```

**Versions**
<details>
Requires sklearn version 1.5

julian-fong:
#### Reference Issues/PRs

Fixes #385 . 

@fkiraly I couldn't specify `inf` as a `max_iter` since the allowed values range between [1, inf) with inf as an open interval. Instead I chose 300 because thats what the docstring said previously. Let me know if you want something else set as the default value.

### DIFF
--- a/skpro/regression/linear/_sklearn.py
+++ b/skpro/regression/linear/_sklearn.py
@@ -19,8 +19,8 @@ class ARDRegression(_DelegateWithFittedParamForwarding):
 
     Parameters
     ----------
-    max_iter : int, default=None
-        Maximum number of iterations. If `None`, it corresponds to `max_iter=300`.
+    max_iter : int, default= 300
+        Maximum number of iterations.
 
     tol : float, default=1e-3
         Stop the algorithm if w has converged.
@@ -90,7 +90,7 @@ class ARDRegression(_DelegateWithFittedParamForwarding):
 
     def __init__(
         self,
-        max_iter=None,
+        max_iter=300,
         tol=1e-3,
         alpha_1=1e-6,
         alpha_2=1e-6,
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
I see, yes, that makes sense, given that 300 is the previous default, and we do not want default behaviour to change.

That is, this change is in line with deprecation policy while my original suggestion to setting the default to inf is not.

Minor, non-blocking issue is the space after `default= `

skpro/regression/linear/_sklearn.py
```diff
@@ -19,8 +19,8 @@ class ARDRegression(_DelegateWithFittedParamForwarding):
 
     Parameters
     ----------
-    max_iter : int, default=None
-        Maximum number of iterations. If `None`, it corresponds to `max_iter=300`.
+    max_iter : int, default= 300
```
fkiraly, line 22:
```suggestion
    max_iter : int, default=300
```

skpro/regression/linear/_sklearn.py
```diff
@@ -188,10 +188,9 @@ class BayesianRidge(_DelegateWithFittedParamForwarding):
 
     Parameters
     ----------
-    max_iter : int, default=None
+    max_iter : int, default= 300
```
fkiraly, line 191:
```suggestion
    max_iter : int, default=300
```

PR #392, opened 2024-06-16
[ENH] refactor `datatypes` mtypes - checkers, converters

fkiraly:
This PR refactors the data type specifications and converters to classes.

Implements https://github.com/sktime/sktime/issues/3512, related to https://github.com/sktime/sktime/issues/2957.

Contains:

* a base class for mtypes, `BaseDatatype`, to replace the more ad-hoc dictionary design
* a complete refactor of the `Table` mtype submodule to this interface
* a base class for converters, `BaseConverter`, also replacing a dictionary based design
* a partial refactor of the `Table` related converters to this interface
* a full refactor of the public framework module with `check` and `convert` logic, in `datatypes`, to allow extensibility with this design

Partial mirror in `skpro` of https://github.com/sktime/sktime/pull/6033


julian-fong:
Question : does that mean something like `pandas` would have a `BaseConverter` and a `BaseDatatype`? What is the intuition for separating it into two different classes?

fkiraly:
> does that mean something like `pandas` would have a `BaseConverter` and a `BaseDatatype`?

Yes.

> What is the intuition for separating it into two different classes?

The intuition is that conversions are indexed by ordered pairs of mtypes, and extensibility.

"ordered pairs"

If you want to put the conversions into classes, you have two "reasonable" options that I can see:

* conversion goes into a class together with the check. In this case *multiple* conversions go into the "mtype" class.
* conversion goes into its own class.

The first choice has some odd implications, e.g., do you put it in the "from" or "to" class? You also suddenly end up coupling dependencies, e.g., where do you put ta conversion between mtypes with two different soft dependencies? You now have code that uses soft dependency 2 in a class that has main soft dependency 1, etc. Sounds unpleasant and difficult to maintain to me.

"extensibility"

Not all conversions are directly implemented, as they grow combinatorially and "exotic" data types are rarely converted to each other. Sometimes developers add direct conversions without modifying the original mtypes. If a conversion is its own class, it is possible to add this without modifying anything else - like with the estimator classes, a new class is added, no class or register is modified.

julian-fong:
Right - so if each conversion is its own class, i.e from `x` to `y`, we don't clog up `BaseDatatype` `x` with various conversion methods from other different datatypes. That does make sense. 

Follow up question: could you give an example of how a `BaseConverter` would look like? Would the naming convention be something like `PandasPolarsConverter`, if we're marking that it converts pandas to polars

julian-fong:
As a follow up point, I looked at some of `sklearn`'s `set_output` functionality on how the user given `transform` specification determines which data container to use. Both polars and pandas 'adapter' classes are created with specific `create_container` methods that allows automatically converts the input dataframe into the desired output mtype. We could potentially leverage the same idea and put that down as part of the `BaseConverter` depending on how we want to use that class.

So as part of the `predict_*` function, we would use which data container the user has specified via `set_output` and then retrieve the corresponding skpro 'adapter' and corresponding methods to convert it to the desired mtype. 

I'll edit the design doc to highlight that soon

fkiraly:
> Follow up question: could you give an example of how a `BaseConverter` would look like? Would the naming convention be something like `PandasPolarsConverter`, if we're marking that it converts pandas to polars

fully worked examples can be found in https://github.com/sktime/sktime/pull/6033

fkiraly:
FYI @julian-fong, this should be ready now

julian-fong:
Looks good! Lets merge this and incorporate some of the changes from #399 into this. I'm proposing we move the `create_container` functions from the adapters and put it into their corresponding `BaseTable` objects

fkiraly:
Can you quickly explain what `create_container` does?

julian-fong:
Somewhat analogous to sklearn's create_container, it transforms the input dataframe into the desired container. 

So if the user specifies polars in their transform, all inputted dataframes will be converted into polars dataframes using the polars' create_container function

fkiraly:
I see - but there is no corresponding functionality in this PR - that would be the converters

julian-fong:
Another idea I had, for BaseTable objects we can include a tag specifying whether or not they have set_output functionality. So right now pd.DataFrame and polars Eager table would have this tag set as True and others false. This tag could be useful if other future modules use this tag for any other features eg. the SUPPORTED_OUTPUT array inside set_output.py

julian-fong:
> I see - but there is no corresponding functionality in this PR - that would be the converters

Right. So i'm proposing that after this is merged, i can port those functions over and delete them from create_container (removing the Pandas and Polars adapters).

fkiraly:
> Another idea I had, for BaseTable objects we can include a tag specifying whether or not they have set_output functionality. So right now pd.DataFrame and polars Eager table would have this tag set as True and others false

I'm not quite sure whether I follow - can you define what it means to "have set_output functionality"?

fkiraly:
> > I see - but there is no corresponding functionality in this PR - that would be the converters
> 
> Right. So i'm proposing that after this is merged, i can port those functions over to their corresponding BaseTable classes and delete them from create_container (removing the Pandas and Polars adapters).

Would it not be BaseConverter, rather?

julian-fong:
> > Another idea I had, for BaseTable objects we can include a tag specifying whether or not they have set_output functionality. So right now pd.DataFrame and polars Eager table would have this tag set as True and others false
> 
> I'm not quite sure whether I follow - can you define what it means to "have set_output functionality"?

Sorry for the confusion, basically what I mean is whether or not a certain `BaseTable` has the capability of outputting its table format as an output. i.e what polars eager table and pandas dataframe have right now

julian-fong:
> > > I see - but there is no corresponding functionality in this PR - that would be the converters
> > 
> > 
> > Right. So i'm proposing that after this is merged, i can port those functions over to their corresponding BaseTable classes and delete them from create_container (removing the Pandas and Polars adapters).
> 
> Would it not be BaseConverter, rather?

I see. Yes let's put this as part of BaseConverter. I can start building some converter classes for both pandas dataframes and polars eager (the `create_container` functions from 399 will be ported under here). I will also create some `BaseConverters` for the other current skpro mtypes once this is merged

julian-fong:
@fkiraly what if we did away with the "from_type" completely? This would automatically reduce the number of converter classes from n^2 to n. 

If we restrict ourselves to the realm of the regression mtypes (the one I'm the most familiar with at the moment), 

The current possible mtypes 
```
ALLOWED_MTYPES = [
    "pd_DataFrame_Table",
    "pd_Series_Table",
    "numpy1D",
    "numpy2D",
    "polars_eager",
]
```

would then be have 5 different `BaseConverter` objects `ConverterPandasTable`, `ConverterSeriesTable`, `ConverterNumpy1DTable`, `ConverterNumpy2DTable` and "ConverterPolarsEagerTable".

All of pandas series/dataframes, polars dataframes, and numpy all follow (somewhat) the same data container convention. They have values, indices, and columns (or some combination of the sort). All we would need to do, is check to see if "from_container" has indices, columns, and values and then pass them along if the "to_container" supports it. 

Example 1: converting pandas to numpy -> we wouldn't have either indices or columns so all we need to do is pass along the values.

Example 2: converting pandas to polars -> polars has columns and values, so if the "from_container" has both columns and values, we will pass it along. Indices are not supported so we won't pass them along (unless we write some extra code to do so as a new column).

In each of the methods, we would check whatever "from_container" has and pass it along if the "to_container" in the `_convert` method.

The cons of this would be: how feasible this could be if there comes along any mtypes that break the 'convention' above. 

I'm still working on improving my understanding of the datatypes portion of `skpro` so maybe this idea may be too naive.

note: this is also my understanding of how `sklearn` does their output container functionality: see [link](https://github.com/scikit-learn/scikit-learn/blob/379a2f196f8794bc18289c37e99e13058238a08b/sklearn/utils/_set_output.py#L152)

fkiraly:
> @fkiraly what if we did away with the "from_type" completely? This would automatically reduce the number of converter classes from n^2 to n.

Thought about that for precisely this reason. There are two main arguments against:

* it would make the code less extensible. Adding one conversion path would require changing an existing class, as opposed to adding a new one.
* coupling of soft dependencies - each converter class would accumulate the union of soft dependencies from all data containers

fkiraly:
@julian-fong, the design for converters is implemented - and two classes, one which contains a single conversion, and one which contains multiple. Review appreciated, before moving to translating the entire module.

fkiraly:
Failure is due to inconsistent handling of soft dependencies in checkers and converter dictionaries - this needs to be looked at, likely arising from the `TableIdentity`, because it loops over all mtypes, even though those requiring soft dependencies.

fkiraly:
any idea why this is failing on 3.12?

julian-fong:
> any idea why this is failing on 3.12?

i checked the soft deps i think its because `pyarrow` is missing on the list. Could be that?

skpro/datatypes/_base/_base.py
```diff
@@ -0,0 +1,357 @@
...
+            Whether obj is of this data type.
+        msg : str, only returned if return_metadata is True.
+            Error message if obj is not of this data type.
+        metadata : instance of self only returned if return_metadata is True.
+            Metadata dictionary.
+        """
+        self_params = self.get_params()
+
+        need_check = [k for k in self_params if self_params[k] is not None]
+        self_dict = {k: self_params[k] for k in need_check}
```
julian-fong, line 83: Just a quick question - what is this code used for exactly?

julian-fong:
Would we need extension templates for users to contribute new `BaseConverter`s and `BaseTable`s?

fkiraly:
> Would we need extension templates for users to contribute new `BaseConverter`s and `BaseTable`s?

Yes, of course! Once the design is final, I'd say.

skpro/datatypes/_base/_base.py
fkiraly: it's a feature, where you instantiate the type with arguments. This defines a subtype, e.g., "numpy 2D without nans", and then checks against the subtype, rather than just the overall type.

julian-fong:
> > Would we need extension templates for users to contribute new `BaseConverter`s and `BaseTable`s?
> 
> Yes, of course! Once the design is final, I'd say.
> 
> Also, in the final verison we should add precise specifications for each mtype in the docstring, and list them on an API reference page.

Maybe we should include these specifications in a separate issue, along with any other new information that could be written inside the documentation of `skpro`

fkiraly:
agree! do you want to open it?

PR #396, opened 2024-06-17
[ENH] Pareto distribution

sukjingitsit:
#### Reference Issues/PRs

Addresses #22 for the Pareto distribution

Implements some methods for the Pareto distribution

No

Current state of code, and how to add energy

No

#### Any other comments?

No

##### For all contributions
- [Yes] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
- [Yes] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
- [NA] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [NA] I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant `Examples` section.
- [NA] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured

### DIFF
--- a/skpro/distributions/pareto.py
+++ b/skpro/distributions/pareto.py
@@ -12,13 +12,13 @@ from skpro.distributions.base import BaseDistribution
 class Pareto(BaseDistribution):
     r"""Pareto distribution (skpro native).
 
-    The scale :math:`x_m` is represented by the parameter ``xm``,
+    The scale is represented by the parameter ``scale``,
     and the Pareto index (or shape parameter) :math:`\alpha`
     by the parameter ``alpha``.
 
     Parameters
     ----------
-    xm : float or array of float (1D or 2D), must be positive
+    scale : float or array of float (1D or 2D), must be positive
         scale of the Pareto distribution
     alpha : float or array of float (1D or 2D), must be positive
         shape of the Pareto distribution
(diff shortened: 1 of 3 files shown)


skpro/distributions/pareto.py
```diff
@@ -0,0 +1,161 @@
...
+class Pareto(BaseDistribution):
+    r"""Pareto distribution (skpro native).
+
+    The scale :math:`x_m` is represented by the parameter ``xm``,
+    and the Pareto index (or shape parameter) :math:`\alpha`
+    by the parameter ``alpha``.
+
+    Parameters
+    ----------
+    xm : float or array of float (1D or 2D), must be positive
```
fkiraly, line 21: xm is as it is on wikipedia, but I think it is confusing since there is another parameter x, below. Shall we name it `scale` instead?

fkiraly (changes requested):
Looks good. May I request a parameter rename, as above?

skpro/distributions/pareto.py
sukjingitsit: Sure, that would be nice. I just used the Wikipedia convention initially as different websites had different names for these two parameters.

sukjingitsit: I have implemented the parameter name change from xm to scale, and updated the docstring as such

fkiraly (approved):
Thanks.

Not blocking, but it would be great if you could add explicitly cdf or pdf in the docstring, so it is clear which parameterization this class is using.

You can see examples for this in other classes.

skpro/distributions/pareto.py
sukjingitsit: Have added the CDF to the docstring for clarity

PR #399, opened 2024-06-21
[ENH] Enhancing polars support by introducing `set_output`

julian-fong:
Introduces files `set_output` inside `skpro.utils` and new tests file `test_set_output` inside the `tests` folder. As part of https://github.com/sktime/enhancement-proposals/pull/34 and the notes written in my mentorship programme . 

In this pr:
- I have introduced basic functions to convert multi column pandas dataframes into single column pandas dataframes and vice versa, these are stored under the polars adapter file `skpro.datatypes._adapter.polars`. In the polars adapter file, `convert_polars_to_pandas_with_index` now checks to see if there was melted multi-index columns (these columns will be denoated via "__foo__bar__" convention) and `convert_pandas_to_polars_with_index` now checks to see if there are multi-index columns inside the pandas DataFrame (like in `predict_interval` and `predict_quantile`. If so, then we will melt down these multi-index columns into single-level columns before converting into a polars dataframe. 
- created `skpro.utils.set_output.check_output_config` to ensure that transformations set by the users are aligned with available `skpro` output data containers. 
- created `skpro.utils.set_output.transform_output` in order to convert the resulting DataFrame into user specified or default data containers. `transform_output` acts like a wrapper around the ordinary `convert` function, but instead it also checks whether to convert based upon the user specified `mtype` or to leverage the default original `mtype` seen in `fit`
- I have introduced `_config` inside `BaseProbaRegressor` and a new function `set_output` which mirrors `sklearn`'s `set_output` for familiarity.

Relates to #342 #449


julian-fong:
@fkiraly I have attempted to create this branch under the branch containing #370, hopefully it worked properly? 😃

julian-fong:
Next action item here is to write technical implementation of each of the `_predict_*` functions as polar frames inside the design doc - will update that soon

skpro/regression/base/_base.py
```diff
@@ -838,3 +847,27 @@ def _check_alpha(self, alpha, name="alpha"):
             raise ValueError(f"values in {name} must be unique, but found duplicates")
 
         return alpha
+
+    def set_output(self, transform):
```
VascoSch92, line 851: should you not check that transform is in {"polars", "pandas", None} ?

julian-fong: Good point! The reasoning I left None out as a possible transform value is because I could not think of an typical situation where the user would need to specify the transform parameter back to None. The intuition of the None being the default value was to simulate the scenario where none the `set_output` code would be used, if that makes sense. I believe that setting the default to 'pandas' would achieve the same effect to a degree

fkiraly: I think it should be possible to set to the default behaviour, which is input mtype = output mtype. That could be `None`, although maybe another str might be more intuitive?

skpro/tests/test_set_output.py
```diff
@@ -0,0 +1,175 @@
+import pandas as pd
```
fkiraly, line 1: should this not be in the `regression` folder?

skpro/utils/set_output.py
```diff
@@ -0,0 +1,157 @@
...
+def check_n_level_of_dataframe(X_input, index_pandas=False):
+    """Convert function to check the number of levels inside a pandas Frame.
+
+    Parameters
+    ----------
+    X_input : polars or pandas DataFrame
+        A given polars or pandas DataFrame. Note that the polars portion of this
+        code requires the soft dependencies polars and pyarrow to be installed
+
+    index_pandas : bool
```
fkiraly, line 27: strange choice of parameter, unintuitive. I would say, change either to string (`"index"`, `"columns"`), or require the user to pass an index in the first place.

skpro/tests/test_set_output.py
julian-fong: yes - will move it, should i rename it to something like `test_set_output_regression`?

skpro/regression/base/_base.py
julian-fong: Good idea - lets replace `None` with `defaults`. `sklearn` does the same. I'm not 100% sure what the `defaults` value in sklearn does for its `set_output` but we can follow the same convention nonetheless

skpro/utils/set_output.py
julian-fong: perhaps we rename is to `axis` and specify 0 or 1?

fkiraly (changes requested):
Question about the design: there already is a `convert` and `convert_to` utility in the `datatypes` module. This will not change with #392, so why are we not using those?

julian-fong:
Do you mean inside the polars / pandas container adapters? We could use the `convert` utility there but I thought it would be simpler/faster if we used the standard dataframe caller.

fkiraly:
Hm, that may be correct, but it introduces repetition/duplication, so my comment is in parts about being DRY. The idea behind the `convert` utility is that checking and conversions are localized in one place, that also ensures that there are no two slightly inconsistent definitions of the output type. Hopefully, there is also not too much computational overhead to it - if there is, we need to think about why.

julian-fong:
@fkiraly Inside the new function `set_output(self, transform):`, i am thinking we could include another parameter named `include_index` which is a boolean value. If set as `True` and transform `polars`, then we would return the index from the pandas dataframe as a column named `__index__`

julian-fong:
> Hm, that may be correct, but it introduces repetition/duplication, so my comment is in parts about being DRY. The idea behind the `convert` utility is that checking and conversions are localized in one place, that also ensures that there are no two slightly inconsistent definitions of the output type. Hopefully, there is also not too much computational overhead to it - if there is, we need to think about why.

Agreed - I think we can easily utilize the corresponding `_convert` function. I'll make this change in a different pr once #392 is merged

skpro/utils/set_output.py
```diff
@@ -0,0 +1,71 @@
...
+    SUPPORTED_OUTPUTS.append("polars")
+
+
+SUPPORTED_OUTPUT_MAPPINGS = {
+    "pandas": ("pd_DataFrame_Table", "Table"),
+    "polars": ("polars_eager_table", "Table"),
+}
+
+
+def check_output_config(estimator):
```
fkiraly, line 20: I would make this function and the next one private, to avoid user calls and allowing us to make changes quickly

skpro/utils/set_output.py
```diff
@@ -0,0 +1,71 @@
...
+        valid = True
+        output_config["dense"] = SUPPORTED_OUTPUT_MAPPINGS[transform_output]
+    else:
+        valid = False
+
+    return valid, output_config
+
+
+def transform_output(
+    obj, valid, from_type, default_to_type, default_scitype, output_config, store
```
fkiraly, line 53: this as many parameters - this is a code smell.

I would suggest to remove the `convert`-call, and only use this function to get the from/to type

skpro/regression/base/_base.py
```diff
@@ -38,6 +39,10 @@ class BaseProbaRegressor(BaseEstimator):
         "C_inner_mtype": "pd_DataFrame_Table",
     }
 
+    _config = {
+        "transform": "default"
```
fkiraly, line 43: maybe transform_output?

fkiraly (changes requested):
Looks great!

I would request two changes:
* make the new utilities private
* decouple `transform_output` from the output, and call it only if the setting is not the default

fkiraly:
let me know in case you need a review

julian-fong:
Another review would be quite helpful.. I think I'll try and re-write the description to be more clear what the goal of the pull request is and whats being added in the PR.

I'll also note that the new code that supports mi-columns and si-columns/indxes for polars exists inside `adapters.polars` and not inside the table conversion utilities. Instead, the polars conversion utilities import the conversion methods from `adapters.polars` and uses them to convert the pandas dataframes and polars dataframes. 

So if needed, I can separate the polars conversion methods inside `adapters.polars` into different functions:
- One for si column pandas to polars (for the `table` scitype)
- One for si column polars to pandas (for the `table` scitype)
- A new conversion method for mi column pandas to polars (for any new mi abstract data types that we decide upon)
- A new conversion methods for mi column pandas to polars i.e reverse mapping (for any new mi abstract data types that we decide upon)
All of these methods will still be inside `adapters.polars` unless you have another recommendation on the placement of these methods. Either way, this should allow the scitype conversion methods to be firmly independent from each other.

I'll also note that although `predict_var` is only single column, there are still some failures in the CI, because one of the notebook-examples (the one with `plot_crossplot_std`)

julian-fong:
I remember asking you before why in the base [code](https://github.com/sktime/skpro/blob/e57630d6e0e3709c86d6e2b1d7cba1ea4b0996a6/skpro/regression/base/_base.py#L163) the output conversion only existed inside the `predict` function and nothing else. Could you kindly remind me why?

And given that reasoning, do you think it makes sense to have the other `predict_*` functions have this line of code also? or should it always be returned in the same data container as the input. If this is indeed the case, we can always allow data container type changes in these methods via `set_output`. But most likely in baby steps first since this is quite a challenging design question to solve !

fkiraly:
From our discussion, summarizing what I think the key points for this PR, to avoid it moving around with minor changes:

* a clear specification and use cases written down for `set_output` and the new config field, what it should do. Scenarios that we need to cover are `pandas` and `polars` inputs both.
* this should, in particular, include a specification for what comes out of the proba methods if a `polars` output is produced
* I would then add a section on how these outputs are getting mapped onto mtypes, and the `datatypes` module, a precise suggestion.

I am happy to write these, if you feel you are stuck - please let me know. Or, if they exist somewhere, links would be appreciated.

> I'll also note that the new code that supports mi-columns and si-columns/indxes for polars exists inside `adapters.polars` and not inside the table conversion utilities.

I see - I would though not suggest to split here too much for now, instead we could just add the 

> I'll also note that although `predict_var` is only single column, there are still some failures in the CI, because one of the notebook-examples (the one with `plot_crossplot_std`)

This seems a bit suspicious, why would these changes impact the plotting function at all? I was under the impression that all existing code should run unchanged. Do you have an explanation for the failure?

fkiraly:
> I remember asking you before why in the base [code](https://github.com/sktime/skpro/blob/e57630d6e0e3709c86d6e2b1d7cba1ea4b0996a6/skpro/regression/base/_base.py#L163) the output conversion only existed inside the `predict` function and nothing else. Could you kindly remind me why?

That is because the different contracts of the underscore-methods, which are hopfully correctly stated in the extension template:

* `_predict` gets input as `X_inner_mtype` and `y_inner_mtype`. It should return an output as `y_inner_mtype` type.
* `_predict_quantiles` ets gets input as `X_inner_mtype` and `y_inner_mtype`. It needs to return the output in the specific `pd.DataFrame` based format, there is currently only a single choice for this.

Due to this, in the case of `_predict`, we need to convert the output too, if the returned `y_pred` should be of the same type as `y` seen in `fit`; but we do not need to do so for `_predict_quantiles`.

fkiraly:
> And given that reasoning, do you think it makes sense to have the other `predict_*` functions have this line of code also?

Yes, but weren't you already adding such a conversion in this PR? Which I think makes sense.

julian-fong:
> * `_predict_quantiles` ets gets input as `X_inner_mtype` and `y_inner_mtype`. It needs to return the output in the specific `pd.DataFrame` based format, there is currently only a single choice for this.

Ah I see, so do you think in your opinion for `predict_interval`, `predict_quantiles`, and `predict_var`, that (for now) it should not trigger any `convert` function unless specified by the user via `set_output`? i.e

Scenario 1a: User passes in a pandas DataFrame as `X_inner_mtype` and `y_inner_mtype` and does not utilize `set_output`, thus the output is returned as `pd.DataFrame`

Scenario 1b: User passes in a polars DataFrame as `X_inner_mtype` and `y_inner_mtype` and does not utilize `set_output`, thus the output is returned as `pd.DataFrame`

Scenario 2a: pandas DataFrame as `X_inner_mtype` and `y_inner_mtype` and utilizes `set_output` for `polars`, thus the output is returned as `pl.DataFrame`

Scenario 2b: polars DataFrame as `X_inner_mtype` and `y_inner_mtype` and utilizes `set_output` for `polars`, thus the output is returned as `pl.DataFrame`

instead of automatically trying to convert it back to the input type of `X_inner_mtype` and `y_inner_mtype` i.e

Scenario 1: User passes in a pandas Series as `y_inner_mtype` and does not utilize `set_output`, thus the code will attempt to also output is returned as `pd.Series` due to the passed input `y_inner_mtype`

Scenario 2: User passes in a polars DataFrame as `X_inner_mtype` and `y_inner_mtype` and does not utilize `set_output`, thus the code will attempt to also output is returned as `pl.DataFrame` due to the passed input `y_inner_mtype`

fkiraly:
I would do the second, not the first set of scenarios, because it is consistent with input/output behaviour as current. The quoted line is just explaining the status quo.

I think once we have the datatype extension though, it will not be difficult to switch between the two.

julian-fong:
> I would do the second, not the first set of scenarios, because it is consistent with input/output behaviour as current. The quoted line is just explaining the status quo.
> 
> I think once we have the datatype extension though, it will not be difficult to switch between the two.

I see, then it would be a good idea to write out the various possible data container types for the `proba` scitype to facilitate this. Maybe polars and pandas dataframes would be enough for now?

julian-fong:
Part of these key points can be found in the [design doc](https://github.com/sktime/enhancement-proposals/blob/f942264f9fa92b61eeaade049e1efca0c4435516/steps/24_skpro_polars_support_integration/step.md)
> * a clear specification and use cases written down for `set_output` and the new config field, what it should do. Scenarios that we need to cover are `pandas` and `polars` inputs both.
This is covered in section 5.2. The main use cases for `set_output` would be 'pandas' and 'polars', with a third option named 'default' which is to indicate default behaviour.
> * this should, in particular, include a specification for what comes out of the proba methods if a `polars` output is produced
Polars dataframe outputs are described in section 3 for proba methods `predict_interval`, `predict_quantiles`, and `predict_var`
> * I would then add a section on how these outputs are getting mapped onto mtypes, and the `datatypes` module, a precise suggestion.
I may need some help for how the outputs get mapped onto mtypes and scitypes, i've described a solution inside section 5.3. Some help regarding how to design the mapping method, how we can integrate data container support for the `Proba` scitype and various scenarios would be appreciated.

PR #410, opened 2024-06-26
[ENH] Multiclass classification reduction using Histograms

### LINKED ISSUE #378: [ENH] proba regression: reduction to multiclass classification
From the discussion today, a short design for a reducer to multiclass classification mentioned in #7.

Parameters are:

* an `sklearn` classifier `clf` capable of multiclass classification
* a `bins` arg, default = 10. Possible values are int, or an ordered list of float.

The algortihm does as follows:

* if `bins` is int, replaces this arg internally by that many bins, at the bins + 1 equally spaced quantiles of the empirical training distribution.
* sorts the training labels into a multiclass label according to which bin it is in
* in `fit`, fits `clf` to this binned training data
* in `predict_proba`, uses `clf.predict.proba` to obtain class probabilities, and uses these together with the bins from `bins` to obtain a `Histogram` distribution

One could also think about another algorithm where the bins are cumulative, i.e., being contained in the bin defined by lowest point to i-th bin. This is also valid but one needs to be  careful that the resulting cdf is monotonic. Could be a choice of strategy.

FYI @ShreeshaM07, @SaiRevanth25.

ShreeshaM07:
#### Reference Issues/PRs

fixes #378 

It adds an adapter to sklearn multiclass classifiers and expresses them as a Histogram probability distribution.

No

##### For all contributions

##### For new estimators


skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
```diff
@@ -0,0 +1,144 @@
...
+
+import numpy as np
+import pandas as pd
+
+from skpro.distributions import Histogram
+from skpro.regression.base import BaseProbaRegressor
+from skpro.utils.sklearn import prep_skl_df
+
+
+class SklearnProbaClassifier(BaseProbaRegressor):
```
fkiraly, line 13: name should be more descriptive

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
```diff
@@ -0,0 +1,144 @@
...
+    _tags = {
+        "authors": ["ShreeshaM07"],
+        "maintainers": ["ShreeshaM07"],
+        "capability:multioutput": False,
+        "capability:missing": True,
+    }
+
+    def __init__(self, clf, bins=10):
+        self.clf = clf
+        if isinstance(bins, int) or isinstance(bins, np.integer):
```
fkiraly, line 25: never change `self.x` to be sth different than arg `x`. If you want to do this, use `self._bins`.

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        from sklearn import clone
+
+        self.clf_ = clone(self.clf)
+        self._y_cols = y.columns
+
+        X = prep_skl_df(X)
+        y = prep_skl_df(y)
+
+        if isinstance(y, pd.DataFrame) and len(y.columns) == 1:
```
fkiraly, line 56: some binning should happen here

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
ShreeshaM07: Yes I did not notice that thanks.

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
ShreeshaM07: I wasn't very sure of what to name it so I just named it along the lines of `SklearnProbaReg`. I'll modify the name giving it more description.

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
ShreeshaM07: I'll be looking into that next. But I am not quite sure what we want to bin here. Could you please elaborate a little on your thought.

fkiraly: The estimator should implement the algorithm from the issue description: https://github.com/sktime/skpro/issues/378#issue-2341097717

In `_fit`, that means:

1. bin the regression target `y` into bins, according to `bins`. This results in a multioutput target `y_binned`.
2. Fit the classifier to `X` and `y_binned`.

fkiraly: > What I would infer from this is
> if y is a dataframe with the output classes

I think there is a bit of a miscommunication, sorry for not explaining clearly.

The `y` arriving in `_fit` will contain floats, not categories, for instance

```
       y
0  5
1  2
2  6
3  3
```

Let's say we decide to bin with `bins = [1,4, 7]`, then internally we convert this to


```
       y_binned
0  "class2"
1  "class1"
2  "class2"
3  "class1"
```

to this we fit `clf`.

In `_predict_proba`, using `clf.predict_proba` will produce sth like


```
      class1    class2
0   0.6          0.4
1   0.3          0.7
2   0.1          0.9
```

And this we then convert to an 3 times 1 `Histogram` distribution.

ShreeshaM07: Ah I get it now thanks for the clarification.

ShreeshaM07: But again with the `int` case for `bins` we will be unclear as to where to start and end the bins array. So would it be a good idea to take it as a tuple again (bins start,bins end,number of bins) like in the Histogram distribution? Else if we want to keep it as `int` itself then I would have to do `bins` = `(min(X),max(X),number of bins inputted by user)`.

fkiraly: I think having the `int` option with min/max would enhance user experience, especially if it also has a default.

Re min/max, why not directly min/max, but 0.999 times etc?

ShreeshaM07: That is to include the min(X) and max(X) values in the bins and not to have them as the bin boundary. It is done in the same way as `pandas.cut` in case of `int`.

`int` : Defines the number of equal-width bins in the range of x. The range of x is extended by .1% on each side to include the minimum and maximum values of x.

ShreeshaM07: I have rethought about it and since I will be using `np.arange` to create the bins array in the `int` case it would be best to take it as `np.arange(min(y),maxy(y)+step,step)` where `step = (max(y)-min(y))/bins`.

fkiraly: I think this is not a good idea, using `np.arange` where `stop - start` is an approximate multiple of `step` leads to a length of the output which is not numerically stable, see the docs of `np.arange` which warn exactly of that.

Instead, I would simply use `np.linspace` with `num=bins` (in the integer case).

Another useful function to be used later on is `np.digitize`.

ShreeshaM07: Yes I have used `np.linspace` now it makes more sense and is more accurate. I will still have to do start = 0.999*min(y)and stop = max(y)*1.001 as the `pandas.cut`, which I am using to bin the y values, does not include the leftmost point of the bin so I have to increase the domain in order to consider the min(y) value.

ShreeshaM07:
@fkiraly , I have completed the classifier and all checks have passed. Let me know if you want any modifications.

docs/source/api_reference/regression.rst
```diff
@@ -159,6 +159,17 @@ Adapters to other interfaces
 
     SklearnProbaReg
 
+Adapters to other classifiers
```
fkiraly, line 162: I would call it "Reduction to probabilistic classification"

docs/source/api_reference/regression.rst
```diff
@@ -159,6 +159,17 @@ Adapters to other interfaces
...
+----------------------------
+
+.. currentmodule:: skpro.classification.sklearn_classifiers._sklearn_proba_class
```
fkiraly, line 165: the result is a regressor, even if internally it uses a classifier, so it should be part of the `regression` module.

Similarly, the name of the estimator should not be "classifier", since it implies the estimator is a classifier.

I will think of a name, perhaps you can also try to come up with sth.

skpro/classification/sklearn_classifiers/_sklearn_proba_class.py
```diff
@@ -0,0 +1,251 @@
...
+import numpy as np
+import pandas as pd
+
+from skpro.distributions import Histogram
+from skpro.regression.base import BaseProbaRegressor
+from skpro.utils.sklearn import prep_skl_df
+
+
+class MulticlassSklearnProbaClassifier(BaseProbaRegressor):
+    """A multiclass classifier fitting a histogram distribution.
```
fkiraly, line 14: as said above, the resulting estimator is a probabilistic regressor, not a multiclass classifier. It is a probabilistic regressor that fits a histogram distribution by presenting binned outcomes to a probabilistic classifier from sklearn.

Now I don't know how to make this into a short description the best way, I'm just saying the above needs to be adapted.

fkiraly (changes requested):
Looks great!

My main comment is about naming and folder structure - the result is a regressor etc, see above.

ShreeshaM07:
I have made some changes to the name and the folder structure please let me know if this is suitable.

fkiraly (changes requested):
Thx!

non-blocking
- I think the name is now very clear
- docstring is also clear - could be improved in formuations but not blocking
- `class_bin_map_` could be described better, ie, which bin, what are values

blocking: the 1D array case is not covered by tests, can be done in `get_test_params` or separately

ShreeshaM07:
> blocking: the 1D array case is not covered by tests, can be done in get_test_params or separately

Is it possible to check for 1D case in `get_test_params` itself? as I will not be able to modify any inputs for `predict` and `predict_proba` other than the classifier instance and number of bins.

fkiraly:
yes, but we know that the inputs for `y` are numbers, so we should be able to come up with a reasonable range.

We should also think of the casewhere the user specifies bins and some `y` fall outside - in this case, I think the estimator should not break but do sth sensible (move to closest bin, or discard - perhaps even as an option on what to do)

ShreeshaM07:
> yes, but we know that the inputs for `y` are numbers, so we should be able to come up with a reasonable range.

Sorry I didn't quite get how this would help in testing for 1D case in `get_test_params`. 

> We should also think of the casewhere the user specifies bins and some `y` fall outside - in this case, I think the estimator should not break but do sth sensible (move to closest bin, or discard - perhaps even as an option on what to do)

I have handled this and moved y values outside bin range to the closest bin.

fkiraly:
> Sorry I didn't quite get how this would help in testing for 1D case in `get_test_params`.

I meant, there should be a test case in `get_test_params`, where the `bins` passed ar an 1D array, currently there isn't one.

ShreeshaM07:
> > Sorry I didn't quite get how this would help in testing for 1D case in `get_test_params`.
> 
> I meant, there should be a test case in `get_test_params`, where the `bins` passed ar an 1D array, currently there isn't one.

Ahh I see. Now that the y values outside bin range are also handled it makes all the more sense to include the array case for bins. I hadn't included earlier as I did not know the range of y and that might break the code when the `bins` are pre-meditated.
Thanks for noticing.

fkiraly (approved):

ShreeshaM07:
Are we waiting on something for this merge ?

fkiraly:
no, not in particular - just in case any other devs want to give it a review

PR #421, opened 2024-07-09
[ENH] Truncated Normal distribution

ShreeshaM07:

Interfaces truncated normal distribution from scipy.

No

Yes
##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -77,6 +77,16 @@ Integer support
     Binomial
     Poisson
 
+Continuous support - truncated distributions
+--------------------------------------------
+
+.. currentmodule:: skpro.distributions
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    TruncatedNormal
 
 Non-parametric and empirical distributions
 ------------------------------------------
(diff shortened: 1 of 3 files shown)


docs/source/api_reference/distributions.rst
```diff
@@ -77,6 +77,16 @@ Integer support
...
+Continuous support - truncated distributions
+--------------------------------------------
+
+.. currentmodule:: skpro.distributions
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    TruncatedNormal
```
fkiraly, line 89: I would put it into full reals, since any float can end up in the support.

Alternatively, one could open a new category for interval support, `Beta` could also move there.

fkiraly (changes requested):
Very strange choice of parameterization in `scipy`, I agree with the choice of reparameterizing as on wikipedia.

I left a comment regarding the right spot in the API reference, would appreciate your opinion.

docs/source/api_reference/distributions.rst
ShreeshaM07: I think for now we can move it to full reals itself as we do not have full fledged TruncatedDistribution class for other distributions.

fkiraly (approved):

PR #432, opened 2024-07-22
[ENH] Syncing datatypes module `_check.py` and `_convert.py` with `sktime`

julian-fong:
As per requested by @fkiraly, we are updating the files `_check.py` and `_convert.py` to match functionality with the `sktime` library. Note that `_common.py` is exactly the same for both libraries, so no changes were needed. 

This pr also should allow list of strings to be accepted as valid `X_inner_mtype` and `y_inner_mtype`.

@fkiraly Could you please do a review to ensure that the code is correct?

Thanks!

### DIFF
--- a/skpro/datatypes/_check.py
+++ b/skpro/datatypes/_check.py
@@ -100,13 +100,18 @@ def check_is_mtype(
         if str, list of str, metadata return dict is subset to keys in return_metadata
     var_name: str, optional, default="obj"
         name of input in error messages
-
     msg_return_dict: str, one of ``"list"`` or ``"dict"``, optional, default="dict"
         whether returned msg, if returned, is a str, dict or list
 
-        *  if "list", msg is str if mtype is str, list of str if mtype is list
+        * if ``msg_return_dict="list"``,
+          returned ``msg`` is ``str`` if ``mtype`` is ``str``,
+          returned ``msg`` is ``list`` of ``str`` if ``mtype`` is ``list``
 
-        *  if "dict", msg is str if mtype is str, dict of str if mtype is list,
+        * if ``msg_return_dict="dict"``,
+          returned ``msg`` is ``str`` if ``mtype`` is ``str``,
+          returned ``msg`` is ``dict`` of ``str`` if ``mtype`` is ``list``.
+          If ``dict``, has str in ``mtype`` as key,
+          and error message for mtype as value.
 
     Returns
     -------
@@ -260,7 +265,7 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
 
 def mtype(
     obj,
-    as_scitype,
+    as_scitype=None,
     exclude_mtypes=AMBIGUOUS_MTYPES,
 ):
     """Infer the mtype of an object considered as a specific scitype.
--- a/skpro/datatypes/_check.py
+++ b/skpro/datatypes/_check.py
@@ -22,7 +22,6 @@ __all__ = [
     "check_raise",
     "mtype",
 ]
-from typing import Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


skpro/datatypes/_check.py
```diff
@@ -104,15 +104,9 @@ def check_is_mtype(
     msg_return_dict: str, one of ``"list"`` or ``"dict"``, optional, default="dict"
         whether returned msg, if returned, is a str, dict or list
 
-        * if ``msg_return_dict="list"``,
-          returned ``msg`` is ``str`` if ``mtype`` is ``str``,
-          returned ``msg`` is ``list`` of ``str`` if ``mtype`` is ``list``
+        *  if "list", msg is str if mtype is str, list of str if mtype is list
```
fkiraly, line 107: I think the docstring was better in `skpro` - could you instead revert changes and apply these changes to `sktime`?

skpro/datatypes/_check.py
```diff
@@ -263,7 +258,11 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
         raise TypeError(msg)
 
 
-def mtype(obj, as_scitype=None, exclude_mtypes=AMBIGUOUS_MTYPES):
+def mtype(
+    obj,
+    as_scitype,
```
fkiraly, line 263: why is the default being removed? This is a breaking change, so should be reverted. If inconsistent with `sktime`, consider adding the default in `sktime`.

skpro/datatypes/_convert.py
```diff
@@ -121,9 +128,14 @@ def convert(
...
 
+    # if to_type is a list, we do the following:
+    # if on the list, then don't do a conversion (convert to from_type)
+    # if not on the list, we find and convert to first mtype that has same scitype
+    to_type = _get_first_mtype_of_same_scitype(
+        from_mtype=from_type, to_mtypes=to_type, varname="to_type"
+    )
+
     # input type checks
-    if not isinstance(to_type, str):
```
fkiraly, line 125: the input check gets removed entirely - should we not check for str or lits of str instead? Might be missing in both `sktime` and `skpro`.

skpro/datatypes/_convert.py
```diff
@@ -234,24 +254,24 @@ def convert_to(
...
-            same_scitype_mtypes = [
-                mtype for mtype in to_type if mtype_to_scitype(mtype) == as_scitype
-            ]
-            if len(same_scitype_mtypes) == 0:
-                raise TypeError(
-                    "to_type contains no mtype compatible with the scitype of obj,"
-                    f"which is {as_scitype}"
-                )
-            to_type = same_scitype_mtypes[0]
+    # # if to_type is a list, we do the following:
```
fkiraly, line 257: why is this commented out? Can this not be entirely removed?

fkiraly: impacts both `skpro` and `sktime`

fkiraly (changes requested):
Great, thanks!

Added a quick review, and it seems there are a few minor things that might be worth changing both in `sktime` and `skpro`.

Also, question: are there no tests accompanying these changes?

skpro/datatypes/_convert.py
julian-fong: the comment out code was not used at all in the sktime `_convert.py` file. will remove it. I only commented it out and left it there to see if it would break anything if it was commented

skpro/datatypes/_check.py
julian-fong: I accidently removed the `None` when i got rid of the type annotations

skpro/datatypes/_check.py
julian-fong: will make a pr in `sktime` to make this change

PR #437, opened 2024-07-28
[ENH] `DummyProbaRegressor` - probabilistic dummy regressor

### LINKED ISSUE #430: [ENH] dummy supervised regressor
We should implement a `DummyProbaRegressor` with following strategies:

* `"empirical"` - default (?), simply returns the empirical unweighted distribution of training labels
* `"normal"` - returns a gaussian with empirical variance and mean of training labels

julian-fong:
Closes #430, implements dummy proba regressor with normal and empirical strategies.


fkiraly (approved):
Looks great!

Made minor changes to docstring, naming of attribute (now `distribution`), and added `_predict_var` since we basically compute it already, to avoid recomputation later.

julian-fong:
@fkiraly For a polars version for DummyProbaRegressor, would another implementation of fit be required? Since the distributions module does not support polars atm

fkiraly:
yes, let´s do that separately

Issue #440, opened 2024-08-01
[ENH] dummy supervised regressor with polars support


julian-fong:
Implement the `DummyProbaRegressor` but with complete end to end support in `skpro`. 

Some current limitations:

`fit` inside `DummyProbaRegressor` uses `skpro.distributions` which only supports pandas dataframes - needs a workaround

`predict_proba` also uses `skpro.distributions` - leading to the same issue, will need a workaround as well

@fkiraly any suggestions on how to implement?

julian-fong:
@fkiraly I've come into a problem with the currently implementation for `polars` support in skpro. 

if an estimator specifies

```
"X_inner_mtype": "polars_eager_table",
"y_inner_mtype": "polars_eager_table",
``` 

Then during the tests, pandas DataFrames will get converted into polars dataframes via `check_X` in the boilerplate code in `regression.base` **but they will lose their index**

Since the index is already lost via the boilerplate code `check_X`, it is not retrievable when calling the private methods (since the input is already in polars dataframe format without the index). This will then fail subsequent index asserts in test files after the DataFrame is converted back into a pandas DataFrame.

fkiraly:
Interesting - I thought it saved the index as a variable `__index__` if it was not a range index.

julian-fong:
I think that would be in the `sktime` implementation, we do not save the `index` anywhere currently in the boilerplate if the incoming mtype is in polars format

fkiraly:
May I suggest to try syncing the two implementations? I think the `sktime` type by @pranavvp16 stores non-range index as a reserved variable.

PR #447, opened 2024-08-03
[ENH] Add polars version of dummy proba regressor 

julian-fong:
Completes #440 and is a polars extension of #437

Adds polars version of dummy proba regressor - also to help test e2e input flow of polars tables

In this pull request:

- Introduces a Dummy probabilistic regressor that supports the x and y inner mtype `polars_eager_table`. However, since `BaseDistribution` objects are currently is not supported for `polars` DataFrames, we leverage pandas DataFrames inside both `_fit` and `_predict_proba`. Otherwise the functionality should mirror #437. 
- Introduces a `skpro.utils.polars` file which contains 2 functions `polars_split_index_values_frame` and `polars_combine_index_values_frame`. The goal of these two functions is to split and combine `__index__` columns in and out of the main polars dataFrame (if they are created via the _convert utils from pandas to polars) in order to properly do predictions or fitting.
- Adds a simple polars e2e test inside `test_polars.py`, as well as other polars tests that is related to the adapter enhancements as part of #449


julian-fong:
tests in `check_estimator` should be failing... for `_predict_*` functions i.e `quantile` and `proba`

julian-fong:
running `check_estimator` locally does incur some failures, but maybe I am doing something wrong?

fkiraly:
that's strange. Could you kindly investigate?

fkiraly:
oh, I think I got it - the estimator is in a private module. The test framework skips estimators that are private and not publicly exported.

julian-fong:
@fkiraly How do we specify to ignore 'pandas dataframe' tests for estimators that do not have pandas specified x/y mtypes?

fkiraly:
we should not skip these - it should still work with back/from conversion.

PR #449, opened 2024-08-14
[ENH] Polars adapter enhancements

julian-fong:
adds index support as part of #440 and is used to sync up polars conversion utilities between skpro and sktime.

Correponding sktime pr for polars conversion utilities is https://github.com/sktime/sktime/pull/6455/.

In this pr:

If a pandas Dataframe is a `from_type` and polars frame is a `to_type` then during the conversion, we will save the index (assumed never to be in multi-index format) and insert it as an individual column with column name `__index__`. Then the resulting pandas dataframe will be converted to a polars dataframe.

In the inverse function, if we are converting from polars dataframe to pandas dataframe, if the column `__index__` exists in the pandas dataframe post-conversion, then we will map that column to the index before returning the pandas Dataframe

After this is merged, #447 will be implemented as a `polars` only estimator. tests will also be written to check polars input end to end and pandas input and output through the polars estimator (i.e pandas input into polars estimator -> polars predictions -> pandas output)

### DIFF
--- a/skpro/datatypes/_adapter/polars.py
+++ b/skpro/datatypes/_adapter/polars.py
@@ -123,14 +123,22 @@ def convert_pandas_to_polars_with_index(
         index from pandas DataFrame will be returned as a polars column
         named __index__.
     """
+    import pandas as pd
     from polars import from_pandas
 
-    obj_index_name = obj.index.name
-    obj = obj.reset_index()
-    if obj_index_name is not None:
-        obj = obj.rename(columns={obj_index_name: f"__index__{obj_index_name}"})
-    else:
-        obj = obj.rename(columns={"index": "__index__"})
+    # if the index of the dataframe is the trivial index (i.e RangeIndex(0,numrows))
+    # we do not return an __index__ column
+    if not (
+        isinstance(obj.index, pd.RangeIndex)
+        and obj.index.start == 0
+        and obj.index.stop == len(obj)
+    ):
+        obj_index_name = obj.index.name
+        obj = obj.reset_index()
+        if obj_index_name is not None:
+            obj = obj.rename(columns={obj_index_name: f"__index__{obj_index_name}"})
+        else:
+            obj = obj.rename(columns={"index": "__index__"})
 
     pl_df = from_pandas(
         data=obj,
--- a/skpro/datatypes/_adapter/polars.py
+++ b/skpro/datatypes/_adapter/polars.py
@@ -72,6 +72,7 @@ def convert_polars_to_pandas_with_index(obj):
     for col in obj.columns:
         if col.startswith("__index__"):
             pd_df = pd_df.set_index(col, drop=True)
+            pd_df.index.name = col.split("__index__")[1]
 
     return pd_df
 
--- a/skpro/datatypes/_adapter/polars.py
+++ b/skpro/datatypes/_adapter/polars.py
@@ -34,7 +34,10 @@ def check_polars_frame(obj, return_metadata=False, var_name="obj", lazy=False):
     if _req("n_features", return_metadata):
         metadata["n_features"] = obj.width
     if _req("feature_names", return_metadata):
-        metadata["feature_names"] = obj.columns
+        if lazy:
+            metadata["feature_names"] = obj.collect_schema().names()
+        else:
+            metadata["feature_names"] = obj.columns
 
     # check if there are any nans
     #   compute only if needed
--- a/skpro/datatypes/_adapter/polars.py
+++ b/skpro/datatypes/_adapter/polars.py
@@ -69,8 +69,9 @@ def convert_polars_to_pandas_with_index(obj):
         obj = obj.collect()
 
     pd_df = obj.to_pandas()
-    if "__index__" in obj.columns:
-        pd_df = pd_df.set_index("__index__", drop=True)
+    for col in obj.columns:
+        if col.startswith("__index__"):
+            pd_df = pd_df.set_index(col, drop=True)
 
     return pd_df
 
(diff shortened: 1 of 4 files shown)


skpro/datatypes/_adapter/polars.py
```diff
@@ -46,3 +46,75 @@ def check_polars_frame(obj, return_metadata=False, var_name="obj", lazy=False):
...
+    Returns
+    -------
+    pl_df : polars DataFrame or polars LazyFrame
+        index from pandas DataFrame will be returned as a polars column
+        named __index__.
+    """
+    from polars import from_pandas
+
+    obj.reset_index()
+    obj.rename(columns={"index": "__index__"})
```
pranavvp16, line 108: How about we retain the index name too. This code won't work for data frames that already have a name for their index. Does skpro assumes that data frame will always have a default `RangeIndex` without any name @fkiraly @julian-fong ??, there may be scenarios where a index having name is passed and the conversion util will fail in this case.

How sktime handles this is by appending the name infront of the `__index__` convention. So suppose if the DataFrame has a index named `1` the util will convert it to `__index__1` in polars frame.

pranavvp16 (commented):
Left some questions about how is index name being handled in the conversion utils. The code will get the job done currently, but suppose I convert a pandas DataFrame having a index with a name to polars, and then convert it back to pandas the index name would be lost.

skpro/datatypes/_adapter/polars.py
julian-fong: thats a good point, i'll make that change then

julian-fong:
Tests are failing in jobs because the `__index__` is being returned and there is no easy way to get rid of it. i.e

```
┌───────────┬──────┐
│ __index__ ┆ a    │
│ ---       ┆ ---  │
│ i64       ┆ f64  │
╞═══════════╪══════╡
│ 0         ┆ 1.0  │
│ 1         ┆ 4.0  │
│ 2         ┆ 0.5  │
│ 3         ┆ -3.0 │
└───────────┴──────┘
``` 

is returned when it expects

```
┌──────┐
│ a    │
│ ---  │
│ f64  │
╞══════╡
│ 1.0  │
│ 4.0  │
│ 0.5  │
│ -3.0 │
└──────┘
```

pranavvp16:
@julian-fong I know why the tests are failing here, you need to change the polars fixture example to match this convention, in skpro the examples are located at `skpra/datatypes/_table/_examples.py`. What you can do here is import the conversion util and convert the pandas DataFrame that is being used

skpro/datatypes/_adapter/polars.py
fkiraly: Yes, I thought that's what is happening in the `dask` converter as well, is it?

fkiraly:
> Tests are failing in jobs because the `__index__` is being returned and there is no easy way to get rid of it. i.e
> 
> ```
> ┌───────────┬──────┐
> │ __index__ ┆ a    │
> │ ---       ┆ ---  │
> │ i64       ┆ f64  │
> ╞═══════════╪══════╡
> │ 0         ┆ 1.0  │
> │ 1         ┆ 4.0  │
> │ 2         ┆ 0.5  │
> │ 3         ┆ -3.0 │
> └───────────┴──────┘
> ```
> 
> is returned when it expects
> 
> ```
> ┌──────┐
> │ a    │
> │ ---  │
> │ f64  │
> ╞══════╡
> │ 1.0  │
> │ 4.0  │
> │ 0.5  │
> │ -3.0 │
> └──────┘
> ```
> 
> @fkiraly any preferred method on how to solve this?

My answer would be to not include an `__index__` if it is a `RangeIndex`, that is:

If you convert from `pd_DataFrame_Table` to the `polars` type:

1. check if index is `RangeIndex`
2. if yes, include no `__index__` column; otherwise, include index as `__index__[indexname]` column.

fkiraly (approved):
Looks good - @pranavvp16, can you kindly review this, too?

julian-fong:
> > Tests are failing in jobs because the `__index__` is being returned and there is no easy way to get rid of it. i.e
> > ```
> > ┌───────────┬──────┐
> > │ __index__ ┆ a    │
> > │ ---       ┆ ---  │
> > │ i64       ┆ f64  │
> > ╞═══════════╪══════╡
> > │ 0         ┆ 1.0  │
> > │ 1         ┆ 4.0  │
> > │ 2         ┆ 0.5  │
> > │ 3         ┆ -3.0 │
> > └───────────┴──────┘
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is returned when it expects
> > ```
> > ┌──────┐
> > │ a    │
> > │ ---  │
> > │ f64  │
> > ╞══════╡
> > │ 1.0  │
> > │ 4.0  │
> > │ 0.5  │
> > │ -3.0 │
> > └──────┘
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @fkiraly any preferred method on how to solve this?
> 
> My answer would be to not include an `__index__` if it is a `RangeIndex`, that is:
> 
> If you convert from `pd_DataFrame_Table` to the `polars` type:
> 
> 1. check if index is `RangeIndex`
> 2. if yes, include no `__index__` column; otherwise, include index as `__index__[indexname]` column.

Forgot to include async discussion but it is summarized as follows.

In the code: we will only ignore the addition of index if the index of the pandas DataFrame is the trivial index i.e RangeIndex(0,n_rows)

Otherwise, we will build a new column called `__index__*` using the `reset_index()` function built in pandas.

pranavvp16:
> Looks good - @pranavvp16, can you kindly review this, too?

LGTM!

julian-fong:
Thank you for the reviews!

PR #452, opened 2024-08-17
[BUG] Update `skpro.utils.git_diff` to fix issue with encoding 

julian-fong:
Matches bug fix in https://github.com/sktime/sktime/pull/6862. Credit to @fnhirwa for the fix!


fkiraly (approved):
Thanks!

Issue #340, opened 2024-05-17
[ENH] `concat` operation on distributions


fkiraly:
There should be a `concat` operation on distributions.

This will require:

* a "concatenated distribution" compositor `skpro.concat`, similar to `pd.concat`
* a dunder-like method in distributions to implement specific concat operations that allow to "flatten" concatenated distibutions of the same type

A direct consumer of this interface would be `sktime.forecasting`'s `predict_proba` vectorization.

SaiRevanth25:
Hackmd notes from the previous discussion

To view, click the dropdown arrow.

<details>
<summary>concat operation</summary>

### Example usage

```python=
d1 = Normal(mu=[[1, 2], [3, 4]], sigma=1)  # 2 x 2
d2 = Normal(mu=0, sigma = [[2, 42]])  # 1 x 2

d = concat([d1, d2], axis=0)  # 3 x 2
```

This `d` should then be the same abstract distribution (not necessarily the same object) as if constructed direcctly by

```python=
d = Normal(
    mu = [[1, 2], [3, 4], [0, 0]],
    sigma = [[1, 1], [1, 1], [2, 42]],
)  # 3 x 2
```
(the repetition of 0s and 1s is due to broadcasting)

We observe the following:

* `pd.concat([d1.mean(), d2.mean()])` is the same as `skpro.concat([d1, d2]).mean()`
* `pd.concat([d1.var(), d2.var()])` is the same as `skpro.concat([d1, d2]).var()`
* `pd.concat([d1.pdf(x1), d2.pdf(x2)])` is the same as `skpro.concat([d1, d2]).pdf(pd.concat([x1, x2]))`


### different distributions

SR - what happens if there are two different distributions, e.g., `Normal` or `Laplace`?

Example:

```python=
d1 = Normal(mu=[[1, 2], [3, 4]], sigma=1)  # 2 x 2
d2 = Laplace(mu=0, sscale=[[2, 42]])  # 1 x 2

d = concat([d1, d2], axis=0)  # 3 x 2
```

What is `d`?

FK - good question, I think it needs to be the "outer product" distribution, i.e., outer product of probability measures. This could be a separate compositee distribution object, and the same as


```python=
ConcatDistr([d1, d2], axis=0)
```

This happens whenever the two distribution types are different, i.e., we are not concating `Normal` with `Normal` or `Laplace` with `Laplace`.

`mean` and `var` behave as one would expect, same as above:

[...]

PR #499, opened 2024-11-19
[ENH] `concat` operation on distributions 

SaiRevanth25:
#### Reference Issues/PRs
 #340 

 Implements the concat operation on distributions.
See https://github.com/sktime/skpro/issues/340#issuecomment-2453301902

##### For all contributions

### DIFF
--- a/skpro/distributions/concat.py
+++ b/skpro/distributions/concat.py
@@ -1,0 +1,96 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Concat operation."""
+
+__author__ = ["SaiRevanth25"]
+
+import pandas as pd
+
+
+class concat:
+    """Concatenate the given distributions along specified axis.
+
+    Parameters
+    ----------
+    distributions : list
+        list of distributions
+    axis : {0/'index', 1/'columns'}, default 0
+        The axis to concatenate along
+
+    Examples
+    --------
+    >>> import skpro.distributions.concat as skpro
+    >>> d1 = Normal(mu=[[1, 2], [3, 4]], sigma=1)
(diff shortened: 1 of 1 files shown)


fkiraly:
changed it to draft, until it is not finished

skpro/distributions/concat.py
```diff
@@ -0,0 +1,96 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Concat operation."""
+
+__author__ = ["SaiRevanth25"]
+
+import pandas as pd
+
+
+class concat:
```
fkiraly, line 9: I think we need two things:

* a `function` `concat`, with signature like `pandas.concat`. It produces a `ConcatDistr` when passed a list of distr.
* a class `ConcatDistr`, that inherits `_HeterogenousMetaObject`
    * this looks like the current class

Although, now that I am thinking about this, it might be worth about just calling the class `concat`. Although, perhaps it is best not to do that, since it does not align with the `UpperCamelCase` naming convention.

fkiraly (changes requested):
Great! Pattern looks good.

I assume this is work in progress, so some comments:

* the `__init__` may need to have an `index` and `columns` argument.
* implementing `mean` etc is the old, deprecated extension pattern. Now you need to implement `_mean` etc, see `extension_templates/distributions`.
* there are more methods to implement, and `get_test_params` should be set.

fkiraly:
I made some changes to fix the delegation - but it is currently missing treatment of the indices.

I think what should happen in `__init__`:

* if no `index` is passed, then the resulting `index` should be as of `pd.concat`, written to self
* if no `columns` are passed, then the resulting `columns` should be as of `pd.concat`, written to self
* if either are passed, they overwrite the resulting index or columns

PR #480, opened 2024-10-09
[DOC] Adding a Notebook companion to BayesianMCMCLinearRegressor, Restructuring Estimator Folder

meraldoantonio:
#### Reference Issues/PRs

Notebook example for `BayesianMCMCLinearRegressor` estimator;  (#358) (merged)
Some renaming/restructuring of folders for both the estimator and the example notebook.

#### What does this implement/fix? Explain your changes.
Provides step by step explanation on how to use  `BayesianMCMCLinearRegressor`, accompanied by high level overview of Bayesian workflow.

#### Does your contribution introduce a new dependency? If yes, which one?
No

#### What should a reviewer concentrate their feedback on?
The usefulness and correctness of the notebook.

#### Did you add any tests for the change?
No

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
  How to: add yourself to the [all-contributors file](https://github.com/sktime/skpro/blob/main/.all-contributorsrc) in the `skpro` root directory (not the `CONTRIBUTORS.md`). Common badges: `code` - fixing a bug, or adding code logic. `doc` - writing or improving documentation or docstrings. `bug` - reporting or diagnosing a bug (get this plus `code` if you also fixed the bug in the PR).`maintenance` - CI, test framework, release.
  See here for [full badge reference](https://allcontributors.org/docs/en/emoji-key)
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

PR #500, opened 2024-11-21
[ENH] Bayesian Linear Regression using Normal Conjugate Prior

meraldoantonio:
Implementing a new Bayesian linear regression estimator module using normal conjugate prior
with known noise variance.

No

Correctness of implementation

No

#### Any other comments?
No

##### For all contributions
- [] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)

##### For new estimators
(https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

### DIFF
--- a/docs/source/api_reference/regression.rst
+++ b/docs/source/api_reference/regression.rst
@@ -195,3 +195,14 @@ Base classes
     :template: class.rst
 
     BaseProbaRegressor
+
+Bayesian
+------------
+
+.. currentmodule:: skpro.regression.bayesian.bayesian_conjugate
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BayesianConjugateLinearRegressor
--- a/skpro/regression/bayesian/bayesian_conjugate.py
+++ b/skpro/regression/bayesian/bayesian_conjugate.py
@@ -20,7 +20,6 @@ class BayesianConjugateLinearRegressor(BaseProbaRegressor):
 
     _tags = {
         "authors": ["meraldoantonio"],
-        "python_dependencies": ["scipy", "matplotlib"],
         "capability:multioutput": False,
         "capability:missing": True,
         "X_inner_mtype": "pd_DataFrame_Table",
--- a/skpro/regression/bayesian/bayesian_conjugate.py
+++ b/skpro/regression/bayesian/bayesian_conjugate.py
@@ -1,0 +1,187 @@
+"""Bayesian Linear Regression Estimator for Probabilistic Regression."""
+
+__author__ = ["meraldoantonio"]
+
+import numpy as np
+
+from skpro.distributions import Normal
+from skpro.regression.base import BaseProbaRegressor
+
+
+class BayesianConjugateLinearRegressor(BaseProbaRegressor):
+    """Bayesian probabilistic estimator for linear regression.
+
+    This estimator uses a Normal conjugate prior for the coefficients.
+    It assumes a known noise variance for the Gaussian likelihood.
+    Upon inference, it returns a Normal distribution for the posterior
+    as well as for predictions.
+    """
+
+    _tags = {
+        "authors": ["meraldoantonio"],
+        "python_dependencies": ["scipy", "matplotlib"],
+        "capability:multioutput": False,
+        "capability:missing": True,
+        "X_inner_mtype": "pd_DataFrame_Table",
+        "y_inner_mtype": "pd_Series_Table",
+    }
+
+    def __init__(self, prior_mean=None, prior_cov=None, noise_variance=1.0, prior=None):
+        """Initialize the Bayesian Linear Regressor with priors.
+
+        Parameters
+        ----------
+        prior_mean : ndarray, optional
+            Mean vector for the Normal prior for coefficients. Default is None.
+        prior_cov : ndarray, optional
+            Covariance matrix for the Normal prior for coefficients. Default is None.
+        noise_variance : float, optional
+            Known variance of the Gaussian likelihood. Default is 1.0.
+        prior : Normal, optional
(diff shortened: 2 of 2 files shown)


docs/source/api_reference/regression.rst
```diff
@@ -195,3 +195,14 @@ Base classes
     :template: class.rst
 
     BaseProbaRegressor
+
+Bayesian
+------------
```
fkiraly, line 200: underline should be as long as the header. I would also put it in a single section with the other regressor.

fkiraly (changes requested):
Great!

Especially kudos for getting the update right!
I have reviewed the math as well, and it looks correct!

Some comments:

* changes to notebook 3 should be avoided - that is not part of this PR
* is it not possible to set different priors for different coefficients? Might be nice to have, since these tend to be on different scales in real applications.

meraldoantonio:
Hi @fkiraly,

Thanks for the review! I’ve restructured the folders to improve organization. Now, we have a `bayesian` subfolder containing both of our new estimators. Similarly, the example notebooks have been moved into a `bayesian` subfolder inside the `examples` directory.

For this conjugate estimator, I haven’t yet enabled the setting of `mu` priors for different coefficients, as implementing this will require a more complex update formula. This will take a bit more time to complete.

Looking ahead, I suggest we create at least two versions of the conjugate estimator:
1. **Simple Version** (the current one): Assumes the prior `mu` is all zeros, and the covariance matrix is isotropic.
2. **Flexible Version**: Allows users to set both the priors `mu` and the covariance matrix `covar`.
3. A longer term goal will be to create a Bayesian Linear Regressor for which both precision and mu are unknown; this will require us to use Normal-Wishart conjugate model. However, this probability distribution is not yet available in `scipy` (it's coming in the next version).

My next goal is to work on these later versions. Once it’s ready, we could rename the current estimator to something like `SimpleBayesianConjugateLinearRegressor` to distinguish it more clearly.

docs/source/api_reference/regression.rst
```diff
@@ -195,3 +195,15 @@ Base classes
     :template: class.rst
 
     BaseProbaRegressor
+
+Bayesian
+--------
+
+.. currentmodule:: skpro.regression.bayesian.bayesian_conjugate
```
fkiraly, line 202: the `BayesianLinearRegressor` is in another folder, so will not show up here.

skpro/regression/bayesian/bayesian_conjugate.py
```diff
@@ -0,0 +1,231 @@
...
+
+class BayesianConjugateLinearRegressor(BaseProbaRegressor):
+    """Bayesian probabilistic estimator for linear regression.
+
+    This estimator uses a multivariate Normal conjugate prior for the coefficients,
+    with the prior mean fixed at zero and precision controlled by a scalar alpha.
+    It assumes a known noise precision beta for the Gaussian likelihood.
+    For inference, it returns a multivariate Normal distribution as the posterior.
+    For prediction, it returns a series of univariate Normals for each data point.
+
```
fkiraly, line 20: do you want to add math to make this precise? Also, the handling of the intercept should be discussed.

fkiraly (changes requested):
Great, only some smaller comments:

* the `BayesianLinearRegressor` is not in the folder linked to in the API ref
* non-blocking suggestions about improving the docstring

Non-blocking, to think about:

The notebook could be about Bayesian linear regression, not just about the conjugate approach. It would be informative to separately showcase the Bayes rule, and different ways to estimate/approximate the actual posterior. Conjucate and MCMC are two of a longer list (other strategies are variational and approximate).

docs/source/api_reference/regression.rst
meraldoantonio: Modified the file

skpro/regression/bayesian/bayesian_conjugate.py
meraldoantonio: I added some high-level explanation on normal conjugacy. I tried adding the math but it looks too cluttered.
Also, added information on the handling of intercept

fkiraly (changes requested):
Excellent work! My comments are mainly on documentation, PR structure, nothing content-wise.

* the notebooks should be reconciled with PR #480, otherwise I see some confusion there. I would suggest to split off the notebooks in a separate PR, with dependency graph either 500, 480 -> new PR, or 500 -> 480 -> new PR.
* in the docstring, I think it is essential to say what exactly the model assumptions are, a little bit of math would be great there, and how these map onto python parameters.
* the changes in #480 are inconsistent with the directory structure here - I would recommend to make all submodues of `bayesian` private, as long as we are not certain about the exact end state desired.

meraldoantonio:
Hi @fkiraly,

Thanks for the review! I've made the following updates:

1. Removed the notebooks and notebook artefacts from this PR. They will be included in a separate PR. I propose the following merge order:
   - Merge this PR first.
   - Followed by the MCMC PR (#480). This PR will simply restructure the MCMC imports. I'll also remove the outdated example notebook from this PR. 
   - Finally, merge the notebook PR, which will contain the two example notebooks and the artefacts.

2. Made the submodule private.

3. Updated the docstring to include mathematical details and model assumptions.

Let me know if this works!

fkiraly (approved):
Works for me, all perfect.

Maybe call the new API section not "Bayesian" but "Bayesian regressors" or similar? "Bayesian" is just a verb (and/or the name of a sunken yacht).

meraldoantonio:
Hi @fkiraly, I've modified the API name in `regression.rst` and I think this PR is ready to be merged. If all OK, could you help with the merge?  Thanks :)

PR #503, opened 2024-12-06
[BUG] Fix residual scale estimation in DoubleResidual

### LINKED ISSUE #492: [BUG] ResidualDouble regressor does not handle transforms correctly.
**Describe the bug**
In skpro.regression.residual.ResidualDouble, for the argument residual_trafo="squared", we should be taking the sqrt of the scale estimate before feeding it into the scale parameters, possibly followed by a further, distribution specific correction. 

For instance, if we're using an error dist X~T(mu=0, dof=v) distribution, then for abs E[|X|] = ComplicatedFunc(dof) * sigma, and E[X^2]=Var(X)=sigma^2 * v / (v-2)

[derivation from stack exchange](https://stats.stackexchange.com/questions/466428/expected-value-of-the-absolute-standardized-t-distribution)

For arbitrary transformations, this quickly gets tricky, even for those limited distributions. Further, arbitrary transformations can't easily be checked for monotonicity or positivity. I would advise removing that capability. 

**To Reproduce**

Running the following should yield uniform quantiles, but that totally breaks down for the squared transform.

```python
from matplotlib import pyplot as plt
import lightgbm as lgb
import numpy as np
import seaborn as sns
import pandas as pd
nn = 100_000
x_df = pd.DataFrame({"a":np.random.randn(nn), "b":np.random.randn(nn), "c":np.random.randn(nn)}).clip(-2,2)
degrees_of_freedom = 3.5
var_adj = (degrees_of_freedom / (degrees_of_freedom - 2))
# DGP
loc_param_vec = pd.Series({"a":-1, "b":1, "c":0})
log_scale_param_vec = pd.Series({"a":0, "b":0.01, "c":0.5})
loc_vec = x_df.dot(loc_param_vec)
log_scale_vec = x_df.dot(log_scale_param_vec)
dist = stats.t(df=degrees_of_freedom, loc=loc_vec, scale=np.exp(log_scale_vec))
y = pd.DataFrame(dist.rvs((2, nn)).T, index=x_df.index, columns=["y0", "y1"])
abs_reg = ResidualDouble(estimator=lgb.LGBMRegressor(verbosity=-1), estimator_resid=lgb.LGBMRegressor(verbosity=-1), distr_params={"df": degrees_of_freedom}, distr_type="t", residual_trafo="absolute", cv=KFold(n_splits=3))
sq_reg = ResidualDouble(estimator=lgb.LGBMRegressor(verbosity=-1), estimator_resid=lgb.LGBMRegressor(verbosity=-1), distr_params={"df": degrees_of_freedom}, distr_type="t", residual_trafo="squared", cv=KFold(n_splits=3))

abs_reg.fit(x_df, y["y0"])
sq_reg.fit(x_df, y["y0"])
abs_pred = abs_reg.predict_proba(x_df)
sq_pred = sq_reg.predict_proba(x_df)
abs_cdf = abs_pred.cdf(y[["y1"]])["y0"]
sq_cdf = sq_pred.cdf(y[["y1"]])["y0"]
cdf_df = pd.DataFrame({"abs": abs_cdf, "sq": sq_cdf})
sns.histplot(cdf_df, multiple="dodge", bins=10)
```

**Confusion**

I don't understand why it does yield uniform quantiles for the absolute transform for the t distribution. I would expect them to deviate from uniform because there is no degrees of freedom correction. Can anyone explain why that works?

Tacking the following code onto the end of the block above, we see that "raw" is the most uniform. That makes no sense to me.

```python

from scipy.special import gamma
def abs_t_correction(dof):
    return 2 *np.sqrt(dof - 2) * gamma((dof + 1) / 2) / (np.sqrt(np.pi) * (dof - 1) * gamma(dof / 2))

dist = stats.t(df=degrees_of_freedom, loc=loc_vec, scale=np.exp(log_scale_vec))
dist_div = stats.t(df=degrees_of_freedom, loc=loc_vec, scale=np.exp(log_scale_vec) / abs_t_correction(degrees_of_freedom))
dist_mult = stats.t(df=degrees_of_freedom, loc=loc_vec, scale=np.exp(log_scale_vec) * abs_t_correction(degrees_of_freedom))
all_3_cdf = abs_pred.cdf(pd.DataFrame({"div":dist_div.rvs(nn), "mult":dist_mult.rvs(nn), "raw":dist.rvs(nn)}, columns=["div", "mult", "raw"]))
all_3_cdf.columns = ["div", "mult", "raw"]
sns.histplot(all_3_cdf, multiple="dodge", bins=10)
```

meh2135:
#### Reference Issues/PRs
Fixes #492  

Fixes the method of moments estimation of scale parameters in ResidualDouble, and adds sample weight support to it as well.

Removed support for Cauchy and for t with df<=2 because neither have second moments (so estimating scale is impossible) and Cauchy doesn't even have a first moment.

no

* My added test
* The failing tests (clone isn't working for reasons that are very unclear to me)
* the deprecation of cauchy support because... it's breaking but what was being done before was totally invalid.

Added tests to confirm uniform quantiles for held out data when the model is correctly specified.

### DIFF
--- a/skpro/regression/base/_base.py
+++ b/skpro/regression/base/_base.py
@@ -72,7 +72,7 @@ class BaseProbaRegressor(BaseEstimator):
         else:
             return NotImplemented
 
-    def fit(self, X, y, C=None):
+    def fit(self, X, y, C=None, sample_weight=None):
         """Fit regressor to training data.
 
         Writes to self:
@@ -89,6 +89,8 @@ class BaseProbaRegressor(BaseEstimator):
         C : ignored, optional (default=None)
             censoring information for survival analysis
             All probabilistic regressors assume data to be uncensored
+        sample_weight : pandas DataFrame, same length as X, default=None
+            sample weights to fit regressor to
 
         Returns
         -------
@@ -112,13 +114,18 @@ class BaseProbaRegressor(BaseEstimator):
 
         # set fitted flag to True
         self._is_fitted = True
-
-        if not capa_surv:
-            return self._fit(X_inner, y_inner)
+        if sample_weight is None:
+            if not capa_surv:
+                return self._fit(X_inner, y_inner)
+            else:
+                return self._fit(X_inner, y_inner, C=C_inner)
         else:
-            return self._fit(X_inner, y_inner, C=C_inner)
+            if not capa_surv:
+                return self._fit(X_inner, y_inner, sample_weight=sample_weight)
+            else:
+                return self._fit(X_inner, y_inner, C=C_inner, sample_weight=sample_weight)
 
-    def _fit(self, X, y, C=None):
+    def _fit(self, X, y, C=None, sample_weight=None):
         """Fit regressor to training data.
 
         Writes to self:
@@ -130,6 +137,11 @@ class BaseProbaRegressor(BaseEstimator):
             feature instances to fit regressor to
         y : pandas DataFrame, must be same length as X
             labels to fit regressor to
+        C : ignored, optional (default=None)
+            censoring information for survival analysis
+            All probabilistic regressors assume data to be uncensored
+        sample_weight : pandas DataFrame, same length as X, default=None
+            sample weights to fit regressor to
 
         Returns
         -------
@@ -137,7 +149,7 @@ class BaseProbaRegressor(BaseEstimator):
         """
         raise NotImplementedError
 
-    def update(self, X, y, C=None):
+    def update(self, X, y, C=None, sample_weight=None):
         """Update regressor with a new batch of training data.
 
         Only estimators with the ``capability:update`` tag (value ``True``)
@@ -159,6 +171,8 @@ class BaseProbaRegressor(BaseEstimator):
         C : ignored, optional (default=None)
             censoring information for survival analysis
             All probabilistic regressors assume data to be uncensored
+        sample_weight : pandas DataFrame, same length as X, default=None
+            sample weights to fit regressor to
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


fkiraly (commented):
Nice!

Would you like a review right now, or is this still a work-in-progress draft?

Some comments:

* we should not change the core interface in the same pull request as changes to estimators. Your idea of adding a `sample_weight` arg is great, but we should deal with this separately, to avoid interaction between the changes!
* changes to the residual estimator look good and sensible. I will have to sit down and work out the math to check, but superficially fine.

skpro/regression/base/_base.py
```diff
@@ -72,7 +72,7 @@ def __rmul__(self, other):
         else:
             return NotImplemented
 
-    def fit(self, X, y, C=None):
+    def fit(self, X, y, C=None, sample_weight=None):
```
fkiraly, line 75: as said, good idea, but should be in a separate PR. Could you also open an issue to track adding sample weights?

What we should do (please copy this recommendation in an issue):

* add a tag `capability:sample_weight` that ells the user whether the weights are used non-trivially
* add tests in `TestAllRegressors` that actually passes sample weights, to see that nothing breaks. In particular, nothing should break for any estimator when passed, the current boilerplate would lead to most estimators breaking, as the argument is passed on to `_fit`!
* in particular, the boilerplate should check the tag, and pass `sample_weight` on to `_fit` only if the tag has value `True`.
* in the tests, we should check that `_fit` does have the arg `sample_weight` if the tag is set `True`.

PR #512, opened 2025-01-10
[ENH] Skew-Normal Distribution

### LINKED ISSUE #510: [ENH] Skew-Normal Distribution
@fkiraly Kindly look into this proposal!!

* The addition of a Skew-Normal Distribution to skpro's set of probability distributions would enhance its capability to model data with asymmetry.
* This would improve the flexibility and utility of skpro in real world prob. regression and Time Series Forec. where data is symm. distriuted

Implementation of Skew-Normal Distribution class:
* A class `SkewNormal` following the structure of existing distributions in `skpro`
* Parameters: `Location (mu)` , `Scale (sigma)` and `Shape (alpha)`
* Compatible with the existing interface for conditional and unconditional distributions

Key Methods:
* pdf(x) : Probability Density Function
* logpdf(x): Log of the probability density function
* cdf(x): Cumulative Distribution function
* rvs(x): Random Variable Sampling

Generalized Gaussian Distribution: Provides flexibility but with a more complex parameterization

**Additional context**

Reference papers for the same:
https://www.ine.pt/revstat/pdf/rs130105.pdf

Spinachboul:
#### Reference Issues/PRs  

Fixes #510  

This implementation adds a new `Skew-Normal` distribution as a probabilistic model in `skpro`. The inspiration comes from the paper (https://www.ine.pt/revstat/pdf/rs130105.pdf).  

The `Skew-Normal` distribution captures skewness in data, improving the flexibility of predictive distributions. Scipy already implements this as `scipy.stats.skewnorm`, which is used as a dependency to avoid redundant code.  

No new core dependencies are introduced. `scipy` is already a core dependency.


fkiraly (changes requested):
Thanks, great start!

Some comments:
* please follow the extension template carefully, all `__init__` parameters should be written to `self`, to an attribute of the same name
* properties should not be used for the `__init__` parameters, this is due to how the base class `get_params` etc works
* `loc` is a reserved attribute and should not be overridden. Another common name is `xi`, but feel free to choose anything as lon gas it is not `loc`.
* code formatting tests are failing, please address

Spinachboul:
@fkiraly 
Thanks for the suggestions
I have updated the PR by making changes in the class implementation where I have followed the extension template
But I have used `BaseDistribution` instead of `ScipyAdapter`, since that might provide more flexibility in terms of functions and general implementation.
What are views on that??

fkiraly (changes requested):
Thanks!

* I do not see a reason to use the `BaseDistribution`, the adapter does the same thing with much less code
* kindly ensure code is properly formatted, see https://www.sktime.net/en/stable/developer_guide/coding_standards.html

fkiraly (changes requested):
Thanks!

Can you kindly make sure you use the extension template? Please read it carefully. For instance:

* `__init__` parameters should be written to self
* the docstring should describe the distribution, see a few other examples in the same folder
* please follow the code formatting guide https://www.sktime.net/en/stable/developer_guide/coding_standards.html

fkiraly:
kindly diagnose the errors and fix them - it should be possible to spot what is wrong if you look at the logs.

Let us know if you need help.

Spinachboul:
@fkiraly 
Thanks for the review
There was probably an import error, now I think it should work fine
Would request for any feedback if possible!!

fkiraly (approved):
Fixed this to have it in the release.

Spinachboul:
@fkiraly 
Thanks for the changes! Meanwhile I was also thinking of having changes in the .rst pages :
image
Currently they appear like this
Maybe if we could update them , have their proper documentation.

PR #514, opened 2025-01-12
upgraded python version to < 3.13 for cyclic_boosting >= 1.4.0

munendr:


fkiraly (changes requested):
According to the install failure, this version does not exist?

PR #520, opened 2025-01-19
[DOC] Finished first two Bayesian notebooks and their artefacts

meraldoantonio:
#### Reference Issues/PRs
Provides example for Bayesian Conjugate Linear Regressor #500 

This contribution includes the first two notebooks in a planned series of four. These two notebooks cover:
1. The general theory behind Bayesian Linear Regression.
2. The conjugate prior method for solving Bayesian Linear Regression and implementation of this approach using the `BayesianConjugateLinearRegressor` estimator.  
In addition, lightweight  `utils.py` and small synthetic dataset are included.

None

Correctness of exposition

No

#### Any other comments?
No

#### PR checklist
No

##### For all contributions

##### For new estimators


meraldoantonio:
Hi @fkiraly,

This PR includes the first two example notebooks for the Bayesian estimators. I suggest we merge these first, and I can submit the remaining notebooks, which are still in progress, in a separate PR.

I haven’t made any major changes to these notebooks since your review in the old PR (#500).

Let me know your thoughts or if I need to make any change to this PR, thanks!

examples/bayesian/1. Introduction to Bayesian Linear Regression.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   "metadata": {},
+   "source": [
+    "# Objective"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This series of notebooks offers an in-depth exploration of the **Bayesian Linear Regression**. \n",
```
fkiraly, line 14: "the" is superfluous

examples/bayesian/1. Introduction to Bayesian Linear Regression.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Linear regression is a widely used model due to its simplicity and interpretability. \n",
+    "\n",
+    "\n",
+    "In its simplest form, it predicts a single target $t$ as the deterministic output of the function $y$, which in turn is a linear combination of input variables $\\mathbf{x} = (x_1, \\dots, x_D)^\\top$ and parameters $\\mathbf{w} = (w_0, w_1, \\dots, w_D)^\\top$:\n",
```
fkiraly, line 88: I think introducing "t" is more confusing than helping - it is basically the same as y, from a rough semantic perspective, but a scalar rather than a function. I would not introduce it, instead I would all the function that you currently call "y" a function "f", and the nintroduce data "y" later, to whic hthe model is fit.

Further, I would use notation $f(x|w)$, the bar is common notation to separate inputs from parameters.

examples/bayesian/1. Introduction to Bayesian Linear Regression.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "$$\n",
+    "\n",
+    "where $\\epsilon \\sim \\mathcal{N}(0, \\beta^{-1})$ represents noise with zero mean and variance $\\beta^{-1}$ (i.e. in $\\beta$ is the *precision* of the Gaussian).\n",
+    "\n",
+    "\n",
+    "Alternatively, we can express  $t$ probabilistically by saying that it follows a Gaussian distribution, centered at $y(\\mathbf{x}, \\mathbf{w})$, with variance $\\beta^{-1}$:\n",
+    "\n",
+    "$$\n",
+    "\\begin{aligned}\n",
+    "p(t | \\mathbf{x}, \\mathbf{w}, \\beta) &= \\mathcal{N}(t | y(\\mathbf{x}, \\mathbf{w}), \\beta^{-1}) \\\\\n",
```
fkiraly, line 142: this notation for N is odd, I think incorrect?

What is it supposed to mean?

examples/bayesian/1. Introduction to Bayesian Linear Regression.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "\n",
+    "1. **Foundations of Linear Regression**  \n",
+    "   The first notebook (this one) lays the groundwork for understanding linear regression. It introduces the mathematical framework and provides an overview of frequentist approaches to estimate the model weights. This serves as a stepping stone for transitioning into the Bayesian perspective.\n",
+    "\n",
+    "\n",
+    "2. **Conjugate Prior in Bayesian Inference**  \n",
+    "   The second notebook delves into the concept of conjugate priors. By using conjugate distributions, we can derive posterior distributions analytically, simplifying Bayesian inference. This notebook also highlights how prior knowledge can influence the model and improve predictions in the presence of limited data.\n",
+    "\n",
+    "\n",
+    "3. **MCMC and Variational Inference**  \n",
```
fkiraly, line 27: there are also 4. various strategies known as "approximate Bayes" here.

fkiraly (changes requested):
Nice notebooks!

High-level, I think two points of improvement are most important:

* it is crucial that a reader gets some signposting about what they can expect from the notebooks, otherwise notebook 1 may disappoint expectations. A typical reader - seeing the `bayesian` folder - will want to see how they can use `skpro`. However, that is not what notebook 1 is about.
    * it is still interesting, but perhaps say at the start that this is more of an explainer intro with code.
* for code notebooks, the textbook style is discouraged, at least I would recommend to use more telegram and bullet point style. That makes the notebook less of a lengthy read.

On the mechanical side:

* in notebook 2, I would suggest to minimize extraneous content that is added only for it. For instance, can we not use one of the `sklearn` datasets, instead of adding csvs in the folder?
* we should also minimize "line-by-line" logic in the cells, 20 lines of data generation is not very useful from a didactic perspective. Can we outsource this somewhere in a data generation routine, in the main repo? Or just use a dummy dataset from `sklearn` and split it up?
* It is also important to split the vignettes more, e.g., have `skpro` vignettes separate from plotting cells.
* for plotting distributions - at least univariate ones - you can use the `skpro` onboard plotting. If you think it is not sufficient, we should add more features to that.
    * multivariate is not yet supported, but that might be also a good small project?
* Regarding the final summary, with advantages and disadvantages: conjucates are also simple, compute efficient algorithms typically. A user may want to use evaluation to see how well they fare, against more memory or compute heavy variants...

meraldoantonio:
> Nice notebooks!
> 
> High-level, I think two points of improvement are most important:
> 
> * it is crucial that a reader gets some signposting about what they can expect from the notebooks, otherwise notebook 1 may disappoint expectations. A typical reader - seeing the `bayesian` folder - will want to see how they can use `skpro`. However, that is not what notebook 1 is about.
>   
>   * it is still interesting, but perhaps say at the start that this is more of an explainer intro with code.
> * for code notebooks, the textbook style is discouraged, at least I would recommend to use more telegram and bullet point style. That makes the notebook less of a lengthy read.
> 
> On the mechanical side:
> 
> * in notebook 2, I would suggest to minimize extraneous content that is added only for it. For instance, can we not use one of the `sklearn` datasets, instead of adding csvs in the folder?
> * we should also minimize "line-by-line" logic in the cells, 20 lines of data generation is not very useful from a didactic perspective. Can we outsource this somewhere in a data generation routine, in the main repo? Or just use a dummy dataset from `sklearn` and split it up?
> * It is also important to split the vignettes more, e.g., have `skpro` vignettes separate from plotting cells.
> * for plotting distributions - at least univariate ones - you can use the `skpro` onboard plotting. If you think it is not sufficient, we should add more features to that.
>   
>   * multivariate is not yet supported, but that might be also a good small project?
> * Regarding the final summary, with advantages and disadvantages: conjucates are also simple, compute efficient algorithms typically. A user may want to use evaluation to see how well they fare, against more memory or compute heavy variants...

Hi Franz,

Thanks for your comments! I'm working on the edits now and will streamline the presentation and reduce extraneous code/explanation.

Regarding the notation and exposition for Linear Regression, I closely followed [Bishop’s book](https://www.microsoft.com/en-us/research/uploads/prod/2006/01/Bishop-Pattern-Recognition-and-Machine-Learning-2006.pdf), particularly pages 140-141. This includes the introduction of the target variable t, the deterministic function y, and the use of  the fancy "N" to denote the Normal pdf.

I understand that some of Bishop's notation might be somewhat unconventional, but since my explanation—both in this notebook and the subsequent ones—is based directly on his book, I felt it would be best to maintain consistency with his style (e.g., retaining t for the target variable instead of changing it to y). This would also make it easier for readers who want to reference the text.

Would love to hear your thoughts on this!

fkiraly:
> Would love to hear your thoughts on this!

To preface, this is all not going to be a blocker for me, as long as it is consistent within the notebooks.

Regarding:

* t - I think this is really badly chosen notation by Bishop, most statistics books and 101 courses (at least those that I have seen in 15 years of academic teaching) have x and y. The letter "t" is usually reserved for time and time only, across math, statistics, physics, this is pretty universal. Which may also cause confusion because `sktime` is the neighbour package. I have a strong preference to use anything else, though this is not blocking.
* regarding the notation "N", I do think there is conceptual conflation, at least on the notational level, going on in the Bishop book. There are at least three different, relevant concepts: the distribution, its pdf, and random variables distributed according to the distirbution. In "normal" (pun intended) statistical literature, it is usually the distribution, not the pdf. The pdf is p with some subscript that may vary.
    * it is much less "bad" than the "t" situation in the sense there is less danger for confusion in the case of the notebook. However, in `skpro` we have a clear conceptual separation between distributions and pdfs, so introducing notation that may blur this line may make understanding the concepts more difficult for users.

Either way, I do not consider these blocking, as you are clearly referencing the Bishop book.

Issue #30, opened 2023-08-25
[ENH] transformed distribution


fkiraly:
The package should contain a transformed distribution, after applying a fitted `sklearn` transformer's `transform` or `inverse_transform`.

As these can be arbitrary, it will probably be necessary to estimate most distribution methods by Monte Carlo.

This will be useful in a probabilistic version of `TransformedTargetRegressor`.

PR #555, opened 2025-06-26
[ENH] transformed distribution

fkiraly:
Fixes https://github.com/sktime/skpro/issues/30

Implements a transformed distribution, which takes a distribution and a transformation.

Useful for transformed regressors and forecasting in `sktime`.

For now, the transformation is done via simple sampling.

### DIFF
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -142,7 +142,7 @@ class TransformedDistribution(BaseDistribution):
             # if the transform returns a DataFrame, we ensure the index and columns
             outer_ppf.index = self.index
             outer_ppf.columns = self.columns
-        else:
+        elif not self._is_scalar_dist:
             # if the transform returns a scalar or array, we  convert it to DataFrame
             outer_ppf = pd.DataFrame(outer_ppf, index=self.index, columns=self.columns)
 


fkiraly:
Stuck at API conformance:

* consistent index after subsetting rows or cols - what should the index be?
* pdf is not defined even if it was defined before. Could use approx for derivative of cdf, if it is monotonous.

tingiskhan:
This would be a really nice feature, especially in the context of predict_proba for some of my use cases

fkiraly:
Any idea how we deal with the `ppf` method?

I think the index is easy to manage, by simply "inheriting" it.

fkiraly:
maybe just making the monotonic assumption default true?

tingiskhan:
> Any idea how we deal with the `ppf` method?
> 
> I think the index is easy to manage, by simply "inheriting" it.

Great question. I have mainly used numpyro for probabilistic computing, and this is how they do it

```python
    def icdf(self, q):
        sign = 1
        for transform in self.transforms:
            sign *= transform.sign
        q = jnp.where(sign < 0, 1 - q, q)
        value = self.base_dist.icdf(q)
        for transform in self.transforms:
            value = transform(value)
        return value
```

fkiraly:
it looks like:

* they assume monotonicity
* unlike `skpro` distributions, they comput eagerly?

tingiskhan:
> it looks like:
> 
>     * they assume monotonicity
> 
>     * unlike `skpro` distributions, they comput eagerly?

* Not sure, their implementation of `TransformedDistribution` is only allowed to be a subclass of `Transform`, see [here](https://num.pyro.ai/en/stable/distributions.html#transforms). They do implement the absolute transform, which is non-monotonic?
* Hm, I might be misunderstanding you, do you have an example of what you mean by eagerly? Because all transforms are performed when calling `sample/log_prob` I think? They do keep track of forward and inverse shapes in order to keep track of batch and event shapes of distributions.

fkiraly:
>  They do implement the absolute transform, which is non-monotonic?

Can you explain to me why/how the `icdf` implementation is then correct? I am genuinely confused, it seems to use the formula

$$Q_{f(X)}(p) = f(Q_X(p))$$, for random variables $X$ and ppf $Q$, which is only correct if $f$ is monotonously increasing.

> Hm, I might be misunderstanding you, do you have an example of what you mean by eagerly

I think you understood correctly, I just was mistaken about when the function body is executed.
(it is not actually eager execution, only on demand)

tingiskhan:
> > They do implement the absolute transform, which is non-monotonic?
> 
> Can you explain to me why/how the `icdf` implementation is then correct? I am genuinely confused, it seems to use the formula
> 
> Q f ( X ) ( p ) = f ( Q X ( p ) ) , for random variables X and ppf Q , which is only correct if f is monotonously increasing.
> 
> > Hm, I might be misunderstanding you, do you have an example of what you mean by eagerly
> 
> I think you understood correctly, I just was mistaken about when the function body is executed.

No, you're absolutely correct, it's very strange. 

But, taking a closer look at the implementation of  `AbsTransform` it does not implement the `log_abs_det_jacobian` method, making it unusable in `log_prob` calculations and therefore only usable when sampling - which is my _guess_.

fkiraly:
oh, it is just the absolute value function and not an arbitrary transformation?

tingiskhan:
> oh, it is just the absolute value function and not an arbitrary transformation?

Hm, it's a transformation that's usable for constructing a `TransformedDistribution`, and then sampling from it using the primitive `numpyro.sample`. But I don't think you would be able to use it for inferring either latent or observable variables, given that it misses the monotonicity constraint you pointed out, meaning that it doesn't implement the `log_abs_det_jacobian` required for calculating the transformed RV's (log) pdf. You would however be able to sample from the RV when sampling from the prior/posterior predictive distribution (I think) since this doesn't require evaluating the pdf.

But back to your question earlier (after this discussion): I think it makes sense to have a variable called `assume_monotonic`, requiring the user to think about the transforms they're passing. An alternative solution would be to (like numpyro) implement our own registry of transforms, but I think that's a bit daunting.

fkiraly:
>An alternative solution would be to (like numpyro) implement our own registry of transforms, but I think that's a bit daunting.

Ouch, I'd rather avoid that rabbit hole for now...

tingiskhan:
> Ouch, I'd rather avoid that rabbit hole for now...

And a deep one at that… I think I implemented my own suite of distributions for an early project (pyfilter) before switching to pytorch, and it was horrible

fkiraly:
You wrote a filter package? We went through a lot of work resurrecting `pykalman` since filtering packages were suddenly disappearing!

Knowing about this would have made a lot of things easier! Have you seen `pykalman` btw?

tingiskhan:
> You wrote a filter package? We went through a lot of work resurrecting `pykalman` since filtering packages were suddenly disappearing!
> 
> Knowing about this would have made a lot of things easier! Have you seen `pykalman` btw?

Haha, yes, focused mainly on joint parameter estimation and filtering using SMC algorithms. Also implemented vanilla Kalman and UKF. 

I had actually completely missed that you've revived `pykalman`, looks really refreshed. Perhaps you've even fixed the bug with masked arrays causing log pdf to fail? I was a pretty heavy user of `pykalman` earlier actually, but needed something that could handle non-linear/non-Gaussian models (hence my own package). I then came across `numpyro` which I use for most of my workflow. 

But I see you've really revived it, nice work - it's a great package! It's really impressive it has survived as long as it has.

fkiraly:
> I had actually completely missed that you've revived `pykalman`, looks really refreshed. Perhaps you've even fixed the bug with masked arrays causing log pdf to fail?

Did you report the bug? Could you point me to an issue?

tingiskhan:
> Did you report the bug? Could you point me to an issue?

This was a bug from way back when I still used it heavily (like 2017), and I think it was reported but the lib had pretty much been abandoned at that time, so it was addressed in some forks. It related to “pykalman/standard.py” where you had to add a line on row 167 (on my phone, can’t format properly)

observation = observation.filled()

in order to calculate the log likelihood without it crashing, something related to masked arrays if my memory serves me well. This might have been related to numpy, and I haven’t used the lib in a long time, so I don’t know if this is still an issue.

fkiraly:
Does not ring a bell - but if you still have the fix, you can open a PR.

Issue #554, opened 2025-06-23
[ENH] Zero-inflated distributions


fkiraly:
Zero-inflated distributions are important in intermittent demand forecasting, so it would be nice to have these as explicit distributions, and/or a compositor.

Based on discussion with @tingiskhan in https://github.com/sktime/sktime/pull/8438.

I can think of three high-priority ones, for discussion:

- [ ] zero-inflated Poisson
- [ ] zero-inflated Negative Binomial
- [ ] zero-inflated compositor, i.e., inflates any distribution with zeros, e.g., `ZeroInflated(any_distribution, zero_rate=0)`. This should have analytic formulae for all of the distribution defining methods.

A possible approach - though challenging - might be to write the zero-inflated compositor *first*, and then define zero-inflated Poisson and NB by delegation, using `_DelegatedDistribution`.

fkiraly:
Come to think of it, `ZeroInflated` is just a special case of `Mixture`, between `any_distribution` and the `Delta` distribution (supported at zero) - still might make sense to implement it separately.

We could use the `_DelegatedDistribution` to delegate to a `Mixture` with `Delta`, though it may be that there is an efficiency loss in doing so. Also, `Mixture` distributions in general have no explicit formula for inverse cdf, why mixtures with `Delta` do, hence I do think we still should implement `ZeroInflated` from scratch.

tingiskhan:
I think we're in luck regarding the Hurdle distribution: https://www.statsmodels.org/dev/examples/notebooks/generated/count_hurdle.html . 

Edit: Or perhaps not since statsmodels treats the distribution as a model rather than distribution.

fkiraly:
good first issue, the `Hurdle` distribution can be taken as a template for this

Khushmagrawal:
Hey @fkiraly, I would love to work on this issue, I have reviewed the existing `Hurdle` implementation.
For Zero-inflated distributions we have to allow zeros coming from base distributions as well, unlike `Hurdle`. I was initially thinking using `TruncatedDistribution` with a bound slightly below zero (e.g. -np.finfo(float).eps) to ensure 0 remains included but would it make more sense to extend `TruncatedDistribution` with a parameter like `inclusive=True` (defaulting to False to preserve current behavior) so that we can properly model lower bound

Khushmagrawal:
[...]

PR #557, opened 2025-06-28
[ENH] Adds `Hurdle` and `LeftTruncatedDiscrete` distributions

tingiskhan:
#### Reference Issues/PRs

Partially adresses #554 

Adds a left truncated and hurdle distribution.

No.

* Correctness of implementation.
* Design choice of using `BaseDistribution` for base class for both distributions. They should perhaps classify as something else, like transform?

No new tests.

#### Any other comments?

Still WIP, just opening a draft PR to discuss design choices before opening a full.

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -21,7 +21,6 @@ class LeftTruncatedDiscrete(BaseDistribution):
     """
 
     _tags = {
-        "object_type": "transform",
         "capabilities:approx": ["energy", "pmf", "cdf"],
         "capabilities:exact": ["ppf", "mean", "var", "log_pmf"],
         "distr:measuretype": "discrete",
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -1,0 +1,49 @@
+import numpy as np
+
+from skpro.distributions import Uniform
+from skpro.distributions.base import BaseDistribution
+
+
+# TODO: Implement a general discrete truncation and let this inherit from it
+class LeftTruncatedDiscrete(BaseDistribution):
+    """A left truncated discrete distribution _not_ including the lower bound.
+
+    This distribution samples from a given discrete distribution but excludes the
+    values below a specified lower bound.
+
+    Parameters
+    ----------
+    distribution : BaseDistribution
+        The base discrete distribution from which to sample.
+
+    lower_bound : int
+        The lower bound below which values are truncated (excluded from sampling).
+    """
+
+    def __init__(self, distribution: BaseDistribution, lower_bound: int):
+        super().__init__()
+        self.distribution = distribution
+        self.lower_bound = lower_bound
+
+    def _sample(self, n_samples: int):
+        u = Uniform(0.0, 1.0).sample(n_samples)
+        return self._ppf(u)
+
+    def _log_pdf(self, x):
+        is_invalid = x <= self.low
+
+        log_prob_base = self.distribution.log_prob(x)
+
+        log_prob_at_zero = self.base_dist.log_prob(self.low)
+        log_normalizer = np.log1p(-np.exp(log_prob_at_zero))
+
+        log_prob_truncated = log_prob_base - log_normalizer
+
+        return np.where(is_invalid, -np.inf, log_prob_truncated)
+
+    def _ppf(self, u):
+        low_cdf = self.distribution.cdf(self.lower_bound)
+        normalizer = 1.0 - low_cdf
+        x = low_cdf + normalizer * u
+
+        return self.distribution.ppf(x)
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
 
 from skpro.distributions import Uniform
 from skpro.distributions.base import BaseDistribution
@@ -70,10 +71,50 @@ class LeftTruncatedDiscrete(BaseDistribution):
     def get_test_params(cls, parameter_set="default"):
         from skpro.distributions import Poisson
 
+        # scalar
         poisson = Poisson(mu=1.0)
         params1 = {
             "distribution": poisson,
             "lower_bound": 0,
         }
 
-        return [params1]
+        # array
+        idx = pd.Index([1, 2])
+        cols = pd.Index(["a", "b"])
+        n_array = Poisson(mu=[[1, 2], [3, 4]], columns=cols, index=idx)
+        params2 = {
+            "distribution": n_array,
+            "lower_bound": 0,
+            "index": idx,
+            "columns": cols,
+        }
+
+        return [params1, params2]
+
+    def _iloc(self, rowidx=None, colidx=None):
+        distr = self.distribution.iloc[rowidx, colidx]
+
+        if rowidx is not None:
+            new_index = self.index[rowidx]
+        else:
+            new_index = self.index
+
+        if colidx is not None:
+            new_columns = self.columns[colidx]
+        else:
+            new_columns = self.columns
+
+        cls = type(self)
+        return cls(
+            distribution=distr,
+            lower_bound=self.lower_bound,
+            index=new_index,
+            columns=new_columns,
+        )
+
+    def _iat(self, rowidx=None, colidx=None):
+        if rowidx is None or colidx is None:
+            raise ValueError("iat method requires both row and column index")
+        self_subset = self.iloc[[rowidx], [colidx]]
+
+        return type(self)(distribution=self_subset.distribution.iat[0, 0], lower_bound=self.lower_bound)
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -36,10 +36,6 @@ class LeftTruncatedDiscrete(BaseDistribution):
 
         super().__init__(index=index, columns=columns)
 
-    def _sample(self, n_samples: int):
-        u = Uniform(0.0, 1.0, index=self.index, columns=self.columns).sample(n_samples)
-        return self._ppf(u)
-
     def _log_pmf(self, x):
         is_invalid = x <= self.lower_bound
 
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -21,6 +21,7 @@ class LeftTruncatedDiscrete(BaseDistribution):
     """
 
     _tags = {
+        "object_type": "transform",
         "capabilities:approx": ["energy", "pmf", "cdf"],
         "capabilities:exact": ["ppf", "mean", "var", "log_pmf"],
         "distr:measuretype": "discrete",
@@ -40,11 +41,11 @@ class LeftTruncatedDiscrete(BaseDistribution):
         return self._ppf(u)
 
     def _log_pmf(self, x):
-        is_invalid = x <= self.low
+        is_invalid = x <= self.lower_bound
 
-        log_prob_base = self.distribution.log_prob(x)
+        log_prob_base = self.distribution.log_pmf(x)
 
-        log_prob_at_zero = self.base_dist.log_prob(self.low)
+        log_prob_at_zero = self.distribution.log_pmf(self.lower_bound)
         log_normalizer = np.log1p(-np.exp(log_prob_at_zero))
 
         log_prob_truncated = log_prob_base - log_normalizer
--- a/skpro/distributions/discrete_truncated.py
+++ b/skpro/distributions/discrete_truncated.py
@@ -1,7 +1,8 @@
+"""Left Truncated Discrete Distribution."""
+
 import numpy as np
 import pandas as pd
 
-from skpro.distributions import Uniform
 from skpro.distributions.base import BaseDistribution
 
 
@@ -19,6 +20,7 @@ class LeftTruncatedDiscrete(BaseDistribution):
 
     lower_bound : int
         The lower bound below which values are truncated (excluded from sampling).
+
     """
 
     _tags = {
@@ -29,7 +31,9 @@ class LeftTruncatedDiscrete(BaseDistribution):
         "broadcast_init": "on",
     }
 
-    def __init__(self, distribution: BaseDistribution, lower_bound: int, index=None, columns=None):
+    def __init__(
+        self, distribution: BaseDistribution, lower_bound: int, index=None, columns=None
+    ):
         assert distribution._tags["distr:measuretype"] == "discrete", ""
 
         self.distribution = distribution
@@ -68,7 +72,7 @@ class LeftTruncatedDiscrete(BaseDistribution):
         return (self.distribution.var() + mean**2) / normalizer - mean**2
 
     @classmethod
-    def get_test_params(cls, parameter_set="default"):
+    def get_test_params(cls, parameter_set="default"):  # noqa: D102
         from skpro.distributions import Poisson
 
         # scalar
@@ -117,4 +121,7 @@ class LeftTruncatedDiscrete(BaseDistribution):
             raise ValueError("iat method requires both row and column index")
         self_subset = self.iloc[[rowidx], [colidx]]
 
-        return type(self)(distribution=self_subset.distribution.iat[0, 0], lower_bound=self.lower_bound)
+        return type(self)(
+            distribution=self_subset.distribution.iat[0, 0],
+            lower_bound=self.lower_bound,
+        )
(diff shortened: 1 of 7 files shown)


fkiraly (changes requested):
Nice!

Only minor documentation requests:

* can you add a little bit of math (not too overbording) to explain what each of the distributions does, in the docstring?
* could you add them to the API reference in `docs/source/api_reference`?

Not a request but a question for discussion, separate from this PR:

I also see you stumbled over `_iloc` and `_iat` - we have not had many composite distributions, but htis `new_index = ` case distinction appears multiple times and looks like it should perhaps be dealt with the boilerplate in `iloc` - what do you think? So the `rowidx` and `colidx` are guaranteed to always be an index - possibly the full index.

That way, the user would have to implement only double subsetting, and not three cases (one None, or no None)

skpro/distributions/discrete_truncated.py
```diff
@@ -0,0 +1,135 @@
...
+
+# TODO: Implement a general discrete truncation and let this inherit from it
+class LeftTruncatedDiscrete(BaseDistribution):
+    r"""A left truncated discrete distribution _not_ including the lower bound.
+
+    Assume that the base distribution is a discrete distribution with a support of
+    non-negative integers (e.g., Poisson, Binomial, etc.). This distribution samples
+    from the base distribution but excludes the values below a specified lower bound.
+    Mathematically, it can be expressed as:
+        .. math::
```
fkiraly: have you checked in the docs that this renders correctly? It is usually about the number of indents and newlines - I would have made an indent less and a newline more at the start, though I cannot predict how it renders (for this, the doc job should give a preview)

skpro/distributions/discrete_truncated.py
```diff
@@ -0,0 +1,135 @@
...
+
+    where :math:`\lambda` is the lower bound, and :math:`f(y)` is the probability mass
+    function.
+
+    Parameters
+    ----------
+    distribution : BaseDistribution
+        The base discrete distribution from which to sample.
+
+    lower_bound : int
```
fkiraly: does this have to be int? Could be float, no?

skpro/distributions/discrete_truncated.py
```diff
@@ -0,0 +1,135 @@
...
+        low_cdf = self.distribution.cdf(self.lower_bound)
+        normalizer = 1.0 - low_cdf
+        x = low_cdf + normalizer * p
+
+        return self.distribution.ppf(x)
+
+    def _mean(self):
+        low_cdf = self.distribution.cdf(self.lower_bound)
+        normalizer = 1.0 - low_cdf
+        return self.distribution.mean() / normalizer
```
fkiraly: I do not think this formula is correct? Take the distribution with equal mass on -3, 1. The mean before was -1, and if we lower truncate at 0, the mean becomes 1. According to your formula, it should be -2 but that is false.

skpro/distributions/discrete_truncated.py
tingiskhan: I tried previewing in my IDE but my Pycharm doesn't render doc strings for some reason, I'll try and see if I can get it working

skpro/distributions/discrete_truncated.py
tingiskhan: For this transform/distribution I've enforced only discrete distributions (line 43)

skpro/distributions/discrete_truncated.py
tingiskhan: This transform is only supported for non-negative discrete distributions, could take a crack at making it general in terms of negative as well?

tingiskhan: But it should work for any distribution now that I think of it. It's the mean that've messed up (as you pointed out)

skpro/distributions/discrete_truncated.py
tingiskhan: But yes, it could be a float but I found it to be a bit confusing since it's intended for discrete distributions

skpro/distributions/discrete_truncated.py
tingiskhan: Hm, I think it might make most sense to skip this calculation altogheter and use Monte Carlo for approximation since I'll have to perform a sum either way (I think?)

tingiskhan:
>     * could you add them to the API reference in `docs/source/api_reference`?

Having some issues with this it seems, I thought [this commit](https://github.com/sktime/skpro/pull/557/commits/f6d772ffea2f1a3608b55c7bcdbce49afd479139) should have done the trick?

Edit: seems to be working now and renders properly.

skpro/distributions/discrete_truncated.py
fkiraly: Hmm, I think there is no general formula in terms of the methods available. If you do not implement the `_mean` method, it is not computed Monte Carlo, but based on the approximation $\mathbb{E}[X] = \int Q_X(p) dp$, where $Q_X$ is the ppf (=icdf = quantile function), this is typically a more effective approximation.

Off-topic, we could introduce an argument to the mean in all distributions, meaning the $a$ in $\mathbb{E}[X|X>a]$? These usually have an analydic form for most parametric distributions. Although that would be a major undertaking.

skpro/distributions/discrete_truncated.py
```diff
@@ -0,0 +1,125 @@
+"""Left Truncated Discrete Distribution."""
+
+import numpy as np
+
+from skpro.distributions.base import BaseDistribution
+
+
+# TODO: Implement a general discrete truncation and let this inherit from it
+class LeftTruncatedDiscrete(BaseDistribution):
```
fkiraly: Silly thought maybe: if we are already doing this, should we not just introduce a `Truncated` distribution? Covering discrete and continuous, and `lower_bound` and `upper_bound`?

If either is left default = `None`, there is no such bound.

skpro/distributions/discrete_truncated.py
fkiraly: what restricts this to discrete distributions? The formulae seem to be fine in general.

skpro/distributions/discrete_truncated.py
fkiraly: you need to look at the CI doc build:

image

skpro/distributions/discrete_truncated.py
```diff
@@ -0,0 +1,125 @@
...
+        log_normalizer = np.log1p(-np.exp(log_prob_at_zero))
+
+        log_prob_truncated = log_prob_base - log_normalizer
+
+        return np.where(is_invalid, -np.inf, log_prob_truncated)
+
+    def _ppf(self, p):
+        low_cdf = self.distribution.cdf(self.lower_bound)
+        normalizer = 1.0 - low_cdf
+        x = low_cdf + normalizer * p
```
fkiraly, line 66: I do not think this formula is correct. If `p=1` the result is `1`, which is in general false.

skpro/distributions/discrete_truncated.py
tingiskhan: Yep, I had to check the “Show diff” in order for to to show the changes

skpro/distributions/discrete_truncated.py
tingiskhan: Sounds like a good plan, I’ll take a crack at it

fkiraly:
hm, can you explain to me why you created three classes for truncated? Is there sth I am missing about the formulae?

skpro/distributions/truncated.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def _ppf(self, p):
+        prob_at_lower = (
+            self.distribution.cdf(self.lower) if self.lower is not None else 0.0
+        )
+        prob_at_upper = (
+            self.distribution.cdf(self.upper) if self.upper is not None else 1.0
+        )
+
+        factor = prob_at_upper - prob_at_lower
```
fkiraly, line 106: should this not be `factor = self._get_normalizer()`, and similar in other locations?

tingiskhan:
> hm, can you explain to me why you created three classes for truncated? Is there sth I am missing about the formulae?

Tests fail if I just use one class (tried that first). So tried separating the discrete and continuous case, which worked. It seems you can’t define a class that implements methods for both pdf and pmf - got an error relating to the result not being a scalar.

Edit: a workaround (I suppose) would be to make _Truncated hidden (like now) and only access _Truncated via a factory. This would require us to write separate tests for this one (I suppose)? Also might make API a bit confusing since _Truncated implements both pmf/pdf and thus won't raise an error itself, but that would come downstream. 

What would you prefer @fkiraly ?

fkiraly:
> So tried separating the discrete and continuous case, which worked. It seems you can’t define a class that implements methods for both pdf and pmf - got an error relating to the result not being a scalar.

Hm, that should work if you set the distribution type - but if not, might be a bug.

I would really like to see the failing code - do you still have it? Perhaps you could open a draft Pr at that state so I can investigate?

fkiraly:
> What would you prefer @fkiraly ?

If I remember the original API design, distirbutions can implement `pmf` and `pdf` both - mixed distributions even have both. So if tihs is not properly working, we should investigate.

We could do in the Friday meet-up, or if you open a draft PR with the issue, I can investiate.

tingiskhan:
No, you were right (again), changing to "mixed" did the trick!

skpro/distributions/hurdle.py
```diff
@@ -0,0 +1,198 @@
+"""Hurdle distribution implementation."""
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+from skpro.distributions.base import BaseDistribution
+from skpro.distributions.left_truncated import LeftTruncated
+
+
+# TODO: how to handle index/columns in these transformed distributions? must they be
```
fkiraly, line 11: is this todo still a todo?

Index must not be the same as in the original distribution, but the index must be set in `__init__` already. If `None`, a sensible default should happen.

skpro/distributions/hurdle.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def __init__(
+        self,
+        p: ArrayLike,
+        distribution: LeftTruncated,
+        index=None,
+        columns=None,
+    ):
+        assert (
+            distribution.lower == 0
```
fkiraly: this is a contract not all distributions satisfy - if `distribution` does not have the `lower` attribute, this will fail.

What I would suggest instead: check if the distribution is `LeftTruncated`, if not, left truncate it at zero.

A more idiomatic approach would be to add supports that are inspectable to all distributions, though that is a major change and out of scope (I wonder if @vascosch92 is stlil working on this?)

fkiraly (changes requested):
Looks briliant! I just left a comment in the `Hurdle` distribution, since it assumes its distribution is `LeftTruncated`.

This should work for all distibutions, or it is not a proper composite.

I would either automatically left-truncate, or introduce a tag for distributions that classify it into non-negative or similar.

tingiskhan:
> Looks briliant! I just left a comment in the `Hurdle` distribution, since it assumes its distribution is `LeftTruncated`.
> 
> This should work for all distibutions, or it is not a proper composite.
> 
> I would either automatically left-truncate, or introduce a tag for distributions that classify it into non-negative or similar.

Makes sense, I'll address that today!

skpro/distributions/hurdle.py
tingiskhan: Sorry, it's still a todo. I'm guessing defaulting to the original distribution's index/columns make sense?

tingiskhan:
I’ll take a look at the failing tests this weekend!

fkiraly:
This is the error, and only on windows, `python 3.10` - if this is something related to the implementation, it would be quite strange.

```
FAILED skpro/distributions/tests/test_all_distrs.py::TestAllDistributions::test_ppf_and_cdf[TruncatedDistribution-1] - assert False
 +  where False = <function allclose at 0x000001D10E5937B0>(array([[1., 3.],\n       [3., 1.]]), array([[1., 3.],\n       [3., 2.]]))
 +    where <function allclose at 0x000001D10E5937B0> = np.allclose
 +    and   array([[1., 3.],\n       [3., 1.]]) =      a    b\n1  1.0  3.0\n2  3.0  1.0.values
 +    and   array([[1., 3.],\n       [3., 2.]]) =      a    b\n1  1.0  3.0\n2  3.0  2.0.values
```

Visually, it looks exactly the same. So I suspect some floating point issue.

fkiraly:
I would just increase `rtol` in `allclose`.

tingiskhan:
@fkiraly, it seems like all the tests pass now? Did you re-run the failing ones?

fkiraly:
Hm, still seems like the issue persists

fkiraly:
So, the test that fails is that computes the `ppf` of the `cdf`, which should be mutual inverse.

The result is an off-by-one error in one entry (note the bottom right entry, 1 vs 2),

```
x          =      a    b
1  3.0  1.0
2  5.0  1.0
x_approx   =      a    b
1  3.0  1.0
2  5.0  2.0
```

... but strangely, only in a single test run.

fkiraly:
ok, that is strange - the error is sporadic.

Why would it be sporadic?

FYI; it is coming from the `TruncatedDistribution`.

fkiraly:
I would say we merge and report as a bug?

tingiskhan:
> I would say we merge and report as a bug?

Sounds like a plan. 

It does sound very strange though... I could see why it would happen (and be sporadic) for Hurdle since we use an approximation for the CDF, but it's strange coming from Truncated I'd say. I think it would be easy to add the theoretical CDF. I'll throw that together real fast now, and let's see if the error persists

fkiraly:
ah, I see! Probably that was the issue, numerical inaccuracy combined with the tag saying that the cdf was exact (so the test was run)

fkiraly:
ok, that definitely made it worse, everything is failing.

```
x          =      a    b    c
0  1.0  5.0  1.0
1  4.0  6.0  0.0
x_approx   =      a    b    c
0  0.0  0.0  0.0
1  0.0  0.0  0.0
```

fkiraly:
you changed the `Hurdle` distribution, the problem was the `TruncatedDistribution` - `Hurdle` seemed fine

tingiskhan:
> you changed the `Hurdle` distribution, the problem was the `TruncatedDistribution` - `Hurdle` seemed fine

Ah, my bad. Adding the exact CDF could be a good idea anyway I guess. But  it seems like something's wrong with Truncated as you say since the tests now fail for me locally (after the bug fix).

fkiraly:
> But it seems like something's wrong with Truncated as you say since the tests now fail for me locally (after the bug fix).

Can you try to run the test 10 times locally? I suspect it is sporadic or stochastic.

tingiskhan:
> ok, that definitely made it worse, everything is failing.
> 
> ```
> x          =      a    b    c
> 0  1.0  5.0  1.0
> 1  4.0  6.0  0.0
> x_approx   =      a    b    c
> 0  0.0  0.0  0.0
> 1  0.0  0.0  0.0
> ```

Ah, wait a sec... I think the test cases I use might be bad since `Poisson` doesn't implement an exact `ppf`?

I could skip checking for Poisson, or replace it with another distribution?

tingiskhan:
Perhaps composite distributions need to fetch the tags from the base distributions?

fkiraly:
> Ah, wait a sec... I think the test cases I use might be bad since `Poisson` doesn't implement an exact `ppf`?

Hmmm, shouldn´t Poisson have an exact `ppf`? I see it does not.

> I could skip checking for Poisson, or replace it with another distribution?

That might work

> Perhaps composite distributions need to fetch the tags from the base distributions?

Could you kindly explain what you mean by this?

fkiraly:
> Perhaps composite distributions need to fetch the tags from the base distributions?

Oh, I get it! It should set the tags of which methods are exact, based on the distribution inside! Yes! That is probably the programmatic solution, and we should include at least one example where everything is exact numerically.

tingiskhan:
> > Perhaps composite distributions need to fetch the tags from the base distributions?
> 
> Oh, I get it! It should set the tags of which methods are exact, based on the distribution inside! Yes! That is probably the programmatic solution, and we should include at least one example where everything is exact numerically.

Sounds good! Do you want me to implement a `CompositeDistribution` class in this PR? And, I think the `get_tag(s)` need to consider a "hierarchy" of the tags for the methods:
1. If composite is exact and base as well, then the result will be `exact`
2. If composite is exact and base is not (or vice versa) then the result will be `approximate`
3. If both are `approximate` then result will be approximate

Makes sense?

fkiraly:
> Makes sense?

This is an excellent idea! I would suggest: let´s do this in a separate PR, I would add it as a method `_default_composite_tags: self -> dict` to `BaseDistribution`, and I need to think about how best to call this. Automatically or manually, for instance.

fkiraly:
(I will merge this for now)

PR #568, opened 2025-07-28
[CI] Addresses stochasticity of Hurdle tests

tingiskhan:
#### Reference Issues/PRs

None referenced.

Addresses the stochasticity of the tests associated with Hurdle distribution by switching out the Poisson test with Negative Binomial instead, since Poisson doesn't implement an exact PPF.

No.

That tests pass deterministically.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/hurdle.py
+++ b/skpro/distributions/hurdle.py
@@ -167,12 +167,12 @@ class Hurdle(BaseDistribution):
     def get_test_params(cls, parameter_set="default"):  # noqa: D102
         import pandas as pd
 
-        from skpro.distributions import Poisson
+        from skpro.distributions import NegativeBinomial
 
         # scalar
         params_1 = {
             "p": 0.3,
-            "distribution": Poisson(mu=1.0),
+            "distribution": NegativeBinomial(mu=1.0, alpha=1.0),
         }
 
         # array 1
@@ -180,10 +180,10 @@ class Hurdle(BaseDistribution):
         idx = pd.Index([0, 1])
         cols = pd.Index(["a", "b", "c"])
 
-        poisson = Poisson(mu=mu, columns=cols, index=idx)
+        negbin = NegativeBinomial(mu=mu, alpha=1.0, columns=cols, index=idx)
         params_2 = {
             "p": 0.3,
-            "distribution": poisson,
+            "distribution": negbin,
             "index": idx,
             "columns": cols,
         }
@@ -191,7 +191,7 @@ class Hurdle(BaseDistribution):
         # array 2
         params_3 = {
             "p": np.array([0.2, 0.3]).reshape(-1, 1),
-            "distribution": poisson,
+            "distribution": negbin,
             "index": idx,
             "columns": cols,
         }
--- a/skpro/tests/_config.py
+++ b/skpro/tests/_config.py
@@ -10,5 +10,4 @@ EXCLUDE_ESTIMATORS = [
 
 EXCLUDED_TESTS = {
     "GLMRegressor": ["test_online_update"],  # see 497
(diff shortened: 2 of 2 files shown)


fkiraly:
oh, poisson does not have exact `ppf`?

Why does the `Hurdle` think i thas an exact `ppf`?

tingiskhan:
> oh, poisson does not have exact `ppf`?
> 
> Why does the `Hurdle` think i thas an exact `ppf`?

No it doesn't seem like it as it uses the general inverse formula for discrete RVs. But it seems as though that's not the main issue as it fails for Negative Binomial (which has an exact ppf).

Technically the Hurdle distribution has an exact inverse (if I'm not entirely wrong), which is given by the ppf of the truncated distribution (which is also exact) - the trouble is that the underlying distributions don't always have exact ppfs.

I'll continue debugging the Hurdle and see if I can resolve the issues as they seem to be related to the discrete cases only (as Normal case passes).

fkiraly:
This PR should set the tags of the composite distribution correctly based on those of the component distribution:
https://github.com/sktime/skpro/pull/569

tingiskhan:
Alright, it seems as though it's some sort of rounding issue tbh... If I debug and set a breakpoint at row 154 in `hurdle.py` and then offset `q_rescaled` with a small float (like 1e-6) I get the correct inverse.

Also seems to happen only for certain parameterizations (mu = 5)

fkiraly:
would it be a solution to shift `ppf` by eps?

tingiskhan:
> would it be a solution to shift `ppf` by eps?

Yep! Did that as per [fdf8063](https://github.com/sktime/skpro/pull/568/commits/fdf806322ccc92f967e3c72da2d014ad69a00348)

tingiskhan:
> This PR should set the tags of the composite distribution correctly based on those of the component distribution: #569

Should we combine this and #569?

fkiraly:
we can simply merge both? I do not think there is any conditional relation between the two PR, since it turns out #569 was not a fix to the bug.

fkiraly:
Do you have an idea why the tests in #569 are failing? It seems some API tests are running now but the return type is wrong.

tingiskhan:
> Do you have an idea why the tests in #569 are failing? It seems some API tests are running now but the return type is wrong.

I think I had the same issue when developing the Hurdle distribution, and I think it was related to the type of the distribution, or whether the distribution implemented some of the methods exactly/approximately

fkiraly:
I think we should try to figure out what is going on?

The risk is that if the tags are set to "approximate", some tests could be entirely skipped, and we do not notice actual bugs.

tingiskhan:
> I think we should try to figure out what is going on?
> 
> The risk is that if the tags are set to "approximate", some tests could be entirely skipped, and we do not notice actual bugs.

My take would be the following: skip the tag setting for now, and only use distributions with exact implementations for test cases in composite distributions? If the exact implementations work as expected, I guess that would imply that the approximate will work as well? Requires careful test construction for composite distributions, but my guess is that we need to implement either a mixin/base class for composite distributions as the use for this logic might pop up again later.

fkiraly:
My interpretation of the issue is that there are some bugs in `Hurdle` which are masked by the tags being approximate, but not causally related.

That is, there are actual bugs in the contract of `Hurdle`, which #569 "uncovers" but does not cause. "holding off" seems like a dangerous idea, since the bugs would be on `main` in this case?

tingiskhan:
> My interpretation of the issue is that there are some bugs in `Hurdle` which are masked by the tags being approximate, but not causally related.
> 
> That is, there are actual bugs in the contract of `Hurdle`, which #569 "uncovers" but does not cause. "holding off" seems like a dangerous idea, since the bugs would be on `main` in this case?

See comments #569

PR #569, opened 2025-07-28
[ENH] correctly set tags of `Hurdle` and `Truncated` depending on inner distribution

### LINKED ISSUE #566: [BUG] `Hurdle` distribution sporadically fails `test_ppf_and_cdf` with "off-by-one" errors
The `Hurdle` distribution fails `test_ppf_and_cdf` with "off-by-one" errors.

This is sporadic, indicating an issue with inverses under approximation, or an error with "greater equal" vs "greater strictly".

fkiraly:
This PR correctly sets the tag of `Hurdle` and `Truncated` depending on the inner distribution.

Hopefully this fixes #566 as a secondary effect.

### DIFF
--- a/skpro/distributions/hurdle.py
+++ b/skpro/distributions/hurdle.py
@@ -41,7 +41,6 @@ class Hurdle(BaseDistribution):
     >>> base = Normal(mu=1.0, sigma=1.0)
     >>> hurdle = Hurdle(0.5, base)
     >>> samples = hurdle.sample(1000)
-
     """
 
     _tags = {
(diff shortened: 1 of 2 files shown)


tingiskhan:
Hmm, this seems really strange:

image

Why is it testing the PMF for a truncated Normal? I think this was the issue I had earlier as well, when I used the wrong "measure" tag.

tingiskhan:
I think that this might be the issue:

image

I think we need to remove the method names that aren't relevant when changing measuretype.

tingiskhan:
@fkiraly , see #571

fkiraly:
see #571, continuous distributions should return 0 for pmf - this should actually be the default?

fkiraly:
The "one-off" error from #566 is still present. Any ideas?

tingiskhan:
Hm, very strange - I see that it tests Poisson for Truncated, I thought I changed all Poisson references to Negative Binomial. What happens if you replace it with NegativeBinomial instead?

fkiraly:
that would not fix the error, just "make the tests pass"?

tingiskhan:
I'll debug and see what's wrong!

tingiskhan:
Can't seem to reproduce the error:

```python
import tqdm

from skpro.distributions import Normal, TruncatedDistribution, Hurdle
import numpy as np


def ppf_and_cdf(d):
    capabilities_exact = d.get_tags()["capabilities:exact"]

    if "ppf" not in capabilities_exact or "cdf" not in capabilities_exact:
        return
    x = d.sample()
    x_approx = d.ppf(d.cdf(x))
    if d.ndim > 0:
        assert np.allclose(x.values, x_approx.values)
    else:
        assert np.allclose(x, x_approx)


cases = TruncatedDistribution.get_test_params()

for i in tqdm.tqdm(range(1_000)):
    for case in cases:
        dist = TruncatedDistribution(**case)
        ppf_and_cdf(dist)
```

Tried bumping to 5,000, still no issue. I'm on scipy==1.16.0 and Python=3.12. Will try to update scipy if see if that's the issue.

Edit: No issue

tingiskhan:
Perhaps we should move the eps fix to truncated instead?

tingiskhan:
@fkiraly, what happens if you merge the changes of #576?

fkiraly:
ok, I will try to merge #576 into this and then we see

fkiraly:
I think there is a genuine bug in #576, unrelated to the fix. What I will do, to make debugging easier: I will merge this without #576. and then you can run `check_estimator` locally to fix the distribution - does this make sense?

tingiskhan:
Sounds good. I had messed up which variable to look at when selecting the eps since I assumed that shifted_p would be an array, hence the failing tests

PR #570, opened 2025-07-28
[BUG] correctly set tags for `Poisson` distribution

fkiraly:
The "approximate vs exact" tag for the `Poisson` distribution were not set correctly - `Poisson` has an exact `ppf` interfaced from `scipy`.

Also fixes a problem with the 2nd test example which had a (non-permissible) mean zero for the Poisson.

### DIFF
--- a/skpro/distributions/poisson.py
+++ b/skpro/distributions/poisson.py
@@ -29,8 +29,8 @@ class Poisson(_ScipyAdapter):
     """
 
     _tags = {
-        "capabilities:approx": ["ppf", "energy"],
-        "capabilities:exact": ["mean", "var", "pmf", "log_pmf", "cdf"],
+        "capabilities:approx": ["energy", "pdfnorm"],
+        "capabilities:exact": ["mean", "var", "pmf", "log_pmf", "cdf", "ppf"],
         "distr:measuretype": "discrete",
         "distr:paramtype": "parametric",
         "broadcast_init": "on",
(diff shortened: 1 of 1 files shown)


fkiraly:
the test checks ppf(cdf(x)) = x, which is only true for continuous distributions for general x.

But, x is obtained from sample, so we know it is in the support of the dist, and on that support, the equality holds with probability one.

tingiskhan:
> oh, the test checks ppf(cdf(x)) = x, which is only true for continuous distributions.
> 
> For mixed ones, it should check whether x has positive mass (pmf), if so then only an inequality holds.

Aaaaah, alright! And I must have misunderstood the Poisson part then, since it seems like it was the other way around

fkiraly:
no, sorry, I was wrong with the point you cited (I amended my post while you were probably writing the response).

I am increasingly getting confused, see above. I think ppf(cdf(x)) holds with probability 1 if x is drawn from the distibution, so the test should never fail if the methods have the following properties:

* `cdf` is exact
* `ppf` is exact
* `sample` is measurable w.r.t the distribution (this suffices, and it is sufficient that `sample` is correct)

fkiraly:
... is this a bug in `scipy`?

fkiraly:
not a bug in scipy, neither is it one in skpro core logic, it was simply an inadmissible test parameter setting (Poisson with mean zero)

PR #571, opened 2025-07-31
[ENH] removes method names in tests for discrete and continuous distributions

tingiskhan:
Minor fixes

### DIFF
--- a/skpro/distributions/truncated.py
+++ b/skpro/distributions/truncated.py
@@ -74,6 +74,18 @@ class TruncatedDistribution(BaseDistribution):
             columns=columns if columns is not None else distribution.columns,
         )
 
+        # the measure type of this distribution is discrete if the inner distribution
+        # is discrete, and it is continuous if the inner distribution is continuous;
+        # if the measure type of the inner distribution is mixed,
+        # the result is indeterminate, but we keep it as mixed for API reasons
+        inner_measuretype = distribution.get_tag("distr:measuretype", "mixed")
+        self.set_tags(**{"distr:measuretype": inner_measuretype})
+
+        methods_to_remove = {
+            "discrete": ["pdf", "log_pdf"],
+            "continuous": ["pmf", "log_pmf"],
+        }
+
         # the capabilities of this (self) distribution are exact
         # if and only if the capabilities of the inner distribution are exact
         self_exact_capas = self.get_tag("capabilities:exact", []).copy()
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
The presence of `pmf` for continuous and `pdf` for discrete distributions is actually intended - it is a feature and not a bug!

Why: we are using the "strategy" pattern, and it is better if all methods are available for all distributions.

This also makes mathematical sense, in the case of discrete distributions `pdf` will always return 0, and in the case of continuous distributions, `pmf` will always return 0.
(by the Lebesgue decompositions theorem for instance - in it, continuous distributions have the zero measure as point mass part, and discrete distributions have the zero measure as continuous part)

From a software perspective, this reduces the amount of boilerplate for these case distinctions.
For example, if you look at mixture distributions, the resulting distribution always has `pdf` and `pmf` being a weighted average. This works well if the underlying distributions are allowed to return zero, instead of raising an exception.

tingiskhan:
Makes sense from a mathematical perspective, but I found that implementing these changes solved the test errors.

Are there any other distributions that implement both pmf and pdf and have  either discrete or continuous measure? I find it strange that these tests would fail for only Hurdle/Truncated if this should be supported already?

And if this is the pattern, I'm closing this PR?

And, just to clarify: I only remove the method names from the tags, not the class. 

Edit: I’m starting to think that we’re talking past each other, hehe. I totally agree on the part regarding the presence of a pmf for continuous distributions (and vice versa), but I can’t find a case for continuous/discrete distributions having pmf/pdf in either exact/approx tag, right?

fkiraly:
> Makes sense from a mathematical perspective, but I found that implementing these changes solved the test errors.

Interesting. Then this may indicate that the defaults are not properly working.

I am conjecturing that this might happen in the scalar case!

> And, just to clarify: I only remove the method names from the tags, not the class.

Removing the tags I think just turns off the tests, it does not solve the matter at the core.

> Edit: I’m starting to think that we’re talking past each other, hehe. I totally agree on the part regarding the presence of a pmf for continuous distributions (and vice versa), but I can’t find a case for continuous/discrete distributions having pmf/pdf in either exact/approx tag, right?

I will need some time to digest this, sorry. There is probably some API issue around pmf/pdf which has nothing to do with `Hurdle` etc, see above.

That this exists is likely a consequence of so many distributions being either approximate or "pure" continuous or discrete.

fkiraly:
Found the bug - it is #573, fix is in #574.

This PR does not fix the bug but just masks the tests by preventing them from running.

tingiskhan:
Nice!! I'll close this PR then!

tingiskhan:
Closed as per #574

PR #583, opened 2025-08-17
[MNT] Check versions in  wheels workflow

szepeviktor:
From https://github.com/sktime/skpro/commit/9d70805ad5b99673c722383d799f24b61d8e959e#commitcomment-164122353


fkiraly:
I have also added a new tag with the correct name. Since this did not impact the release, no new release is necessary.

szepeviktor:
> I would also rename `VERSION` to `GH_TAG_NAME` or similar, for more clarity

Done 🍏

Issue #601, opened 2025-09-25
[ENH] TransformTargetRegressor doesn't allow LogLoss


joshdunnlime:
**Is your feature request related to a problem? Please describe.**

LogLoss is a very commonly used proper scoring rule (e.g. logistic regression and NNs) and has efficient implementations for most distributions. The TransformTargetRegressor (TTR) as implementation does not include this. The TTR is lacking a `log_pdf` method, likely due to the numerical complications of mapping the `log_pdf` across domains. E.g. the transform function must be "[a one-to-one, monotonic function with a differentiable inverse](https://mc-stan.org/docs/reference-manual/transforms.html#univariate-changes-of-variables)".

In reality, many of the functions used to transform the target have trivial differentiable inverses and therefore, trivial solutions to mapping the log_pdf. The MinMaxScaler internals for the proposed in [SupportScaler](https://github.com/sktime/skpro/issues/588) is one such - it is a linear transformation so the derivative of the inverse is a constant.

Here is an walkthrough:

```python
import numpy as np
import pandas as pd
from sklearn.preprocessing import MinMaxScaler

from skpro.metrics import LogLoss, CRPS
from skpro.regression.xgboostlss import XGBoostLSS
from skpro.regression.compose import TransformedTargetRegressor

size = 1000

X = np.random.normal(1, 1, size)
y = 2 * X + np.random.normal(0, 0.5, size)

Xy = pd.DataFrame(
    {"target": y, "feature": X},
    index=range(size)
)

# use lower n_trails for speed-up
params = {"n_trials": 2, "dist": "Normal"}

# simple example with no scaling - logloss works
xgb = XGBoostLSS(**params)
xgb.fit(X=Xy[["feature"]], y=Xy["target"])
p = xgb.predict(Xy[["feature"]])
pp = xgb.predict_proba(Xy[["feature"]])

print(CRPS()(y_true=Xy["target"], y_pred=pp))
print(LogLoss()(y_true=Xy["target"], y_pred=pp))
```
> 0.25535
> 0.56026

Now use the TTR:

```python
xgb = XGBoostLSS(**params)
mms = MinMaxScaler()
pipe = TransformedTargetRegressor(regressor=xgb, transformer=mms)

# pipeline with TTR logloss fails
pipe.fit(X=Xy[["feature"]], y=Xy["target"])
pp = pipe.predict_proba(Xy[["feature"]])
p = pipe.predict(Xy[["feature"]])

[...]

PR #605, opened 2025-09-26
[ENH] Add LogLoss to TransformedTargetRegressor (TTR)

joshdunnlime:
#### Reference Issues/PRs
#601
Problem Summary: LogLoss not included due to lack of a `log_pdf`. It can be computed but requires calculating the Jacobian of the transform. For many transformers this is non-trivial.

I have implemented several pieces here. They breakdown as follows:
1) Add `log_pdf` method to TTR.
2) Add `_jacobian` method to TransformedDistribution (TD).
3) Replace `transform` method/function with `transformer` class in the TD.
4) Add `ordered_gradient` function to handle `gradient(f, x)` where the arrays are not sorted by x.

No.

1) Correctness of numerical calculations in `log_pdf` in TTR.
2) Edge-case of the Jacobian on TD.
3) Does replacing transform with transformer have any unintended consequences or cause a considerable breaking change?

I haven't yet added tests. I have modified the `get_test_params` to take a transformer instead of a single function.

##### For all contributions

##### For new estimators


skpro/distributions/trafo/_transformed.py
```diff
@@ -62,14 +64,16 @@ class TransformedDistribution(BaseDistribution):
     def __init__(
         self,
         distribution,
-        transform,
+        transformer,
```
fkiraly: sorry, we cannot change the names of already released arguments, or it might break user code. Not without deprecation and warnings ate least.

skpro/distributions/trafo/_transformed.py
```diff
@@ -133,7 +199,7 @@ def _ppf(self, p):
                 "set `assume_monotonic=True` to use this method"
             )
 
-        trafo = self.transform
+        trafo = self.transformer.inverse_transform
```
fkiraly: why are we changing this?

fkiraly (changes requested):
Great idea! Though we cannot merge this as is, since:

* this changes the arguments of existing released classes, possibly breaking user code
* I think it introduces too much coupling between TTR and the `TransformedDistribution`, we should avoid that.

Suggestion: how about we allow the user to pass both `transform` and `inverse_transform` (optional) to the `TransformedDistribution`? Then we can do numerical differentiation.

skpro/distributions/trafo/_transformed.py
joshdunnlime: The `self.transform` is just the inverse function at present, not the whole tranformer class. The idea was to expose the whole class so we can do numerical differentiation or access the `scale_` attr. Based on the comment above this isn't favourable.

skpro/distributions/trafo/_transformed.py
joshdunnlime: Understood. I'll make some changes.

skpro/distributions/trafo/_transformed.py
fkiraly: I think not easily possible, because the `__init__` resets every estimator that is a component to a pre-fitted state on construction, and because of deprecation.

Overall, I think it is a very good idea, though one may have to find the right implementation.

Would passing `inverse_transform` as an optional additional argument work?

Alternatively, we have to pass something like `sklearn` `FrozenEstimator` wrapping the transformer, though it feels like a bit of an overhead software engineering wise.
The general question is to whether, and if yes how, to allow, API-wise, fitted estimators to be passed to `__init__` - currently this violates the `scikit-learn`-like design.

fkiraly (changes requested):
may I request:

* remove the `BaseTransformer` from this pull request and open a separate one, since we will have some API discussions; let's aim to complete this with numerical differentiation, and later extend it with exact jacobians in a separate PR
* may I suggest to move the numerical differentiation to TD for it to happen only inside `_pdf` and `_log_pdf`? And leave TTR untouched? This we could merge quickly, and it could be "version 0.1" of the feature you want.e
    * testing it might also highlight aspects we have not yet thought about

joshdunnlime:
Apologies, I made the DiffTransformer changes on the wrong branch.

fkiraly:
No problem.

May I suggest to change this PR to revert changes across the classes, and add only an approximate `log_pdf` and `pdf` to the TD?

That would be a good starting point for adding more features in separate PR (e.g., autodiff-like).

skpro/distributions/trafo/_transformed.py
```diff
@@ -86,6 +88,72 @@ def __init__(
 
         super().__init__(index=index, columns=columns)
 
+    def pdf(self, x: pd.DataFrame):
```
fkiraly: this should be `_pdf` and `_log_pdf`, like the other methods. See the extension template.

fkiraly (changes requested):
Thanks - almost ready to merge. The public methods should not be overridden, private methods should be implemented.

See how, for instance, `_cdf` is implemented.

skpro/distributions/trafo/_transformed.py
```diff
@@ -86,6 +88,72 @@ def __init__(
...
+            points where the pdf is evaluated
+
+        Returns
+        -------
+        pd.DataFrame, same shape as ``self``
+            pdf values at the given points
+        """
+        dist = self.distribution
+
+        if isinstance(self.transformer_, "BaseDifferentiableTransformer"):
```
fkiraly: this attribute will not exist given current `__init__`

skpro/distributions/trafo/_transformed.py
```diff
@@ -86,6 +88,72 @@ def __init__(
...
+            )
+
+        if hasattr(dist, "_distribution_attr"):
+            obj = getattr(dist, dist._distribution_attr)
+            args, kwds = dist._get_scipy_param()
+            pdf = partial(obj.pdf, *args, **kwds)
+        else:
+            raise NotImplementedError
+
+        jac = np.abs(self.transformer_.inverse_transform_diff(x))
```
fkiraly: this we should replace with an approximate jacobian

skpro/distributions/trafo/_transformed.py
joshdunnlime: Reverted to `.transform`

skpro/distributions/trafo/_transformed.py
joshdunnlime: Fixed

joshdunnlime:
I have implemented the `_pdf` and `_log_pdf` so that they give approximations (e.g. not including the Jacobian). We could try to approximate the Jacobian but I think it is better to compute it directly (as per our discussions in the this PR and the other PR).

This PR could be merged but it does have to potential to return very approximate pdf and log_pdf results. My suggestion would instead be to make the DifferentiableTransformer or DifferentiableFunction changes on a branch from this (branch stacking) and merge it all at once.

Notes for reviewer/@fkiraly :

I have added some warnings to inform the user that these are approximations. I have added this to the docs as well. This can be removed when we implement the Jacobian.

When changing `pdf` and `log_pdf` to the private functions I ran into some issues. The first was that the internal (non-transformed) distribution in TTR didn't have the same index and columns as TTR. This caused `inf` to be returned. Setting the index and columns during TTR training fixed this. This explains the changes to [ttr](https://github.com/sktime/skpro/pull/605/files#diff-250f0e6f5c35534e02c71782ae50bf09cb2f997c031ed4c729dcb41653638342R108).

The seconds issue that was rather subtle was `"distr:measuretype": "discrete"` was causing the `_pdf` and `_log_pdf` calls to be ignored. I have changed this to [`"mixed"`](https://github.com/sktime/skpro/pull/605/files#diff-c89d9489064d49d9caf2fb49333090330830b224c416d6cd8b0bd3eb2285cf71R62). I assume this is correct (given the three choices). I'm not sure if the API allows this but perhaps setting the TTR `"distr:measuretype"` based on the distribution it is wrapping would be beneficial.

Finally, to pass the "pdf_log_pdf" tests, these had to be added to the [approx tags](https://github.com/sktime/skpro/pull/605/files#diff-c89d9489064d49d9caf2fb49333090330830b224c416d6cd8b0bd3eb2285cf71R58).

fkiraly:
> I have added some warnings to inform the user that these are approximations. 

Yes, good thinking - this is consistent with the default approximations in `BaseDistribution`, if any are called the user also receives warnings about those.

> I assume this is correct (given the three choices)

Yes, I think that is the correct choice to hard code, it is the most general class. We can dynamically change this later if we want to.

> Finally, to pass the "pdf_log_pdf" tests, these had to be added to the approx tags

Well, looks like the tests were doing their job.

fkiraly (changes requested):
Some questions which may imply changed requests:

* why are we doing the branch "`if hasattr(dist, "_distribution_attr")`? This feels like a violation of the law of demeter. We should only use `dist.pdf`.
* the wrong indices are explained by this, I think. You should not need to correct them if all we do is call `pdf.dist`.

Can you also explain to me how the computation is an approximation to transformed pdf? It is missing the jacobian.

joshdunnlime:
> Can you also explain to me how the computation is an approximation to transformed pdf? It is missing the jacobian.

I guess that depends on the definition of approximate - I might have been somewhat loose with it. Refer to my sentence from above:

> This PR could be merged but it does have the potential to return very approximate pdf and log_pdf results. My suggestion would instead be to make the DifferentiableTransformer or DifferentiableFunction changes on a branch from this (branch stacking) and merge it all at once.

I recommend not merging this independently but waiting until the we have a solution for the exact Jacobian in place. My suggestion is to use branch stacking to implement the DiffT or DiffFunc classes. I have this done already and will open a new PR (in place of https://github.com/sktime/skpro/pull/608).

fkiraly:
> > I assume this is correct (given the three choices)

> Yes, I think that is the correct choice to hard code, it is the most general class. We can dynamically change this later if we want to.

Actually, if it was `"discrete"`, the code in `__init__` should set it to `"discrete"` via `set_tags`. Otherwise no setting needs to be done, it should be `"mixed"`.

fkiraly:
> I guess that depends on the definition of approximate - I might have been somewhat loose with it. 

I was under the impression you were going to implement numerical derivatives.

I have now made a draft here in the base class: https://github.com/sktime/skpro/pull/610

This is due to realizing that this is nothing specific to the transformed distribution - any distribution that does not have a `pdf` implemented but a `cdf` can produce a numerical `pdf` like this.

This should, in particular, fix it for the `TransformedDistribution` - do you want to give it a test?

joshdunnlime:
> * why are we doing the branch "`if hasattr(dist, "_distribution_attr")`? This feels like a violation of the law of demeter. We should only use `dist.pdf`.
> * the wrong indices are explained by this, I think. You should not need to correct them if all we do is call `pdf.dist`.

Perhaps a misunderstanding on my behalf. This is now fixed.

Would we like to rollback the changes to ttr or are these beneficial? E.g. is a user wants to dig into the internals of TTR/TD the indices are shared between TD and the wrapped distribution.

joshdunnlime:
> > > I assume this is correct (given the three choices)
> 
> > Yes, I think that is the correct choice to hard code, it is the most general class. We can dynamically change this later if we want to.
> 
> Actually, if it was `"discrete"`, the code in `__init__` should set it to `"discrete"` via `set_tags`. Otherwise no setting needs to be done, it should be `"mixed"`.

It was `discrete` so I changed it explicitly to `mixed`. It seems to inherit `mixed` from the base class. Should I remove it?

fkiraly:
> It was `discrete` so I changed it explicitly to `mixed`. It seems to inherit `mixed` from the base class. Should I remove it?

I would say:

* leave for class at `"mixed"`
* in `__init__`, set to `"discrete"` if and only the inner distribution is `"discrete"`.

fkiraly:
> Would we like to rollback the changes to ttr or are these beneficial? E.g. is a user wants to dig into the internals of TTR/TD the indices are shared between TD and the wrapped distribution.

What are the indices currently? I think they should be the same between row and column index.

joshdunnlime:
> > I guess that depends on the definition of approximate - I might have been somewhat loose with it.

OK - some confusion around approximate here:

1) Approx in 

> I was under the impression you were going to implement numerical derivatives.

Yes, but I am also adding explicit differentiation in two cases: 1) the user passes the explicit `inverse_func_diff` or 2) a scaler transform is used.

In these cases, the Jacobian is exact and therefore so is the log_pdf.

> I have now made a draft here in the base class: #610
> 
> This is due to realizing that this is nothing specific to the transformed distribution - any distribution that does not have a `pdf` implemented but a `cdf` can produce a numerical `pdf` like this.

Understood, but the TD doesn't have a `cdf` so I'm not sure this helps with the LogLoss issue.

(edit) or TD is this implemented via the BaseDistribution `cdf`?

> This should, in particular, fix it for the `TransformedDistribution` - do you want to give it a test?

Yes - happy to test. It will be interesting to compare this to the exact methods above.

fkiraly:
> Yes, but I am also adding explicit differentiation in two cases: 1) the user passes the explicit inverse_func_diff or 2) a scaler transform is used.

> In these cases, the Jacobian is exact and therefore so is the log_pdf.

I know, but I think we need to add these gradually. Numerical differentiation is a sensible default in the absence of anything else.

Therefore imo it makes sense to put it even in the base class as a the last fallback.

> Understood, but the TD doesn't have a `cdf` so I'm not sure this helps with the LogLoss issue.

Oh, I see. You would need the inverse transformation for that.

joshdunnlime:
> > Would we like to rollback the changes to ttr or are these beneficial? E.g. is a user wants to dig into the internals of TTR/TD the indices are shared between TD and the wrapped distribution.
> 
> What are the indices currently? I think they should be the same between row and column index.

Prior to the change, always a pandas `RangeIndex` for the columns. I have named columns so it was raising an index error like "`col` not found in y.columns".

fkiraly:
> Prior to the change, always a pandas RangeIndex for the columns. I have named columns so it was raising an index error like "col not found in y.columns".

Is this a bug in the wrapped regressor then? It should return index and columns in line with the `X` and `y` of the `predict`.

fkiraly:
I have now opened an additional PR #611, this adds an exact `cdf` for the TD using the transform of the distribution.

Combined with this, #610 should be able to produce a reasonable `pdf` through numerical differentiation.

The sequence of computations would be, when calling `pdf`:

* the TTR passes the inner transforms fitted `transform` to the TD `inverse_transform`
*  an exact `cdf` is computed using the `inverse_transform` argument that is passed (the TTR always has this because the inverse of the inverse is simply `transform`)
* finally, the base class default of `pdf` does numerical differentiation on `cdf`, calling it multiple times

(the user should see nothing of this, optimally)

joshdunnlime:
closed in favour of #610 #611 and #612

PR #608, opened 2025-10-03
Base and Differentiable Transformer

joshdunnlime:
#### Reference Issues/PRs
#605 and #601

Implements the Differentiable Transformer as discussed in #605.

No

See #605 for details.

Not yet. Tests to follow

##### For all contributions

##### For new estimators

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,4 +41,5 @@ sdist/*
 docs/api/*
 docs/_build/*
 cover/*
+.coverage.*
 MANIFEST
--- a/skpro/compose/__init__.py
+++ b/skpro/compose/__init__.py
@@ -1,0 +1,13 @@
+"""Composition utilities for skpro."""
+
+__all__ = [
+    "BaseTransformer",
+    "BaseDifferentiableTransformer",
+    "DifferentiableTransformer",
+]
+
+from skpro.compose._transformer import (
+    BaseDifferentiableTransformer,
+    BaseTransformer,
+    DifferentiableTransformer,
+)
--- a/skpro/compose/_transformer.py
+++ b/skpro/compose/_transformer.py
@@ -1,12 +1,13 @@
-"""skpro Transformers for regression."""
+"""Transformers for skpro."""
 
 import numpy as np
-from sklearn.base import clone
+from sklearn.base import TransformerMixin, check_is_fitted, clone
+from sklearn.exceptions import NotFittedError
 
 from skpro.base import BaseEstimator
 
 
-class BaseTransformer(BaseEstimator):
+class BaseTransformer(BaseEstimator, TransformerMixin):
     """Base class for transformer objects."""
 
     def __init__(self, transformer):
@@ -17,8 +18,17 @@ class BaseTransformer(BaseEstimator):
         transformer : callable, optional
             Maybe only allow sklearn transformers for now.
         """
-        self.transformer_ = clone(transformer) if transformer else None
+        self.transformer = transformer
         super().__init__()
+
+    def _fit_with_fitted_transformer(self):
+        """Fit with already fitted transformer if possible."""
+        try:
+            check_is_fitted(self.transformer)
+            self.transformer_ = self.transformer
+        except NotFittedError:
+            pass
+        return self
 
     def fit(self, X, y=None):
         """Fit transformer to y.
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -41,4 +41,5 @@ sdist/*
 docs/api/*
 docs/_build/*
 cover/*
+.coverage.*
```
joshdunnlime, line 44: Added as tests create `.coverage.Mac` for each process.

fkiraly (commented):
Great!

Would this solve the fitting/reset problem?

joshdunnlime:
I had already made some changes yesterday before your reply stating your preference to just passing the additional functions ([option 2](https://github.com/sktime/skpro/issues/601#issuecomment-3366441361)).

In the BaseTransformer, I have implemented an sklearn -> skpro transformer. The main idea is that the (skpro) BaseTransformer (BaseT) can take an sklearn transformer an wrap it in one of two ways:
1) Very typical use case: an unfitted sklearn transformer is passed to BaseT at initialisation and is then fitted. We can then extend BaseT for DiffBaseT and DiffT as expected.
2) The less typical case: a fitted sklearn transformer is passed to BaseT at initialisation. The BaseT uses this instance as if it was fitted with it. This only happens when the `_fit_with_fitted` method is called. If we want to refit then calling `fit` does that as normal.

Why is this needed?
Whether we wish to pass all three functions/methods (transform, inverse and inverse_diff) or the transformer class itself, we would like to separate the differentiation logic out from the TransformedDistribution (TD). When the TransformedTargetRegressor (TTR) passes a transform (either function or class) off the TD, the TD has no way of accessing `y` (or named `X` internally when working with transformers) and therefore fitting the DiffT.

Reset Issue
I don't think that the reset issue is a problem here. `predict_probab` returns the TD but doesn't seem to ever reset the DiffT. Perhaps this is because inside TD `fit` is never called (either on TD or on DiffT).

joshdunnlime:
This gives us the following options for using a TTR that returns a TD that can use log_loss:

Pass sklearn transformer

```python
# wraps the fitted MMS in the DiffT internally
xgb = XGBoostLSS(**params)
mms = MinMaxScaler()
pipe = TransformedTargetRegressor(regressor=xgb, transformer=mms)

pipe.fit(X=Xy[["feature"]], y=Xy["target"])
td = pipe.predict_proba(Xy[["feature"]])
LogLoss()(y_true=Xy["target"], y_pred=pp)
```

Pass skpro DiffT

```python
# this calls fit on the DiffT and subsequent MMS
xgb = XGBoostLSS(**params)
mms = DifferentiableTransformer(transformer=MinMaxScaler())
pipe = TransformedTargetRegressor(regressor=xgb, transformer=mms)

pipe.fit(X=Xy[["feature"]], y=Xy["target"])
td = pipe.predict_proba(Xy[["feature"]])
LogLoss()(y_true=Xy["target"], y_pred=pp)
```

The transformer that is not used in the TD `td` is always a `DifferentiableTransformer` so LogLoss can always be called (for some "well-behaved" function).

This also means we can still pass an `mms.transform` and wrap it in a DiffT:

```python
mms = MinMaxScaler()
mms.fit(Xy[["target"]])

trafo = TransformedDistribution(
    distribution=Normal(),
    transform=mms.transform,
    assume_monotonic=True,
    index=Xy.index,
    columns=['target'],
)

trafo.log_pdf(Xy[["target"]])
```

fkiraly:
> we would like to separate the differentiation logic out from the TransformedDistribution (TD)

Just to clarify, I think it would be good if TD had:

* some default for numerical differentiation in it.
* the option to pass inverse transform.

What makes sense is to separate differentiation logic that is at least one of:

* customisable, e.g., allow the user to pick the degree of approximation or similar parameters determining numerical strategy
* auto-diff-like, e.g., dispatching on arguments to add auto-diff-like features to an `sklearn` transformer

joshdunnlime:
> > we would like to separate the differentiation logic out from the TransformedDistribution (TD)

> Just to clarify, I think it would be good if TD had:
> 
> * some default for numerical differentiation in it.
> * the option to pass inverse transform.

Done in #612

> What makes sense is to separate differentiation logic that is at least one of:
> 
> * customisable, e.g., allow the user to pick the degree of approximation or similar parameters determining numerical strategy
> * auto-diff-like, e.g., dispatching on arguments to add auto-diff-like features to an `sklearn` transformer

Solution in #612 makes it very easy to add this. Closing this in favour of #612

PR #610, opened 2025-10-05
[ENH] distributions approximate `pdf` and `log_pdf` via `cdf` numerical derivative if not implemented

fkiraly:
Adds a new approximative default to `BaseDistribution`: if  `pdf` and `log_pdf` are not implemented, approximates them via 6-th order central differences if an exact `cdf` is available.

This works towards https://github.com/sktime/skpro/issues/601 but does not address it entirely.


fkiraly:
FYI @joshdunnlime, this adds approximative `pdf` and `log_pdf`

Issue #559, opened 2025-06-30
[ENH] Default `_iloc` and `_iat` to include handling of distribution objects


fkiraly:
In a number of PR, it became necessary to write custom `_iloc` and `_at`, in a couple different cases of composite distributions: #555, #557.

The boilerplate to subset distributions always seems the same (simply calling their respective `iloc`), so it feels like this should be the default for handling distribution objects.

The current default `_iloc` handles array-like objects, so we could add another branch to the logic handling distribution objects (after checking type) the same way.

SimonBlanke:
I'll look into this issue.

PR #611, opened 2025-10-05
&joshdunnlime [ENH] `TransformedDistribution` and `TransformedTargetRegressor` `cdf` support

fkiraly:
This PR adds `cdf` support to `TransformedTargetRegressor`, via changes in `TransformedDistribution`:

* `TransformedDistribution` now accepts `inverse_transform`, which can be used for an exact `cdf`
* `TransformedTargetRegressor` passes the fitted `self.transformer_transform` to the `inverse_transform` of `TransformedDistribution`

Together with https://github.com/sktime/skpro/pull/610, it means that `TransformedTargetRegressor` can now produce distributions with reasonably reliable `cdf` and `pdf`.

Goes partially towards #601.

### DIFF
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -275,7 +275,24 @@ class TransformedDistribution(BaseDistribution):
             "columns": pd.Index(["a", "b"]),  # this should override n_row.columns
         }
 
-        return [params1, params2]
+        # array case example with inverse transform
+        n_array = Normal(mu=[[1, 2], [3, 4]], sigma=1, columns=pd.Index(["c", "d"]))
+        params3 = {
+            "distribution": n_array,
+            "transform": np.exp,
+            "inverse_transform": np.log,
+            "index": pd.Index([1, 2]),
+            "columns": pd.Index(["a", "b"]),  # this should override n_row.columns
+        }
+
+        # scalar case example with inverse transform
+        params4 = {
+            "distribution": n_scalar,
+            "transform": np.exp,
+            "inverse_transform": np.log,
+        }
+
+        return [params1, params2, params3, params4]
 
 
 def is_scalar_notnone(obj):
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -150,8 +150,8 @@ class TransformedDistribution(BaseDistribution):
                 "ppf is implemented only for monotonic transforms, "
                 "set `assume_monotonic=True` to use this method"
             )
-        elif self.inverse_transform is not None:
-            return super()._ppf(p)
+        elif not self.assume_monotonic and self.inverse_transform is not None:
+            return super().ppf(p)
 
         trafo = self.transform
 
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -286,8 +286,8 @@ class TransformedDistribution(BaseDistribution):
         n_array = Normal(mu=[[1, 2], [3, 4]], sigma=1, columns=pd.Index(["c", "d"]))
         params3 = {
             "distribution": n_array,
-            "transform": np.log,
-            "inverse_transform": np.exp,
+            "transform": lambda x: 2*x,
+            "inverse_transform": lambda x: x/2,
             "index": pd.Index([1, 2]),
             "columns": pd.Index(["a", "b"]),  # this should override n_row.columns
         }
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -157,7 +157,9 @@ class TransformedDistribution(BaseDistribution):
         2D np.ndarray, same shape as ``self``
             ppf values at the given points
         """
-        p = pd.DataFrame(p, index=self.index, columns=self.columns)
+        if self.ndim != 0:
+            p = pd.DataFrame(p, index=self.index, columns=self.columns)
+
         if not self.assume_monotonic and self.inverse_transform is None:
             raise ValueError(
                 "if inverse_transform is not given, "
@@ -200,7 +202,9 @@ class TransformedDistribution(BaseDistribution):
 
         inv_trafo = self.inverse_transform
 
-        x = pd.DataFrame(x, index=self.index, columns=self.columns)
+        if self.ndim != 0:
+            x = pd.DataFrame(x, index=self.index, columns=self.columns)
+
         inv_x = inv_trafo(x)
         cdf_res = self.distribution.cdf(inv_x)
 
--- a/skpro/distributions/trafo/_transformed.py
+++ b/skpro/distributions/trafo/_transformed.py
@@ -168,7 +168,7 @@ class TransformedDistribution(BaseDistribution):
 
         return outer_ppf
 
-    def _cdf(self, p):
+    def _cdf(self, x):
         """Cumulative distribution function.
 
         Parameters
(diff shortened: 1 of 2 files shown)


fkiraly:
FYI @joshdunnlime, this is what I meant for a quick fix

fkiraly:
@joshdunnlime, any idea why the test is failing? cdf and ppf do not seem to be inverse to each other.

joshdunnlime:
@fkiraly is `cdf` meant to be approximate? The class tags are approx for cdf but the `get_tags` show `cdf` is exact. I can see this get set if there is an inverse passed. However, I get:

> <class 'skpro.distributions.trafo._transformed.TransformedDistribution'> does not have a numerically exact implementation of the 'cdf' method, it is filled in by approximating the expected value by the indicator function on 1000 samples.

So the `_cdf` seems to be falling back to the approx (I get that warning when running in a script). I had similar issue with the `_pdf` and `_log_pdf` and it was an indices issue (set index and column names on the transformed). See [here](https://github.com/sktime/skpro/pull/612/files#diff-c89d9489064d49d9caf2fb49333090330830b224c416d6cd8b0bd3eb2285cf71R158) and [here](https://github.com/sktime/skpro/pull/612/files#diff-250f0e6f5c35534e02c71782ae50bf09cb2f997c031ed4c729dcb41653638342R110).

joshdunnlime:
@fkiraly - What are the methods `loc/iloc` doing? It's being called by the `_pdf` via boilerplate but the class is being initialed again but without passing the inverse_transform.

joshdunnlime:
@fkiraly merge https://github.com/joshdunnlime/skpro/tree/ttr-cdf-pdf-add-inv or just add

```python
        inverse_transform=self.inverse_transform,
```

to the cls call in `_iloc`.

joshdunnlime:
This still raises a `X does not have valid feature names, but MinMaxScaler was fitted with feature names` so I think there is still the indices issue mentioned above, but the tests pass.

fkiraly:
> @fkiraly - What are the methods `loc/iloc` doing?

These are methods to subset or reorder indices of the array distribution. This is called in `loc` and `iloc` which works like for `pandas.DataFrame`, the inner `_loc` / `_iloc` must define equivalent operations for distributions etc.

If the distribution is parametric, this is taken care of by the default, but in general, e.g., if the distribution is composite, it needs to be done manually currently.

There is space to define a broader default including distribution objects as components, but that is currently an open issue:
https://github.com/sktime/skpro/issues/559

> It's being called by the `_pdf` via boilerplate but the class is being initialed again but without passing the inverse_transform.

I see, that must be it!

fkiraly:
> This still raises a `X does not have valid feature names, but MinMaxScaler was fitted with feature names` so I think there is still the indices issue mentioned above, but the tests pass.

can you outline your understanding of why this occurs? Which two objects exactly where, when passed to `MinMaxScaler` have missing or inconsistent feature names?

skpro/distributions/trafo/_transformed.py
```diff
@@ -99,13 +123,19 @@ def _iloc(self, rowidx=None, colidx=None):
...
+        params_dict = self.get_params(deep=False)
+        [params_dict.pop(param) for param in POP_PARAMS]
+
         return cls(
             distribution=distr,
-            transform=self.transform,
-            assume_monotonic=self.assume_monotonic,
             index=new_index,
             columns=new_columns,
+            **params_dict,
```
fkiraly, line 138: I think this is safer and more extensible. This might already be an almost solution for #559 (only needs to be combined with checks for type?)

fkiraly:
> > This still raises a `X does not have valid feature names, but MinMaxScaler was fitted with feature names` so I think there is still the indices issue mentioned above, but the tests pass.
> 
> can you outline your understanding of why this occurs? Which two objects exactly where, when passed to `MinMaxScaler` have missing or inconsistent feature names?

If you have a fix, @joshdunnlime, could you open a PR with only the fix? So we can merge it quickly while the more complicated design questions remain open?

joshdunnlime:
Done. It is literally that one line to allow the exact `cdf`.

skpro/distributions/trafo/_transformed.py
```diff
@@ -148,6 +181,37 @@ def _ppf(self, p):
...
+        -------
+        2D np.ndarray, same shape as ``self``
+            cdf values at the given points
+        """
+        if self.inverse_transform is None:
+            return super()._cdf(x)
+
+        inv_trafo = self.inverse_transform
+
+        inv_x = inv_trafo(x)
```
joshdunnlime, line 208: This is where `MinMaxScaler` in my local script raised the warning.

I used:

```python
warnings.filterwarnings('error')
```

to catch and debug.

joshdunnlime:
> > This still raises a `X does not have valid feature names, but MinMaxScaler was fitted with feature names` so I think there is still the indices issue mentioned above, but the tests pass.
> 
> can you outline your understanding of why this occurs? Which two objects exactly where, when passed to `MinMaxScaler` have missing or inconsistent feature names?

I have added a comment on where the warning is raised. The `inverse_tranform` function returns a numpy array. This was easy to find when raising warning as errors (see comment).

I think the solution I have implemented in #612, where every function or transformer is wrapped in a DifferentiableTransformer would be the nicest way to fix this. It is then very easy to keep all of the transformation logic in there and out of the `TransformedDistriubtion`. We can guarantee that `tranform`, `inverse_transform` and their `_diff` functions always return a dataframe. This means much less code like:

```python
if not isinstance(x_t, pd.DataFrame):
    x_t = pd.DataFrame(x_t, index=x.index, columns=x.columns)
else:
    x_t.columns = x.columns
    x_t.index = x.index
```

This is the case for `_pdf`, `_log_pdf`, `_cdf` and probably all the other functions that need transformer outputs.

skpro/distributions/trafo/_transformed.py
fkiraly: I see.

I think this is resolved by ensuring `x` is a `pd.DataFrame` when we pass it.

fkiraly:
> We can guarantee that `tranform`, `inverse_transform` and their `_diff` functions always return a dataframe.

Can you explain why this would be the case? The data arrives inside `_cdf` and `_pdf` as `numpy` already.

This may have been an unfortunate design choice (not sure about this yet), but that also means whatever we do with `x` or `p`, it starts with numpy. So, the `DifferentiableTransformer` may not get the `pd.DataFrame` it expects?

fkiraly:
I have opened a new design issue here: https://github.com/sktime/skpro/issues/615
I think there are some interesting questions about nested/composite distributions where I would appreciate your opinion! Given that you have dug into the problems above.

I will now merge this PR to expedite the various improvements.

joshdunnlime:
> > We can guarantee that `tranform`, `inverse_transform` and their `_diff` functions always return a dataframe.
> 
> Can you explain why this would be the case? The data arrives inside `_cdf` and `_pdf` as `numpy` already.
> 
> This may have been an unfortunate design choice (not sure about this yet), but that also means whatever we do with `x` or `p`, it starts with numpy. So, the `DifferentiableTransformer` may not get the `pd.DataFrame` it expects?

I'm talking more about the output of the cdf, ppf and pdf. We are nearly always applying some transform in the TD so having the outputs (e.g. Jacobian) as dataframe would tidy up the TD code. It does deviate from _default_ sklearn behaviour but sklearn does have a transformer to dataframe setting.

Skpro TD method outputs are dataframes so this keeps consistency with that and makes debugging easier IMO.

> I have opened a new design issue here: #615 I think there are some interesting questions about nested/composite distributions where I would appreciate your opinion! Given that you have dug into the problems above.
> 
> I will now merge this PR to expedite the various improvements.

I'll take a look. Thanks.

PR #612, opened 2025-10-05
[ENH] Add Differentiable Transformer and Exact PDFs for TargetTransformRegressor

joshdunnlime:
#### Reference Issues/PRs
Background on why we want this: #601 and #605
Supersedes this PR: #608

This implementation adds a DifferentiableTransformer (DT). This acts as a wrapped around an sklearn transformer and gives the user the option to use either (in order):
1) an explicit derivative via `inverse_func_diff`.
2) an explicit derivative for a scaler transformer via `scaler_`.
3) Numerical differentiation of the `inverse_transform`.

The derivative available for both the (forward) transform and inverse_transform.

The DT has a coerce classmethod that takes a sklearn transformer or a function and coerces it to a DT.

In addition to the above, it also makes changes to #605 to apply the Jacobian (derivative of the transform/inverse_tranform) to the `pdf` and `log_pdf`. In cases 1) and 2) above this returns the exact pdf and log_pdf.

This preserves the current functionality of being able to pass a function to TTR (TransformedTargetRegressor) and allows a user to extend this by passing an sklearn transformer or their own DT. The final point allows the user to configure their DT with either the explicit derivative, or to pass kwargs that can be used for numerical differentiation. In theory, it also allows a user to pass these kwargs as hyperparameters via a gridsearch/optimisation.

No.

1) Implementation of the DT is the most important part:

   - Are we happy with the transformer inheriting from sklearn?
   - Do we need this much abstraction at this stage?
   - Do we wish to add/implement tags for this?

2) Changes to TD (TranformedDistribution) - the main change here is internally handling a transformer instead of a function.
3) Changes to the TTR are minor.

On points 2) and 3), note that if we wish to explicitly pass `transform` only as a method/function, it is very straightforward to change this back. We would simply need to add `inverse_transform` and `inverse_func_diff` as kwargs to the TD. This does mean adding more kwargs if we wish to support numerical derivative kwargs.

Secondary feedback 

Yes, added a param3 to TD, passing a FunctionTransformer instead of just a function.

#### Any other comments?

I had the bulk of this implementation down prior to the discussion on what we pass to TTR/TD [here](https://github.com/sktime/skpro/issues/601#issuecomment-3362967573). Again, it is trivial to move the creation of the DT into TTR and pass the `transform`, `inverse_transform` and `inverse_func_diff`, but I feel the current approach is much cleaner as it keep the majority of the new logic in the DT and within the new _transformer.py module.

##### For all contributions

##### For new estimators


joshdunnlime:
Usage examples:

Create some data:

```python
import numpy as np
import pandas as pd
from sklearn.preprocessing import MinMaxScaler

from skpro.metrics import LogLoss, CRPS
from skpro.regression.xgboostlss import XGBoostLSS
from skpro.compose import DifferentiableTransformer

size = 1000

X = np.random.normal(1, 1, size)
y = (2 * X + np.random.normal(0, 0.5, size)) # / 15 + 0.5 # for Beta

Xy = pd.DataFrame(
    {"target": y, "feature": X},
    index=range(size)
)
```

Now we can create pass an sklearn transform to TTR:

```python
xgb = XGBoostLSS(**params)
mms = MinMaxScaler((0.1, 0.9))
pipe = TransformedTargetRegressor(regressor=xgb, transformer=mms)
```

or a DT:

```python
xgb = XGBoostLSS(**params)
mms = MinMaxScaler((0.1, 0.9))
mms = DifferentiableTransformer(transformer=MinMaxScaler())
pipe = TransformedTargetRegressor(regressor=xgb, transformer=mms)
```

Both are handled the same:

```python
pipe.fit(X=Xy[["feature"]], y=Xy["target"])
pp = pipe.predict_proba(Xy[["feature"]])
p = pipe.predict(Xy[["feature"]])
CRPS()(y_true=Xy["target"], y_pred=pp)
LogLoss()(y_true=Xy["target"], y_pred=pp)
```
> 0.41678

And in either case, if we inspect the TD (`pp`) returned by `predict_proba` we get a DT as the transformer.

```python
pp.transformer_
```
> DifferentiableTransformer(transformer=MinMaxScaler(feature_range=(0.1, 0.9)))

fkiraly:
I tried to resolve the conflicts, have a look

joshdunnlime:
@fkiraly

This should give working implementation.

There are a couple design choice that it might be worth discussing:

1) **numdifftools** was added to handle fast and accurate numerical differentiation.
   - Added as an optional extra. We could add an "install numdifftools/extras" warning when the fallback is used so users know there is a better option.
   - scipy derivative function has some breaking changes between versions. We could handle this by checking versions however, the scipy implementation is a simplified version of numdifftools so numdifftools is likely to give more robust results.

2) Do we need a DifferentiableTransformer (DT)? And does it need to inherit from BaseTransformer, BaseEstimator etc?
   - We could have a DifferentiableFunction class you as mentioned previously. My main concern with it being a Transformer is that it makes design choices about the transformer base class which might narrow the design choices for that in the future.
   - Does it need to inherit from sklearn's BaseEstimator or should it have a pure skpro/skbase implementation?

3) This implementation can pass a function, an sklearn transformer (skT) or DT to TTR (all via the `transform` kwarg). It doesn't support passing `inverse_transform` or `inverse_tranform_diff` but it have kept that in mind while designing this. It should be trivial to extend this (allow passing an SKT, DT or `inverse_transform` + `inverse_transform_diff`), or to constrain this to only allow `inverse_transform` + `inverse_transform_diff`.
   - I am in favour of supporting skT or DT as it feels more sklearn-like. E.g. wrap an sklearn-like object in another sklearn-like object and then just fit-predict.
   - Supporting all of the above options doesn't introduce any breaking changes.

4) Should we consider numerical differentiation as approx? The current implementation does, however, the "change-of-variables" only works with well-behaved functions. These are incredibly easy for numerical differentiation to handle, with errors typically being in the order of the 7th decimal place when comparing the exact derivative of a function to the numerical derivative for CRPS, LogLoss and MAE of the log_pdf.

5) Currently, multivariate targets are handled completely independently, that is to say, the two targets are transformed independently. Thus, the Jacobian is treated as a diagonal matrix with no partial derivatives. This is mathematically correct for all sklearn transformers where the transform (F_i(y)) is only dependent on it's respective target (y_i). Other custom transformers could be created, where some F_i(y) is dependent on any/all y, but I would consider this extremely unlikely.

In addition to the above, the documentation needs some tidying, and comments, tags and licenses need updating.

Further improvements:
1) Tidy up repeated handling of index and columns.
2) Tests - do we need specific tests added?
3) Tests don't catch false-negative exact for pdf_log_pdf. E.g. if the tag logic is changed and something that is exact get's tagged as "approx" this won't be caught.
4) The fallback derivative could also be improved.

fkiraly:
> This should give working implementation.

Oh, very nice! I will review but first leave some comments on the design choices.

> 1. numdifftools was added to handle fast and accurate numerical differentiation.

I think we should not take a dependency on `numdifftools` - the package looks abandoned. Last release in 2022, and sole maintainer. It implies python 3.10 or python 3.11, so will lapse into outdated python within 2 years.

I will contact the author to see what is going on, but for now I think we should avoid it.

> 2. Do we need a DifferentiableTransformer (DT)?

That is unclear to me and we should really think carefully about pros and cons.

> We could have a DifferentiableFunction class you as mentioned previously.

But then we would need to turn the fitted transformer into a `DifferentiableFunction`, right? Feels less extensible to me. Or how would everything fit together? Would appreciate your thoughts.

> Does it need to inherit from sklearn's BaseEstimator or should it have a pure skpro/skbase implementation?

If transformer, I am not in favour strongly of either option. If a function, I would favour an skbase implementation strongly, because it is too far from the sklearn transformer.

> 3. This implementation can pass a function

I like this, though I still need time to think about it too.

> 4. Should we consider numerical differentiation as approx?

Where exactly? I have already implemented numerical differentiation (with sixth-order approximation) in the `pdf` and `log_pdf` defaults.

> 5. Currently, multivariate targets are handled completely independently,

That is correct, currentls `skpro` cannot support multivariate distributions. That is a bit of a project, and the API is not fleshed out.

I think that API question needs to be answered before we move to the differentiation question (but it might make sense to look at both in close sequence ot validate the design).

E.g., do we need to have a separate method for multivariate `pdf`? Note that the current `pdf` produces marginals of an independent distribution across variables. But for joint distributions the marginals do not carry the full information, so there needs to be a way to return joint `pdf`.

fkiraly:
FYI, I opened an issue to collect API discussion for multivariate distributions here:
https://github.com/sktime/skpro/issues/622

joshdunnlime:
> I think we should not take a dependency on `numdifftools` - the package looks abandoned. Last release in 2022, and sole maintainer. It implies python 3.10 or python 3.11, so will lapse into outdated python within 2 years.

I'll revert to scipy with version handling for `misc.derivative` and `differentiation.derivative` function or our own.

> But then we would need to turn the fitted transformer into a `DifferentiableFunction`, right? Feels less extensible to me. Or how would everything fit together? Would appreciate your thoughts.

I've had much more of a think about having a `DifferentiableFunction` vs `DifferentiableTransformer`. I am confident the DiffT is the right way to go:
- It follows existing sklearn-like patterns. DiffF introduces a new pattern.
- Adding numerical differentiation is unrestrictive to future development of BaseTransformer. The `BaseDiffT` only requires `inverse_diff` and `transform_diff` as public methods and `_fit_with_fitted_transformer` as a private method and currently has `_numerical_diff` but that is not a hard requirement.

My impression is that if we were implementing TTR and TD from scratch, with `LogLoss`, `pdf` and `log_pdf` support, I would probably go with something like:

```python
# what this PR has
xgb = XGBoostLSS(**params)
mms = MinMaxScaler((0.1, 0.9))
dt = DifferentiableTransformer(transformer=mms)
pipe = TransformedTargetRegressor(regressor=xgb, transformer=dt)
```

or

```python
# we could easily add "dft" with the current PR
xgb = XGBoostLSS(**params)
dft = DifferentiableFuncTransformer(func, inverse_func, func_diff, inverse_diff)
pipe = TransformedTargetRegressor(regressor=xgb, transformer=dft)
```

I think the main issue here is extending/changing the TTR input options, e.g. point 3 above. For example, the `_fit_with_fitted_` is only used to accommodate patterns unlike the above examples.

> If transformer, I am not in favour strongly of either option. If a function, I would favour an skbase implementation strongly, because it is too far from the sklearn transformer.

I would keep it inheriting from sklearn for now then. My understanding is it should be straightforward to replace this with skbase if/when needed.

> > 3. This implementation can pass a function
> 
> I like this, though I still need time to think about it too.

It would be good to get some more thoughts on this. As mentioned above, I think this is the main constraint in how this gets implemented.

> > 4. Should we consider numerical differentiation as approx?

> Where?

In the `pdf` and `log_pdf`, when applying the Jacobian. The plain `pdf` and `log_pdf` are always exact for the transformed distribution. For the original distribution, the `pdf` and `log_pdf` are always exact when using `scale_` or `inverse_transform_diff` for the Jacobian. When using `_numerical_diff`, the `pdf` and `log_pdf` are still exact, but the Jacobian is technically an approximation. Therefore, in this case, I have set the tag as `"approx"`. However, for all [well-behaved functions](https://mc-stan.org/docs/2_18/reference-manual/change-of-variables-section.html) (which is a theoretical prerequisite), numerical differentiation would effectively give an exact Jacobian (to some some relatively small rounding error). Do we wish to keep this as `"approx"` or change it to "`exact`"?

> 5. Currently, multivariate targets are handled completely independently,

I assumed that this was fully supported. Until the these other parts are implemented it will be difficult to design around it. I would consider this a edge-case - sklearn transformers don't even handle combined column transformations (afaik).

fkiraly:
ok, I also think that transformer is the better way to go after some thinking.

How do we proceed practically - do you want to give it a stab and then we maybe refine?

joshdunnlime:
@fkiraly - I have removed any additional packages and the differentiation is now done by scipy with both the old and new derivative APIs implemented.

The transformer implementation we spoke about is implemented and now just needs refining. I have moved some methods down to the DiffT class so as to keep the BaseT as generic as possible.

fkiraly:
Nice! Will review in the coming days - this is much appreciated, but I think I need more time to digest the API design.
(please ping if I need more than a few days)

joshdunnlime:
> Nice! Will review in the coming days - this is much appreciated, but I think I need more time to digest the API design. (please ping if I need more than a few days)

@fkiraly - just bringing this one your attention again. Thanks

fkiraly (changes requested):
Great!

Two remarks on the design:

* can you make sure that `TransformedDistribution` does not deviate from the current logic in the function case? The current design causes default approximations for the pdf etc to live in two places.
* I think your transformation design is not quite there yet:
    * it should work if `transform` is called on a different `X` than `fit`
    * I do not get the index saving and restoring logic in `fit` and `transform`. Why is this necessary? Should the index not be the same as `X` passed in `transform`, and the columns determined by what the transformation does?

joshdunnlime:
I've made some changes to the structure here:
1. Refactored BaseTransformer. Previously, this was too strict. It means if anyone wanted to add an skpro transformer in the future they would have had a bit of a battle to do so. It now follows the `_fit`, `_transform` style of skpro. Repeated this for `BaseDifferentiableTransformer` so that transformers with differentiable `transform` or `inverse_transform` can be created, without needing to pass an sklearn transformer (as before).
2. Added a `WrapTransformer` class to keep the need for wrapping outside of `BaseDifferentiableTransformer`.
3. Tidied up the naming used inside the `TransformedDistribution` (TD), transformers and `TransformTargetRegressor`. The main issue here is one we have spoken about before... The `transform` in TD is actually the `inverse_transform` that you have on an sklearn or skpro transformer. I have added info to the docstrings and left explicit comments in the part of the code where this happens.
4. Raise a warning if a user passes an sklearn transformer-method to TD and suggests passing the whole transformer to `transform` instead. This could easily be a point of confusion if a user on their first iteration passes a function, then refactors to passing a `FunctionTransformer.transform` method with their function wrapped in it. They are likely to do `TransformDistribution(transform=ft.transform)` which would incorrect.
5. Minor improvements to the Jacobian calculations structure to fit with the issue in 3.
6. Added tests for `_coerce_to_differentiable` (this is THE key component in linking the new DifferentiableTransformer to the TD).
7. Added fast end-to-end tests for the Jacobian calculations on linear (MinMaxScaler) and non-linear transformations to check that "[change-of-variables](https://mc-stan.org/docs/2_18/reference-manual/change-of-variables-section.html)" brings PDF in TD inline with PDFs that are adjusted with a manually computed Jacobian.

joshdunnlime:
@fkiraly
Ready for a review. I've added some tests to satisfy the previous queries you raised. The refactored transformer classes should be a lot clearer now.

The final part if just check the docs and I'd like to add some docstring examples but think this can be finalised after your review.

joshdunnlime:
@fkiraly - are you happy with the new structure? If so, I happy to improve docs and add examples, ready for merging. Thanks.

skpro/compose/_transformer.py
```diff
@@ -0,0 +1,649 @@
...
+from sklearn.exceptions import NotFittedError
+
+from skpro.base import BaseEstimator
+
+# Optional import for scipy derivative
+HAS_SCIPY_DERIVATIVE = False
+scipy_derivative = None
+SCIPY_DERIVATIVE_NEW_API = False
+
+try:
```
fkiraly: it is very dangerous to put try/except at module level, since it is always executed at module load, and masks genuine exceptions; also, try/except has a high runtime due to the exception propagation.

I would recommend: move this logic into a function, and inside the function use the `_check_soft_dependencies` utility.

skpro/compose/_transformer.py
```diff
@@ -0,0 +1,649 @@
...
+                else pd.RangeIndex(Xt.shape[0])
+            )
+            columns = (
+                self.columns_
+                if self.columns_ is not None
+                else pd.RangeIndex(Xt.shape[1])
+            )
+            return pd.DataFrame(Xt, index=index, columns=columns)
+
+    def _fit_with_fitted_transformer(self, index=None, columns=None):
```
fkiraly: `sklearn` has a `FrozenEstimator` that does something similar - pointing this out since that is an alternative design that might be interesting:
https://scikit-learn.org/stable/modules/generated/sklearn.frozen.FrozenEstimator.html

skpro/distributions/trafo/_transformed.py
```diff
@@ -18,26 +23,50 @@ class TransformedDistribution(BaseDistribution):
     * ``transform``: a function that is applied to the distribution.
       This can be entry-wise, or a ``pandas.DataFrame`` to ``pandas.DataFrame``
       function that can be applied to samples from ``distribution``.
+      NOTE: this is the transform that is applied to the distribution, not
```
fkiraly: I do not quite get the note. I think this is too confusing, instead I would add more details on what the function is in terms of random variables. It is simply the distribution of f(X) if X is a random variable distributed according to `distribution`, and f=`transform`, no?

I think the docstring should be very explicit what the distribution this represents is, in all cases of what `transform` can be.

skpro/distributions/trafo/_transformed.py
```diff
@@ -93,23 +128,172 @@ def __init__(
                 if columns is None:
                     columns = pd.RangeIndex(n_cols)
 
+        self.transformer_ = _coerce_to_diff_transformer(
```
fkiraly: hm, I would not coerce it but just check if it is one.

What is the point of coercing the transformer if it is not already one?

joshdunnlime: It seems a lot cleaner to move all of the `if isinstance` logic into a single, testable function. I can't think of a good reason to move it out of there, so maybe the name is a bit misleading? Happy to change that if that is the main problem.

skpro/distributions/trafo/_transformed.py
joshdunnlime: Agreed. This could be clearer and should be explicit about all potential cases of `transform`.
Theoretically, the RV framing is useful - I will add that.

skpro/distributions/trafo/_transformed.py
fkiraly: this is not about encapsulation - external function is fine - but why coerce everything to a `DifferentiableTransformer`, is my question. 

Namely, in the case where the transformer or function does not have autodiff capability. I think we still want to distinguish at least the two cases internally, where we have a transformer only, vs where the case where it is also autodiff.

fkiraly (changes requested):
I have a few questions for my understanding before I continue the review:

* how are we solving the problem that the `TransformedDistribution` resets the fitted `transformer_` upon passing from `TransformedTargetRegressor`? Previously, we were circumventing this by passing the methods of `self.transformer_` instead of the object.
* Did you, or not, flip around the convention where the `inverse_transform` from the `self.transformer_` in `TransformedTargetRegressor` gets passed as `transform` to `TransformedDsitribution`? There are also two steps where it could flip around after this PR: once when passing it to the `DifferentiableTransformer`, and once when the `DifferentiableTransformer` is applied within the `TransformedDistribution`. Can you explain what your intention is, whether it has changed or not compared to prior, and why you think what you are doing is the right design?

Regarding the design:

* I would wrap in `DifferentiableTransformer` *only*  if we know an exact derivative is available.
* I think there are too many case distinctions and non-DRY duplications throughout - can you please look into this? Is it possible to use, instead of case distinctions, the pattern of coercions? E.g., case distinctions based on `scipy`, why not wrap them in a facade pattern, etc.

joshdunnlime:
> Did you, or not, flip around the convention where the inverse_transform from the self.transformer_ in TransformedTargetRegressor [sic (TTR)] gets passed as transform to TransformedDsitribution [sic (TD)]?

When passing a func or func-inv pair, backward compat on the TD is maintained. However, they are passed to an sklearn FunctionTransformer `ft` via their inverse names: `func -> ft.inverse_func` and `inverse_func -> ft.func`.

> Can you explain what your intention is, whether it has changed or not compared to prior, and why you think what you are doing is the right design?

The rational here is fairly straightforward (though opinionated): sklearn has a very clear standard where the `transform` method on a TTR maps from `y` (original target) to `y_t` (target to model on). Passing a transformer and expecting the user to realise (and some how fit) an sklearn-like transformer is impractical. Therefore, when passing a transformer to `transform`, this is handled internally - the user shouldn't have to worry about this.

> There are also two steps where it could flip around after this PR: once when passing it to the DifferentiableTransformer, and once when the DifferentiableTransformer is applied within the TransformedDistribution.

If a transformer is passed, it simply passes transform -> transform until we get to the TD where TD calls the backward transform `transform`. A "double-flip" only occurs if someone passed a func to transform (which is coerced to an FT).

joshdunnlime:
> I would wrap in DifferentiableTransformer (DT) only if we know an exact derivative is available.

The idea of the DT is to use the facade pattern (your second point). By having the DT we can use explicit differentiation when we have it, otherwise, numerical differentiation is used and the PDF, log-PDF, energy-score and CRPS can all be computed. If we do not coerce to a DT then we need to handle more cases distinctions closer to the user.

skpro/distributions/trafo/_transformed.py
joshdunnlime: Passing to a `DifferentiableTransformer` gains you two things:
1) standardisation - we only need to handle DT case internally and not DT + Transformers.
2) a (numerical issues aside) guaranteed differentiation -> jacobian -> pdf/log_pdf -> CRPS. (for transforms with unstable/badly behaved diffs, change of variables is not mathematically sound).

It adds a small layer of abstraction if you inspect your TransformedDistribution, but I believe this adds a lot more to the user.

joshdunnlime:
> how are we solving the problem that the TransformedDistribution resets the fitted transformer_ upon passing from TransformedTargetRegressor? Previously, we were circumventing this by passing the methods of self.transformer_ instead of the object.

Here we use the `_fit_with_fitted` method. If the transformer is fitted it returns the original instance without re-fitting. You suggested looking into using the `FrozenEstimator` so I will do that but it would be worth aligning on the coercion side of things first as that seems a much more fundamental decision.

fkiraly:
> A "double-flip" only occurs if someone passed a func to transform (which is coerced to an FT).

I do not think this should happen. I think there should only ever be a single flip. Since there are two, something in the design must be problematic, though I cannot pin it down.

Can you please explain, again, very explicitly, in which situation there is a double-flip? Optimally with pseudo-code, since I feel it will be very hard to follow if just described in English.

> The idea of the DT is to use the facade pattern (your second point). By having the DT we can use explicit differentiation when we have it, otherwise, numerical differentiation is used and the PDF, log-PDF, energy-score and CRPS can all be computed. If we do not coerce to a DT then we need to handle more cases distinctions closer to the user.

But we are still duplicating even if we do that; the logic for numerical differentiation that gets called moves from the distribution default `_pdf` to the DT. I think that introduces more case distinctions on the logic in `_pdf` that can be called, instead of reducing their number?

Please correct me if you think I'm wrong.

> so I will do that but it would be worth aligning on the coercion side of things first as that seems a much more fundamental decision.

Can you please write up the exact cases that we would be handling?

I think the cases should be as follows:

**in TD**, on `transform`:

* if function, applied directly; `inverse_transform` can also be passed
* if `sklearn` transformer, applied (fitted) via `transform` and `inverse_transform`
* if DT, its `transform` , `transform_diff`, and `inverse_transform` are used

**in TTR**:

no case distinctions, fitted transform is passed through directly to TD

joshdunnlime:
> I do not think this should happen. I think there should only ever be a single flip. Since there are two, something in the design must be problematic, though I cannot pin it down.

I would say the issue is just that the `TransformedDistribution` uses the term `transform` to mean the backward transform: `g(y_t) -> y`; and uses the term `inverse_transform` to mean the forward transform: `f(y) -> y_t`, where `y` and `y_t` are the original target and transformed target. This is contrary to the standard sklearn nomenclature.

> Can you please explain, again, very explicitly, in which situation there is a double-flip? Optimally with pseudo-code, since I feel it will be very hard to follow if just described in English.

Yes - this happens in only one occasion:
```python
td = TransformedDistribution(transform=fx)
```

where `fx` is some callable AND `fx` is not a `DifferentiableTransformer` AND `fx` is not a `Transformer`. E.g.

```python
assert callable(fx)  # True
assert not isinstance(transform, TransformerMixin)  # True
assert not isinstance(transform, DifferentiableTransformer)  # True
```

In this case only, the first flip happens in the `_coerce_to_diff_transformer` as `transform` is passed to the [FunctionTransformer](https://github.com/joshdunnlime/skpro/blob/07df29d583a8e0b2e50d7f26a509c77c3e7db796/skpro/distributions/trafo/_transformed.py#L617) as `inverse_func=transform` (and `func=inverse_transform`).

In all other cases, the passed transformer (Diff or Sklearn) is just directly [passed through](https://github.com/joshdunnlime/skpro/blob/07df29d583a8e0b2e50d7f26a509c77c3e7db796/skpro/distributions/trafo/_transformed.py#L595) or [lightly wrapped](https://github.com/joshdunnlime/skpro/blob/07df29d583a8e0b2e50d7f26a509c77c3e7db796/skpro/distributions/trafo/_transformed.py#L597).

Any other "flips" are done internally of the TD. In the callable (`fx`) case, this is the second flip. In the passed Transformer case, this will be the only flip. An example of this happening in the TD is [here](https://github.com/joshdunnlime/skpro/blob/07df29d583a8e0b2e50d7f26a509c77c3e7db796/skpro/distributions/trafo/_transformed.py#L375).

IMO, the "cleanest" solution would be to deprecate the passing of functions to TD and only pass transformers. This would mean at most, there was a single flip (internally to TD) and this private API could be inverted so no flips are done. Of course, the issue here is back-compat!

joshdunnlime:
> But we are still duplicating even if we do that; the logic for numerical differentiation that gets called moves from the distribution default _pdf to the DT. I think that introduces more case distinctions on the logic in _pdf that can be called, instead of reducing their number?

Are you referencing the `BaseDistribution._pdf` where you differentiate the `cdf` to get the pdf? If so, I see your point now! It does depend on the `cdf` being `exact` - is there ever a case where we have an undefined/approx `cdf` and an exact `pdf`? Seems unlikely.

Numerically, is the derivative of the inverse_transform * pdf (for a well-defined transform) more stable than the derivative of the `cdf`? Also seems unlikely. This is more for my own interest than a reason to keep it.

fkiraly:
> I would say the issue is just that the `TransformedDistribution` uses the term `transform` to mean the backward transform: `g(y_t) -> y`; and uses the term `inverse_transform` to mean the forward transform: `f(y) -> y_t`, where `y` and `y_t` are the original target and transformed target. This is contrary to the standard sklearn nomenclature.

I think we need a synchronous discussion in one of the tech meetings to resolve this, I think we are not moving closer to a resolution here. Do 13 UTC on Mondays or Fridays suit you?

I think you are conflating forward/backward based on context. The `TransformedTargetRegressor` inverts the transformation in prediction, so it should pass the inverse of the transformation to `TransformedDistribution`. The `TransformedDistribution` itself should not invert. That is what, in my opinion, makes conceptual sense.

You seem to mentally couple the API design of `TransformedDistribution` to its use in `TransformedTargetRegressor`, and by that you are not conceptually precise where flips should take place or not, and how this maps when we add the differentiable transformer.

In particular, I do not think any flips should happen inside TD.

joshdunnlime:
Next steps:
1) Skpro TransformedTargetRegressor to add `func` and `inverse_func` kwargs to conform with sklearn TTR. `func` and `inverse_func` will be applied to an internal `FunctionTransformer`.
2) TransformedDistribution (TD) to add `inverse_func`, `transformer` and `inverse_transformer`. I also suggest deprecating `transform` in favour of `func`. This then conforms with both skpro and sklearn TTR, but in the opposite direction, as it currently is (e.g. the transform is w.r.t to distribution: forward transform `func` is f(D) -> T, backward transform `inverse_func` is f-1(T) -> D.
3) TD `func` and `inverse_func` initialise a `FunctionTransformer`, assigned to `self.transformer_`
4) TTR will initialise TD with `inverse_transformer` and will map to `self.transformer_` but with inverted `transform` and `inverse_transform`.

skpro/distributions/trafo/_transformed.py
```diff
@@ -15,25 +20,41 @@ class TransformedDistribution(BaseDistribution):
     Constructed with:
 
     * ``distribution``: a skpro distribution object, which is transformed
-    * ``transform``: a function that is applied to the distribution.
```
fkiraly, line 18: I think we should leave the `transform` argument in order to be downwards compatibility.

fkiraly (changes requested):
Nice!
 
* I think the API is a bit confused now - why do we nee `transformer` * and* `inverse_transformer`, if a transformer has both? I do not think the second, new argument makes sense, but happy to be convinced otherwise.
* I would suggest to leave only the `transform` argument for now, and allow it to be a differentiable transformer. And add potential additional arguments and features in separate PR. This is to keep complexity and number of iterations low.
* the try/except problem still needs to be addressed

joshdunnlime:
1) - The TransformedTargetRegressor (sklearn + skpro as of this PR) has the following API:
`func` and `inverse_func` as functions, `transformer` for the transformer class. The TransformedDistribution in main only has the `transform` kwarg which takes a function (not transformer) which I believe is confusing. I think it is better to not overload `transform` with functions and classes but to add `func`, `inverse_func` and `transformer` to comply with TTR. `func` would always be the forward transform w.r.t to some object `TransformedObject`, `inverse_func` the backward transform and the `transformer` class' `transform` and `inverse_transform` methods forward and backward respectively.
    - As discussed in the call, we need to do a single "flip" of the `transform` and `inverse_transform` methods when going from TTR to TD. The `inverse_transformer` kwarg makes this very explicit. It only happens in a single place and does so exactly at the point where we switch our transform reference point (switching objects for `TransformedObject`).

2) We could continue to support `transform` as an alias for `func`. This ensures back-compat and can always be deprecated in the future if we decide to. But I think adding the `func` would make things a lot clearer and conform to TTR as mention above.

3) I will switch to using `soft_deps` etc.

joshdunnlime:
An alternative to `inverse_transform` kwarg would be to add a classmethod to TD called `_from_inverse_transform`. This avoids polluting the public API and has all of the benefits listed above plus is much more explicit.

joshdunnlime:
1. "TransformedObject" pattern used across TTR and TD to conform with sklearn TTD.
2. `inverse_transformer` kwarg removed in favour of `TD._from_inverse_transformer` classmethod
3. _check_soft_dep used instead of try-except for scipy derivate

fkiraly:
> `func` and `inverse_func` as functions, `transformer` for the transformer class. The TransformedDistribution in main only has the `transform` kwarg which takes a function (not transformer) which I believe is confusing. I think it is better to not overload `transform` with functions and classes but to add `func`, `inverse_func` and `transformer` to comply with TTR.

This would be a breaking change though, because we would rename `transform` to `transformer`. I would suggest to not add other options by which to invoke TTR in this PR in order to avoid complexity and breakage, and leave changes to a future PR.

> As discussed in the call, we need to do a single "flip" of the `transform` and `inverse_transform` methods when going from TTR to TD. The `inverse_transformer` kwarg makes this very explicit. It only happens in a single place and does so exactly at the point where we switch our transform reference point (switching objects for `TransformedObject`).

Are you sure? I thought we arrived at a conclusion that no flips are needed - only adding the additional option that `transform` in TTR can be a differentiable transform, and that that differentiable functions can be used inside TD.

>We could continue to support `transform` as an alias for `func`. This ensures back-compat and can always be deprecated in the future if we decide to. But I think adding the `func` would make things a lot clearer and conform to TTR as mention above.

As discussed, I would leave renames and added args to the future, in order to limit complexity and risk for errors. The high-level principle is to modularize and compartmentalize changes where possible, and isolate breaking changes.

joshdunnlime:
> This would be a breaking change though, because we would rename transform to transformer. I would suggest to not add other options by which to invoke TTR in this PR in order to avoid complexity and breakage, and leave changes to a future PR.

No, it would just extend the `TransformedDistribution` as the `transform` kwarg only accepts a function. I propose adding `transformer` which accepts a full transformer object.

> Are you sure? I thought we arrived at a conclusion that no flips are needed - only adding the additional option that transform in TTR can be a differentiable transform, and that that differentiable functions can be used inside TD.

I followed the notes from the Meet Up:
> The computation during `fit` is
```python
TransformedTargetRegressor.fit(X, t.transform(y))
```
>the computation during `predict` is
```python
y_pred_inner = regressor.predict_proba(X_test)
y_pred_outer = TransformedDistribution(y_pred_inner, t.inverse_transform)
```

So `TransformedDistribution` uses the `inverse_transform` (w.r.t `TransformedTargetRegressor`), which is the flip.

fkiraly:
> So `TransformedDistribution` uses the `inverse_transform` (w.r.t `TransformedTargetRegressor`), which is the flip.

Yes, I see - there is currently a flip on `main` already, but what I meant, nothing else needs to get flipped compared to the current state on `main`.

> No, it would just extend the `TransformedDistribution` as the `transform` kwarg only accepts a function. I propose adding `transformer` which accepts a full transformer object.

Ok, makes sense - how do you plan though to handle the resetting when the transform is already fitted? Does it work to just pass it?

joshdunnlime:
Yes exactly. This is just that single flip being made explicit.

> Ok, makes sense - how do you plan though to handle the resetting when the transform is already fitted? Does it work to just pass it?

AFAIK, we always return a new `TransformedDistribution` on `predict_proba`, even when using `update` on an online regressor (it is the regressor that gets updated, not the TD which is always returned anew). As it stands, the `fit_with_fitted` method handles passing the fitted transformer. Your previous comment about using `FrozenEstimator` makes a lot more sense than adding a new `WrappedTransformer` (a more descriptive would be `_PrefittedEstimatorMixin`), so I plan to implement that next.

Also, I think the `DifferentiableTransformerBase` is overkill. That functionality can just live on the `DifferentiableTransformer` so I plan to simplify that.

Does this sound right to you?

fkiraly:
> Also, I think the `DifferentiableTransformerBase` is overkill. That functionality can just live on the `DifferentiableTransformer` so I plan to simplify that.

> Does this sound right to you?

I have a bit of a negative feeling about it, because I remember that maybe more transformations may have the differentiation property later on. But no strong feeling.

joshdunnlime:
The DifferentiableTransformer itself really wasn't adding much, just combining the "Wrapped" and the BaseDiff. We now just have a private method `_wrap_if_fitted`, which is much cleaner and simpler. It would be trivial to add the Base class if we need to in the future.

fkiraly:
ok - let me know if it is ready to review. A test is still failing.

joshdunnlime:
There is some very ugly behaviour in scipy:
1) scipy<=1.12.0 has the derivative function in `misc`
2) scipy~=1.13 and scipy~=1.14 has no derivative function.
3) scipy>=1.15.0 has derivative in the `differentiate` module (with a new api).

In case 2), we use the custom central difference implementation.

joshdunnlime:
Tests are now all passing. We are ready for a review @fkiraly

joshdunnlime:
@fkiraly Any more thoughts on this one? Thanks!

joshdunnlime:
@fkiraly - this should be complete now. Would love to get this merged if you're happy with it.

PR #621, opened 2025-10-14
WIP [ENH] Multivariate distribution framework support

fkiraly:
WIP PR to discuss how to support multivariate distributions.

Currently contains one additional method `pdfj` which would explain how we distinguish the current variable-marginal `pdf`, and the joint `pdf`.
The default for `pdfj` would be the row-wise product of the `pdf` output, but can be overridden for multivariate distributions.

Questions about API:

* is this a reasonable design?
* this would directly generalize to `log_pdfj`, `cdfj` etc - it feels a bit duplicative.
* another option would be to include an argument `marginal` to all the methods, and it could be `"entry"` (default and current behaviour), `"row"`, or `"none"`?

### DIFF
--- a/skpro/distributions/tests/test_all_distrs.py
+++ b/skpro/distributions/tests/test_all_distrs.py
@@ -50,12 +50,10 @@ def _has_capability(distr, method):
 
 METHODS_SCALAR = ["mean", "var", "energy"]
 METHODS_SCALAR_POS = ["var", "energy"]  # result always non-negative?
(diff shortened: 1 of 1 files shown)


fkiraly:
@joshdunnlime, I turned this branch into a PR for discussion - your opinions would be appreciated.

fkiraly:
FYI, I opened an issue to collect API discussion for multivariate distributions here:
https://github.com/sktime/skpro/issues/622

PR #630, opened 2025-11-04
[MNT] Remove <3.13 bounds from soft dependencies in all_extras

### LINKED ISSUE #478: [MNT] gradually remove `<3.13` bounds from soft dependencies in `all_extras`
We should gradually remove `<3.13` bounds from soft dependencies in `all_extras` as the soft dependencies become `python 3.13` and `numpy 2` compatible.

Here is a recipe for that:

1. in `pyproject.toml`, remove the `<3.13` marker from *a single* soft dependency
2. check whether the environment installs with `numpy 2` and without failure, locally and on the CI

repeat until either all the above are removed, or it is clear that updates to 3.13 are unlikely, for a given package.

Abdulwahid7027:
Removes the `python_version < '3.13'` constraints from the optional dependencies in the `all_extras` group of `pyproject.toml`, as per issue #478. No other version constraints were modified. Compatibility with Python 3.13 and NumPy 2 will be checked via CI.

Fixes #478.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,10 +50,10 @@ dependencies = [
 [project.optional-dependencies]
 all_extras = [
     "distfit",
-    "lifelines<0.31.0",
+    "lifelines<0.31.0; python_version < '3.14'",
     "mapie",
     "matplotlib>=3.3.2",
-    "ngboost<0.6.0",
+    "ngboost<0.6.0; python_version < '3.14'",
     "polars<1.36.0",
     "pymc<3.14",
     "statsmodels>=0.12.1",


fkiraly:
you may have to change some of the bounds to <3.14

Abdulwahid7027:
Updated the dependency bounds as suggested (`pymc<3.14`). Ready for CI.

Abdulwahid7027:
Updated the optional dependencies to restrict only lifelines and ngboost on Python < 3.14, as requested. Other <3.13 markers remain removed. Ready for CI.

Issue #369, opened 2024-05-28
[ENH] `sample_weight` for metrics


fkiraly:
A `sample_weight` parameters should be added to metrics.

Where it is not clear what to do, the following default can be used:

* compute jackknife pseudo-samples
* weight the pseudo-samples
* average the weighted pseudo-samples

For consideration, a tag is perhaps needed here.

PR #631, opened 2025-11-16
[ENH] Added sample_weight parameter

ParamThakkar123:
#### Reference Issues/PRs

Fixes https://github.com/sktime/skpro/issues/369

This PR adds a new `sample_weight` parameter to metrics as described in #369 

#### Any other comments?

I am open to suggestions to change this PR according to the reviewers.

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/metrics/base.py
+++ b/skpro/metrics/base.py
@@ -78,7 +78,7 @@ class BaseProbaMetric(BaseObject):
         """
         return self.evaluate(y_true, y_pred, **kwargs)
 
-    def evaluate(self, y_true, y_pred, **kwargs):
+    def evaluate(self, y_true, y_pred, sample_weight=None, **kwargs):
         """Evaluate the metric on given inputs.
 
         Parameters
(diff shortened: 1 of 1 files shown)


skpro/metrics/base.py
```diff
@@ -254,14 +267,28 @@ def _evaluate_by_index(self, y_true, y_pred, **kwargs):
...
-            x_bar = self.evaluate(y_true, y_pred, self.multioutput, **kwargs)
+            x_bar = self.evaluate(y_true, y_pred, sample_weight=sample_weight, **kwargs)
+
             for i in range(n):
-                out_series[i] = n * x_bar - (n - 1) * self.evaluate(
-                    np.vstack((y_true[:i, :], y_true[i + 1 :, :])),  # noqa
-                    np.vstack((y_pred[:i, :], y_pred[i + 1 :, :])),  # noqa
-                    self.multioutput,
-                    **kwargs,
+                if isinstance(y_true, np.ndarray):
```
fkiraly, line 273: could you explain why the case distinction based on `if isinstance(y_true, np.ndarray)`? How does introducing `sample_weight` change this?

ParamThakkar123: The `isinstance(y_true, np.ndarray)` branch exists because the code removes the i-th sample using NumPy slicing/stacking which only works reliably for plain NumPy arrays. Adding `sample_weight` requires removing the same sample from the weight array/Series with the same method (np.delete/mask for NumPy or .iloc/.drop for pandas) to keep alignment, otherwise metrics might become incorrect?

fkiraly: * what are the two cases that this tries to cover, exactly?
* why do we not need to do this before your change?

PR #635, opened 2025-11-19
[ENH] Inverse Gaussian distribution

Omswastik-11:
### Summary
This PR adds the `InverseGaussian` distribution class to `skpro.distributions`. This implementation wraps `scipy.stats.invgauss` while ensuring proper parameter mapping to match standard parameterizations (specifically `numpy.random.wald`).

### Key Features & Formulation
* **Parameterization:** Uses Mean $\mu$ and shape parameter $\lambda$ (scale).
* **Formula:**
    $$f(x; \nu, \lambda) = \sqrt{\frac{\lambda}{2\pi x^3}} \exp\left(-\frac{\lambda(x-\nu)^2}{2\nu^2 x}\right)$$
* **Correspondence:** Equivalent to `numpy.random.wald(mean=mu, scale=λ)`.
* **Broadcasting:** Full support for array-valued parameters.

### Implementation Details
* **Backend:** Wraps `scipy.stats.invgauss` with specific parameter mapping:
    * `mu = mu`
    * `scale = λ`
* **Capabilities:** Implements exact mean, variance, PDF, CDF, and PPF. Includes approximate energy and PDF normalization.
* **Files Modified:**
    * `inversegaussian.py`: Added new distribution class.
    * `__init__.py`: Exported the new class.

### DIFF
--- a/skpro/distributions/inversegaussian.py
+++ b/skpro/distributions/inversegaussian.py
@@ -20,14 +20,10 @@ class InverseGaussian(_ScipyAdapter):
     parameters of ``scipy.stats.invgauss(mu, scale=scale)``. The
     mean of the distribution is given by ``mean = mu * scale``.
 
-    The common shape-scale parameterization with mean :math:`\nu` and
-    shape :math:`\lambda` has pdf:
+    The pdf in terms of :math:`\mu` = ``mu`` and :math:`\sigma` = ``scale`` is:
 
-    .. math:: f(x; \nu, \lambda) = \sqrt{\frac{\lambda}{2 \pi x^3}}
-              \exp\left(-\frac{\lambda (x - \nu)^2}{2 \nu^2 x}\right)
-
-    This is equivalent to SciPy's parameterization with
-    ``mu = nu / lam`` and ``scale = lam``.
+    .. math:: f(x; \mu, \sigma) = \sqrt{\frac{\sigma}{2 \pi x^3}}
+              \exp\left(-\frac{(x - \mu \sigma)^2}{2 \mu^2 \sigma x}\right)
 
     Parameters
     ----------


fkiraly:
Code quality checks are failing, see https://www.sktime.net/en/stable/developer_guide/coding_standards.html

Omswastik-11:
Hi @fkiraly !! Corrected the code quality checks .

Omswastik-11:
image


I have implemented the requested changes for `InverseGaussian` in `inversegaussian.py`.

### Summary of Changes

* **Parameter Renaming**: Renamed `scale` $\to$ `lambda_`.
    * *Rationale*: Addressed the feedback that this is not a scale parameter. I kept `mu` as `mu` because `mean` conflicts with the inherited `.mean()` method, and `shape` conflicts with the `.shape` property. `lambda_` is used to represent $\lambda$ while avoiding Python keyword conflicts.
* **Documentation**:
    * Updated docstrings to use double backticks .
    * Updated descriptions, examples, and Scipy/Numpy mappings to reflect the change to `lambda_`.
* **Implementation**: Updated `__init__`, `_get_scipy_param`, and `get_test_params`.

**Verification**: Ran `pytest` for InverseGaussian and all tests passed successfully.

image

image

Omswastik-11:
Then Should I change it to 'Scale' ? because `shape` is creating error . I think to use ` lam` instead as `scipy`
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.invgauss.html

image

fkiraly:
Yes, I would say, use the `numpy` convention, which has `scale`.

Omswastik-11:
hi @fkiraly !! renamed `lam` to `scale` .

fkiraly (changes requested):
Great! Looks fine now regarding code, but I think the mathematical formula is now incorrect. If `scale` is a multiplicative scale parameter, then you should divide and not multiply by `scale`.

While you are at it, I would use greek symbols $\mu$ and $\sigma$ in the formula rather than writing out "scale" etc, that makes it easier to read.

Omswastik-11:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.invgauss.html

This time I used Scipy's formula

fkiraly (changes requested):
Thanks, but the formula is not directly applicable. Could you not just give a formula in terms of a $\mu$ = `mu` and $\sigma$ = `scale`?

Omswastik-11:
I think It would look like this if we transform scipy's eqn

image

PR #637, opened 2025-11-20
[ENH] interface ondil OnlineGamlss wrapper

### LINKED ISSUE #470: [ENH] interface `rolch` online proba regressors
It would be nice to interface the `rolch` package:
https://github.com/simon-hirsch/rolch

The new online update feature should e useful to cover the update feature of the alogrithm.

arnavk23:
#### Reference Issues/PRs
Fixes #470

Adds support for `ondil` online probabilistic regressors, enabling `skpro` users to use `OnlineGamlss` with the standard skpro API and tooling.
Keeps `ondil` optional and minimizes maintenance surface by lazy imports and best-effort API adaptation.

##### For all contributions

##### For new estimators

### DIFF
--- a/extension_templates/distributions.py
+++ b/extension_templates/distributions.py
@@ -122,7 +122,7 @@ class ClassName(BaseDistribution):
     # if not implemented, the base class will try to fill it in
     # from the other implemented methods
     # at least _ppf, or sample should be implemented for the distribution to be usable
-    # if _ppf is implemented, sample does not need to be implemented (uses ppf sampling)
+    # if _ppf is implemented, _sample does not need to be (uses ppf sampling)
 
     # todo: consider implementing
     # if not implemented, uses Monte Carlo estimate via sample


skpro/regression/rolch.py
```diff
@@ -0,0 +1,222 @@
...
+    Notes
+    -----
+    * The rolch dependency is optional and imported inside methods.
+    * The wrapper uses a best-effort strategy to call the appropriate
+      fit/update/predict methods of the upstream estimator. If rolch's
+      API changes in incompatible ways, this wrapper may need updates.
+    """
+
+    _tags = {
+        "authors": ["fkiraly"],
```
fkiraly: the author is not me but you

skpro/regression/ondil.py
```diff
@@ -0,0 +1,222 @@
...
+      fit/update/predict methods of the upstream estimator. If rolch's
+      API changes in incompatible ways, this wrapper may need updates.
+    """
+
+    _tags = {
+        "authors": ["fkiraly"],
+        "maintainers": ["fkiraly"],
+        "python_dependencies": ["rolch"],
+        "capability:multioutput": False,
+        "capability:missing": True,
```
fkiraly, line 44: the estimator should have `capability:update`

fkiraly (changes requested):
Thanks! I think the package is no longer called `rolch` but `ondil`

fkiraly:
code quality checks are failing, see https://www.sktime.net/en/stable/developer_guide/coding_standards.html

arnavk23:
@fkiraly made changes according to your review comments.

arnavk23:
@fkiraly Fixed the failing lint checks(isort). Please run the workflows again.

skpro/regression/ondil.py
```diff
@@ -0,0 +1,201 @@
...
+
+    _tags = {
+        "authors": ["arnavk23"],
+        "maintainers": ["fkiraly"],
+        "python_dependencies": ["ondil"],
+        "capability:multioutput": False,
+        "capability:missing": True,
+        "capability:update": True,
+        "X_inner_mtype": "pd_DataFrame_Table",
+        "y_inner_mtype": "pd_DataFrame_Table",
```
fkiraly, line 48: please add the tag `"tests:vm": True`, or tests will not run

skpro/regression/ondil.py
```diff
@@ -0,0 +1,201 @@
...
+
+        The method tries several common fitting/update method names in the
+        upstream estimator to support different ondil versions.
+        """
+        # defer import to keep ondil optional
+        import importlib
+
+        module_str = "ondil.estimators.online_gamlss"
+        ondil_mod = importlib.import_module(module_str)
+        OnlineGamlss = ondil_mod.OnlineGamlss
```
joshdunnlime: I am getting:
```AttributeError: module 'ondil.estimators.online_gamlss' has no attribute 'OnlineGamlss'```
so I switched to:
```OnlineGamlss = ondil_mod.OnlineDistributionalRegression```

File: skpro/regression/ondil.py [hunk 5]
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    def __init__(self, distribution="Normal", **ondil_kwargs):
+        self.distribution = distribution
+        # store any kwargs forwarded to the ondil OnlineGamlss constructor
+        self._ondil_kwargs = ondil_kwargs
+
+        super().__init__()
+
+    def _fit(self, X, y):
```
joshdunnlime, line 68: Add `(..., C=None):` to override the baseclass method correctly.

fkiraly: actually, `C` is not necessary if the regressor does not support censoring. The base class is smart enough :-)

skpro/regression/ondil.py
arnavk23: Added a fallback there, please see.

skpro/regression/ondil.py
joshdunnlime: Good to know. If including `C=None` isn't too confusing, it might be worth adding as it raises a lot of false-positives in my local type-checker and may do so for other users.

skpro/regression/ondil.py
```diff
@@ -0,0 +1,212 @@
...
+
+            Normal = distr_mod.Normal
+            return Normal(mu=loc, sigma=scale, index=X.index, columns=self._y_cols)
+
+        # fallback: try to call distribution class with all columns as kwargs
+        if hasattr(distr_mod, dist):
+            Distr = getattr(distr_mod, dist)
+            # construct args dict using column names
+            vals = {c: df.loc[:, [c]].values for c in df.columns}
+            return Distr(**vals, index=X.index, columns=self._y_cols)
```
joshdunnlime, line 238: Looks like there is an issue here is the column names are not strings. In my case, I am transforming `X` and integer column values are used. Therefore some of the keys in `vals` are ints.

`TypeError: keywords must be strings`

skpro/regression/ondil.py
fkiraly: ah, I see.

FYI, unfortunately this is not possible since originally censoring was not supported (only ordinary probabilistic supervised regression), so a lot of third party code would use `_fit(X, y)` as signature. Requiring `C` suddenly would break all this code, including all estimators then present in `skpro` itself.

Hence the more lenient extension contract - which also means, the same problem would be caused when starting to require `C` at any point in time, breaking custom extender code.

fkiraly: From a power user perspective I think this is also a good compromise solution for the long term - since users not interested in censoring do not need to worry about what that `C` is.

arnavk23:
@fkiraly please review this pr further.

skpro/regression/ondil.py
joshdunnlime: I just meant just passing `C=None` and nothing else - it keeps the linters happy and I've noticed it elsewhere in the codebase. If it's not advised then please ignore my comment.

skpro/regression/tests/test_ondil.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+try:
```
fkiraly: import `ondil` only in the test, nad add a `pytest.mark.skip` decorator to it, like in other tests dependent on soft depenencies. Search for use of `pytest.mark.skip` in the repo to see examples.

fkiraly:
@arnavk23, please correct the code formatting errors.
See here for the guide: https://www.sktime.net/en/stable/developer_guide/coding_standards.html

arnavk23:
@fkiraly These pre-commit failures are unrelated to my changes (they're in distributions.py and _base.py).

extension_templates/distributions.py
```diff
@@ -122,7 +122,8 @@ def __init__(self, param1, param2="param2default", index=None, columns=None):
     # if not implemented, the base class will try to fill it in
     # from the other implemented methods
     # at least _ppf, or sample should be implemented for the distribution to be usable
-    # if _ppf is implemented, _sample does not need to be implemented (uses ppf sampling)
+    # if _ppf is implemented, sample does not need to be implemented
```
fkiraly: why are you changing this line? seems unrelated

fkiraly: reverted

PR #639, opened 2025-11-20
[ENH] LightGBMLSS interface

### LINKED ISSUE #589: [ENH] Add LightGBMLSS to skpro
This would closely follow the work done on [XGBoostLSS](https://github.com/sktime/skpro/blob/v2.9.2/skpro/regression/xgboostlss.py#L6-L321). Although the models are quite similar, XGBoost and LightGBM both have unique benefits such that it would be worth having both of these as an option.

```python
from skpro.regression import lightgbmlss

lgblss = lightgbmlss.LightGBMLSS()
```

yuvimittal:
#### Reference Issues/PRs

Fixes #589 

Adds interface for lightgbmlss probabilistic regressor

##### For all contributions

##### For new estimators


skpro/regression/lightgbmlss.py
```diff
@@ -0,0 +1,355 @@
...
+            # LightGBMLSS hyper_opt returns "opt_rounds" key for best iteration
+            n_rounds = opt_params.pop("opt_rounds", self.num_boost_round)
+
+        # Train Model with optimized hyperparameters
+        lgblss.train(opt_params, dtrain, num_boost_round=n_rounds)
+
+        self.lgblss_ = lgblss
+        return self
+
+    def _hyper_opt(self, lgblss, dtrain):
```
fkiraly: this method is not required, or used - I would suggest, just remove it?

fkiraly:
code quality checks are failing, see https://www.sktime.net/en/stable/developer_guide/coding_standards.html

yuvimittal:
> code quality checks are failing, see https://www.sktime.net/en/stable/developer_guide/coding_standards.html

@fkiraly , i will look into it. Thank you

skpro/regression/lightgbmlss.py
yuvimittal: `_hyper_opt` is actually being used inside` _fit` when `n_trials > 0`:

`opt_params = self._hyper_opt(lgblss, dtrain)`

If the intention is to remove custom hyperparameter code and rely directly on lightgbmlss.hyper_opt, I can refactor accordingly.
Just let me know if you prefer removing _hyper_opt and calling lgblss.hyper_opt inline in _fit, or keeping the helper method for clarity (since XGBoostLSS uses the same pattern).

fkiraly:
> `_hyper_opt` is actually being used inside` _fit` when `n_trials > 0`:

Ah, I see - you are right.

In this case, I believe this feature requires `optuna`?

In the `__init__`, then, `set_tag` should be used to set `python_dependencies` including `optuna`, if `n_trials>0`.

docs/source/api_reference/regression.rst
```diff
@@ -145,6 +145,16 @@ Distributional boosting
 
     XGBoostLSS
 
+-----------------------
```
fkiraly: this is an empty header. Remove.

skpro/regression/lightgbmlss.py
```diff
@@ -0,0 +1,354 @@
...
+    n_trials: int, optional, default=30
+        The number of trials in tuning.
+        If this argument is set to None, there is no limitation on the number of trials.
+        If set to 0, no tuning is done, and default parameters of LightGBMLSS are used.
+    """
+
+    _tags = {
+        # packaging info
+        # --------------
+        "authors": ["StatMixedML", "EchoDel", "fkiraly"],
```
fkiraly: I do not think I am an author of this, or lightgbmlss. Add yourself instead of myself as author.

skpro/regression/lightgbmlss.py
fkiraly: I see, I made a mistake since I did not see the call. So my request is not useful.

yuvimittal:
> Nice! Looks like it should work.
> 
> Two comments:
> 
> * the estimator seems to be failing tests - is this stochastic? FYI @StatMixedML
> * this looks largely identical to the `xgboostlss` interface - good that you recognized this! Of course copy-pasting is a good starting idea, but I think we should not leave it with duplicate code, due to the DRY principle https://en.wikipedia.org/wiki/Don%27t_repeat_yourself. I would move methods or pieces of code that are common to both estimators into a new mixin class in a new module `regression.adapters._statmixedml` (if this looks reasonably modularizable)

Thanks for the review!

• I’ve introduced a new shared module `regression.adapters._statmixedml` and refactored the overlapping logic from both `XGBoostLSS` and `LightGBMLSS`  into it. 

•Regarding the failing CI tests, I'm trying to reproduce them locally. The CI output shows that `test_fit_does_not_overwrite_hyper_params` is failing for `LightGBMLSS`, but this test is being skipped locally because `LightGBMLSS` is not picked up during local test discovery. Could you please let me know the recommended way to run the CI-equivalent test suite locally so I can reproduce the same failure before pushing further commits?

fkiraly:
> but this test is being skipped locally because `LightGBMLSS` is not picked up during local test discovery.

It should get picked up, can you share how you are trying to run, or how you are discovering tests?

> Could you please let me know the recommended way to run the CI-equivalent test suite locally so I can reproduce the same failure before pushing further commits?

The easiest is to run `check_estimator` from `skpro.utils` on your new estimator.

yuvimittal:
> > but this test is being skipped locally because `LightGBMLSS` is not picked up during local test discovery.
> 
> It should get picked up, can you share how you are trying to run, or how you are discovering tests?
> 
> > Could you please let me know the recommended way to run the CI-equivalent test suite locally so I can reproduce the same failure before pushing further commits?
> 
> The easiest is to run `check_estimator` from `skpro.utils` on your new estimator.

Initially i was looking at the failed tests from CI and running them, but now i ran check_estimator on both xgboostlss and lightgbmlss, xgboostlss is running perfectly but i am unable to install lightgbm

this is the error i am getting :

Screenshot 2025-11-23 at 2 03 21 AM

fkiraly:
I think the package is available only for up to python 3.11 - so you nay want to set the `python_dependencies` tag to exclude the failing versions.

StatMixedML:
The original LightGBMLSS is now aligned with @fkiraly's implementation for XGBoostLSS. just wanted to inform you.

marrov:
@yuvimittal - Are you still working on this PR? Otherwise we can assign to someone else.

PR #643, opened 2025-11-22
[BUG] `XGBoostLSS` dynamic dependencies: add optuna to `python_dependencies` when `n_trials != 0`

ravjot07:
## Summary

This PR fixes an issue where `XGBoostLSS` requires `optuna` when `n_trials != 0` for hyperparameter optimization, but the dependency was not properly declared in the `python_dependencies` tag.

Depends on #681 for testing.

## Changes

- **Modified `XGBoostLSS.__init__`**: Added logic to dynamically include `optuna` in `python_dependencies` when `n_trials != 0`
## Problem

When `n_trials` is set to a non-zero value, `XGBoostLSS` uses hyperparameter optimization via the `_hyper_opt` method, which requires `optuna`. However, the `python_dependencies` tag only included `xgboostlss`, causing dependency checking to fail when `optuna` was not installed.

## Solution

The fix dynamically updates the `python_dependencies` tag at the end of `__init__`:
- When `n_trials == 0`: Only `xgboostlss` is required (no hyperparameter optimization)
- When `n_trials != 0`: Both `xgboostlss` and `optuna` are required (hyperparameter optimization enabled)

This ensures the dependency checking system properly validates and installs `optuna` when needed, without removing test coverage.

### DIFF
--- a/skpro/regression/xgboostlss.py
+++ b/skpro/regression/xgboostlss.py
@@ -343,16 +343,17 @@ class XGBoostLSS(BaseProbaRegressor):
             `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
             `create_test_instance` uses the first (or only) dictionary in `params`
         """
-        params0 = {"max_minutes": 1}
+        params0 = {"max_minutes": 1, "n_trials": 2}
         params1 = {
             "stabilization": "L2",
             "loss_fn": "crps",
             "max_minutes": 1,
+            "n_trials": 2,
         }
-        params2 = {"dist": "Gamma", "max_minutes": 1}
-        params3 = {"dist": "Weibull", "max_minutes": 1}
-        params4 = {"dist": "TDistribution", "max_minutes": 1}
-        params5 = {"dist": "Laplace", "max_minutes": 1}
+        params2 = {"dist": "Gamma", "max_minutes": 1, "n_trials": 2}
+        params3 = {"dist": "Weibull", "max_minutes": 1, "n_trials": 2}
+        params4 = {"dist": "TDistribution", "max_minutes": 1, "n_trials": 2}
+        params5 = {"dist": "Laplace", "max_minutes": 1, "n_trials": 2}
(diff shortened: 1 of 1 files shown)


skpro/regression/xgboostlss.py
```diff
@@ -115,6 +115,21 @@ def __init__(
         else:
             self._n_cpu = n_cpu
 
+        # If n_trials is not zero, optuna is required for hyperparameter optimization
+        if n_trials != 0:
+            current_deps = self.get_tag("python_dependencies")
```
fkiraly: too complicated - just `set_tag` to the list containing two strings.

fkiraly (changes requested):
Thanks!

Kindly simplify, see above.

skpro/regression/xgboostlss.py
ravjot07: done @fkiraly

fkiraly:
the failure comes from an oversight in VM testing that does not take into account that more dependencies may need to be installed than in the class tag. I have to think about how to resolve this, but it is not an issue for your PR.

skpro/regression/xgboostlss.py
```diff
@@ -115,6 +115,10 @@ def __init__(
         else:
             self._n_cpu = n_cpu
 
+        # If n_trials is not zero, optuna is required for hyperparameter optimization
+        if n_trials != 0:
+            self.set_tags(**{"python_dependencies": ["xgboostlss", "optuna"]})
```
joshdunnlime, line 121: This probably doesn't cover enough of the dependencies. we also need optuna-integration:
https://github.com/StatMixedML/XGBoostLSS/blob/931750e765d1ce6a22200316891b0f918f5e8abe/xgboostlss/model.py#L329

If this is valid syntax we should be using: `"python_dependencies": ["xgboostlss[all_extras"]`

fkiraly: added

PR #648, opened 2025-11-24
[ENH] zero-inflated distribution

Khushmagrawal:
#### Reference Issues/PRs
Partially towards #554 
~Blocked by #667~

Adds support for Zero-inflated distribution

No.

- Correctness of the implementation
- ~Design decision of making the lower bound inclusive in `truncated.py`~

No
#### Any other comments?
After the ZeroInflated implementation is reviewed, I plan to add `ZeroInflatedPoisson` and `ZeroInflatedNegativeBinomial` as well using `_DelegatedDistribution`

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/zeroinflated.py
+++ b/skpro/distributions/zeroinflated.py
@@ -1,6 +1,4 @@
-"""Zero-Inflated distribution implementation."""
-
-__author__ = ["khushmagrawal"]
+"""Zero-Inflated distribution."""
 
 import numpy as np
 from numpy.typing import ArrayLike
@@ -9,7 +7,7 @@ from skpro.distributions.base import BaseDistribution
 
 
 class ZeroInflated(BaseDistribution):
-    r"""A Zero-Inflated distribution.
+    r"""A Zero-Inflated distribution, adding a Bernoulli gate to any distribution.
 
     Combines a Bernoulli gate for structural zeros with a with a base distribution
     defined on non-negative values, allowing zeros to arise from both components.
@@ -43,6 +41,11 @@ class ZeroInflated(BaseDistribution):
     """
 
     _tags = {
+        # packaging info
+        # --------------
+        "authors": ["khushmagrawal"],
+        # estimator tags
+        # --------------
         "capabilities:approx": ["energy"],
         "capabilities:exact": ["ppf", "mean", "var", "log_pmf", "pmf", "cdf"],
         "distr:measuretype": "mixed",


skpro/distributions/zeroinflated.py
```diff
@@ -0,0 +1,281 @@
...
+from numpy.typing import ArrayLike
+
+from skpro.distributions.base import BaseDistribution
+from skpro.distributions.truncated import TruncatedDistribution
+
+
+class ZeroInflated(BaseDistribution):
+    r"""A Zero-Inflated distribution.
+
+    Combines a Bernoulli gate for structural zeros with a with a standard count
```
fkiraly: But `distribution` does not need to be a count distribution, does it?

Khushmagrawal: right, it can be any distribution defined on non negative values (including continuous), now fixed 
Thanks

Khushmagrawal:
Fixed the off-by-one errors causing `test_ppf_and_cdf` to fail. The issue was due to floating-point error when `q_rescaled` is divided by `p`. Resolved by subtracting a small epsilon from `q_rescaled` (similar to #568). 
The workflow [passes](https://github.com/Khushmagrawal/skpro/actions/runs/19740436912) on my fork with this change

fkiraly (changes requested):
very nice!

Could we please move the modifications to `TruncatedDistribution` to another PR, since I think there is some discussion to be had about how to parameterize the options, e.g., why only lower inclusive and not upper inclusive, etc.

The zero inflated distribution looks good.

Khushmagrawal:
This can now be re-reviewed as #667 is merged

Khushmagrawal:
@fkiraly, just bringing this to your attention this needs a review. Thanks!

skpro/distributions/zeroinflated.py
```diff
@@ -0,0 +1,296 @@
...
+
+        inner_paramtype = distribution.get_tag("distr:paramtype", "parametric")
+        if inner_paramtype != "parametric":
+            self.set_tags(**{"distr:paramtype": inner_paramtype})
+
+    @property
+    def _truncated_distribution(self) -> TruncatedDistribution:
+        # TODO: return self.distribution if it is naturally non-negative.
+        # requires support tag (check #244)
+        if (
```
fkiraly: this is unintuitive behaviour, in a case where `distribution` already is a `TruncatedDistribution, with a `lower` of 1, say. IT gets replaced by double wrapped truncation.

I would suggest to consider not manually truncating, but merely suggesting to the user that `distribution` is non-negative.
Then we do not need to take care of too many modificiation cases that are too hard to explain.

skpro/distributions/zeroinflated.py
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    def __init__(
+        self,
+        p: ArrayLike,
+        distribution: BaseDistribution,
+        index=None,
+        columns=None,
+    ):
+        if isinstance(p, np.ndarray) and p.ndim == 1:
```
fkiraly: I would advise to make `p` broadcastable, by using the `broadcast` tags. Then you do not need to handle it manually.

fkiraly (changes requested):
Very nice!

Two main comments above:

* make `p` use the broadcast logic built into the base class, instead of replacing that with a different logic. That way it is consistent.
* I would also avoid truncating, actually. I would treat the distribution as a special mixture, instead, this also reduces hard to track case distinctions.

skpro/distributions/zeroinflated.py
Khushmagrawal: yes right, but actually manual truncation here was mirrored from the current  `Hurdle` implementation to ensure the base distribution is mathematically consistent
To fix the double-wrapping, we could change the check to `lower >= 0`  or as you suggested should we remove manual truncation entirely, if we do that , should we update `Hurdle` as well for consistency in maybe another PR?

skpro/distributions/zeroinflated.py
Khushmagrawal: okay, I will update the code for this

Khushmagrawal:
Thanks for the review, and apologies for the delayed reply. I can work on addressing the comments however as #713 works on the same issue., should I continue working on this PR?

fkiraly:
@Khushmagrawal, yes, please continue. I think there are useful things that both PR have that the other does not have. In either case you would be credited as a contributor.

marrov:
@Khushmagrawal - Are you still actively working on this? If so, I'm ok with merging the generic ZI distribution work. I can focus on the bespoke ZI distributions for the most used distributions.

Khushmagrawal:
> @Khushmagrawal - Are you still actively working on this? If so, I'm ok with merging the generic ZI distribution work. I can focus on the bespoke ZI distributions for the most used distributions.

Ah yes, I’m still working on this. I was waiting to reach a conclusion on the truncation discussion [above](https://github.com/sktime/skpro/pull/648#discussion_r2659671636). For now,  I have added a commit removing the truncation part, if the current code looks good then we can proceed with merging

marrov:
@Khushmagrawal - By "merging" in me previous message I meant to say combine, as I've also worked a bit on this myself. I'll try to get around to compare/review it soon.

skpro/distributions/zeroinflated.py
fkiraly: I would say, another PR.

PR #652, opened 2025-11-25
[BUG] Handle Hurdle values less than zero properly

### LINKED ISSUE #651: [BUG] Hurdle CDF
**Describe the bug**

The CDF of the Hurdle distribution should return 0 for values < 0, currently returns 1-p.

**To Reproduce**

E.g. 

```python
from skpro.distributions import Hurdle, Normal

dist = Normal(mu=0.0, sigma=1.0)
hurdle = Hurdle(0.5, dist)

samples = -1.0

p = hurdle.cdf(samples)
```

**Expected behavior**

Should be zero for all values less than 0.

**Environment**

Not nocessary for skpro <= 2.10

**Additional context**

tingiskhan:
#### Reference Issues/PRs

Fixes #651 

Adds a wrapper for the methods s.t. values outside of support are set to 0 and -np.infty.

No.

Yes.

#### Any other comments?

No,

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/tests/test_hurdle.py
+++ b/skpro/distributions/tests/test_hurdle.py
@@ -1,5 +1,4 @@
 import numpy as np
-import numpy.testing as npt
 import pytest
 
 from skpro.distributions.hurdle import Hurdle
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks!

Could we avoid decorators to ensure the extension contract has a fixed signature?
That would also mean having a "three casess" distinction in `_cdf` rather than two cases only.

We should also add a small regression test.

tingiskhan:
> Thanks!
> 
> Could we avoid decorators to ensure the extension contract has a fixed signature? That would also mean having a "three casess" distinction in `_cdf` rather than two cases only.
> 
> We should also add a small regression test.

Absolutely, no worries. Would you be fine with just removing the decorator and re-using the inner-function? I.e.

```python
    def _set_to_constant_where_negative(x: np.ndarray, const: float) -> np.ndarray:
        return np.where(x < 0, const, result)

    def _log_pmf(self, x):
        log_prob_zero = np.log(1.0 - self.p)
        log_prob_hurdle = np.log(self.p)

        is_zero = x == 0
        result = np.where(is_zero, log_prob_zero, log_prob_positive)

        return _set_to_constant_where_negative(result, const=-np.inf)
```

fkiraly:
> Would you be fine with just removing the decorator and re-using the inner-function? I.e.

Yes, sure - as long as it is vectorized (which it is)

tingiskhan:
@fkiraly, this is something I must have missed when first working on skpro, but are vectorized operations for ~~univariate~~ zero-dimensional distributions not supported? I can e.g. not run the below snippet of code without getting an exception

```python
import numpy as np
import pandas as pd

dist = Normal(0, 1)
xs = -1 + np.arange(0, 10)

pdf_vals = dist.pdf(xs)
log_pdf_vals = dist.log_pdf(xs)
cdf_vals = dist.cdf(xs)
```
And wondering if this is to be expected, or something wrong with my environment.

fkiraly:
this is expected, currently only broadcasting arguments up is supported, not broadcasting the distributions up.

Feels like a great idea to add, would you like to open an issue? Alghough I feel this might leed to too many complications if the broadcasting logic is not reworked cleanly to rely on `numpy`.

fkiraly (approved):
LGTM now!

tingiskhan:
> this is expected, currently only broadcasting arguments up is supported, not broadcasting the distributions up.
> 
> Feels like a great idea to add, would you like to open an issue? Alghough I feel this might leed to too many complications if the broadcasting logic is not reworked cleanly to rely on `numpy`.

Sure thing!

Yes, it might becoma a bit nasty tbh. I could open an issue and then we could discuss how a possible solution would look. This question sort of goes hand-in-hand with the one relating to distribution support if we go the same route as pytorch/pyro with event and batch shapes.

fkiraly:
yes, issue would be much appreciated, @tingiskhan! I will merge this now then.

Issue #653, opened 2025-11-25
[ENH] Conform on `distribution` or accept multiple dist* key options


joshdunnlime:
**Is your feature request related to a problem? Please describe.**

There is an incnosistent input kwarg across several models for the distribution input kwarg. Here are some:
- XGBoostLSS, NGBoost (+ LightGBMLSS PR): [dist](https://github.com/sktime/skpro/blob/main/skpro/regression/xgboostlss.py#L87)
- Residual: [distr_type](https://github.com/sktime/skpro/blob/main/skpro/regression/residual.py#L143)
- CyclicBoosting: [dist_type](https://github.com/sktime/skpro/blob/main/skpro/regression/cyclic_boosting.py#L86)
- Ondil and pyGAM (Both PRs): [distribution](https://github.com/sktime/skpro/pull/637/files#diff-bace3987d7de384258ca0f9a91f3781ba147506f31c739b1aec7e2b53b4d97c3R51)
- GLMRegressor: [family](https://github.com/sktime/skpro/blob/main/skpro/regression/linear/_glm.py#L25)

**Describe the solution you'd like**
Either:
1) Conform on a single standard - choose one of `dist`, `distribution` or `dist_type` and coerce the chosen kwarg to the package specific kwarg. (most user friendly for skpro users)
OR
2) Allow all options: `dist`, `distribution` or `dist_type` etc (most user friendly for people coming from the original package).
OR
3) Choose an skpro default, e.g. `distribution` but allow package specific types (most backward compat and with "best of both" worlds approach).

**Describe alternatives you've considered**
Do nothing - which is pretty messy. With so many new PRs for new models with differing APIs, now is a good time to act.

**Additional context**

Not changing this means it is very difficult to just switch in a new model, e.g. try XGBoostLSS in place of Cyclic. It also prevent something like the [SupportTransformer](https://github.com/sktime/skpro/issues/588) as the kwarg name changes for each regressor.

**Edit**

Added NGBoost and GLMRegressor

Aymuos22:
Hi @joshdunnlime 
[...]

PR #654, opened 2025-11-26
Issue resolved

Aymuos22:
Made distribution the single source of truth for probabilistic regressors: added a helper to reconcile the old aliases, rewired CyclicBoosting, ResidualDouble, and XGBoostLSS to use it, updated docs/changelog, and added regression tests to ensure aliases still work and conflicting values raise a clear error.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -16,6 +16,17 @@ For planned changes and upcoming releases, see roadmap in the
 `issue tracker <https://github.com/sktime/skpro/issues>`_.
 
 
+Unreleased
+==========
+
+Enhancements
+~~~~~~~~~~~~
+
+* [ENH] harmonize the ``distribution`` keyword across probabilistic regressors
+  (``ResidualDouble``, ``CyclicBoosting``, ``XGBoostLSS``) to support consistent
+  aliases such as ``dist_type`` and ``distr_type`` across models.
+
+
 [2.10.0] - 2025-10-09
 =====================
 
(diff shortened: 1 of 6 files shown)


joshdunnlime:
Although this is nice solution for point [2)](https://github.com/sktime/skpro/issues/653#issue-3664838495)
> Allow all options: dist, distribution or dist_type etc

it adds a lot of boilerplate so I would say it needs some refactoring. Perhaps a baseclass/mixin e.g. `DefinedParametricDist` that handles these getters and setters. These seems like a lot of complexity for the small added gain though.

That suggests option [3](https://github.com/sktime/skpro/issues/653#issue-3664838495)
> Choose an skpro default, e.g. distribution but allow package specific types

with `dist` or `distr` as the default might be a better solution. It is also favoured by @fkiraly in #653. It will be much more lightweight by just checking if the extra-package dist-like parameter is non-None.

Here is a simple example:

```python
    def __init__(
        distribution="normal",
        ...
        dist=None,
    ):
    """
    ...
    distribution/dist : ...
    """
    self.distribution = distribution
    self.dist = dist

    def _fit():
        self.distribution_ = self.dist or self.distribution
```

We would also need to add deprecation warnings as Franz mentioned.

Aymuos22:
Hi @joshdunnlime 
Update:
got the context;Working on the issue

joshdunnlime:
@Aymuos22 - Maybe also change the PR to a bit more of a descriptive name.

fkiraly:
Here is the deprecation guide: I would say, let's use the "changing parameter name" template:
https://www.sktime.net/en/stable/developer_guide/deprecation.html

Aymuos22:
Hi @joshdunnlime @fkiraly 
I am very new to open-source

I'll do the needful
Thanks for helping

PR #657, opened 2025-11-26
[ENH] JohnsonQPDRegressor for quantile to probability conversion

### LINKED ISSUE #377: [ENH] general Johnson QPD regression wrapper to reduce proba to quantile predictions
The Johnson QPD allow to turn quantile predictions into proba predictions.

We should add a wrapper around sklearn API copmliant methods for quantile predictions.

RakshithaKowlikar:
This PR adds JohnsonQPDRegressor, a wrapper that transforms any sklearn-style quantile regressor into a probabilistic regressor using the Johnson Quantile-Parameterized Distribution (QPD).

Fixes #377

- Fits three quantile models at (α, 0.5, 1−α)
- Constructs a QPD_Johnson distribution from predicted quantiles
- Enforces quantile ordering and minimum spread for stability
- Optional bounded support via (lower, upper)
- Compatible with any sklearn regressor that has a `quantile` parameter

### Modified Files
* `skpro/regression/compose/_johnson_quantile.py` - New implementation
* `skpro/regression/compose/__init__.py` - Added import
* `docs/source/api_reference/regression.rst` - Added to API documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,9 @@ docs/api/*
 docs/_build/*
 cover/*
 MANIFEST
+
+# Virtual environments
+.venv/
+venv/
+env/
+ENV/
--- a/docs/source/api_reference/regression.rst
+++ b/docs/source/api_reference/regression.rst
@@ -173,7 +173,19 @@ Linear regression
     ARDRegression
     BayesianRidge
     GLMRegressor
+    GlumRegressor
     PoissonRegressor
+
+Generalized Additive Models
+---------------------------
+
+.. currentmodule:: skpro.regression.gam
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    GAMRegressor
 
 Gaussian process and kernel regression
 --------------------------------------
--- a/skpro/regression/gam/__init__.py
+++ b/skpro/regression/gam/__init__.py
@@ -1,0 +1,6 @@
+"""GAM regressor using pyGAM."""
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+
+from skpro.regression.gam._gam import GAMRegressor
+
+__all__ = ["GAMRegressor"]
--- a/skpro/regression/gam/_gam.py
+++ b/skpro/regression/gam/_gam.py
@@ -1,0 +1,323 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Generalized Additive Models (GAM) Regressor using pygam."""
+
+__author__ = ["Omswastik-11", "ravjot07"]
+
+import numpy as np
+import pandas as pd
+
+# from skpro.distributions.inversegaussian import InverseGaussian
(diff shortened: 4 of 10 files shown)


fkiraly (commented):
Very nice - I've started the tests.

RakshithaKowlikar:
Thanks for the feedback @fkiraly! I ran the automated tests via pytest which showed them being skipped with NOTSET. I haven't run `check_estimator` directly yet. Will resolving the failing checks.

fkiraly:
@RakshithaKowlikar, sounds like something is wrong with your local setup or our test setup with `pytest`. Could you outline on the dev-chat in the discord how you tried to run the tests and what happened? Might be a bug.

RakshithaKowlikar:
> @RakshithaKowlikar, sounds like something is wrong with your local setup or our test setup with `pytest`. Could you outline on the dev-chat in the discord how you tried to run the tests and what happened? Might be a bug.

Yes! The problem was on my end. I have rectified the code and now it doesn't show any fails. Also I think we should add venv to `.gitignore`
@fkiraly please let me know if further changes are required.

fkiraly (changes requested):
Quick question: why are you making changes to the qpd distributions? If these need changes, then maybe open a separate PR and explain what you are doing?

RakshithaKowlikar:
Changes to `qpd.py`:
- sample() method was added to fix `TypeError: object of type 'NoneType' has no len()`. Because QPD_Johnson was created with scalar parameters (no index/columns specified), self.index was None.
- theta was clamped to prevent arccosh from receiving invalid inputs when symmetric distribution occurs.
- wrap in `np.errstate` to suppress RuntimeWarnings during intermediate calculations because intermediate calculations can produce inf, NaN, or divide-by-zero even though the final result might be valid
- kappa calculation safety to check for zero before dividing.

Changes in `_johnson_quantile.py`:
- changed the spread calculation so that it works across different data scales, works when quantiles are too close together now
- added _predict_var because when ppf(scalar) is called on a distribution with shape (13, 1), it tries to return shape (13, 13), because ppf is in the base class, shared across all distributions. Fixing it there could break other distributions

@fkiraly should i still open a pr for the changes in `qpd.py`?

RakshithaKowlikar:
Also the errors seem to be linked with `GAMRegressor` where `get_test_params()` is returning a single dict

RakshithaKowlikar:
@fkiraly 
I've gone through the issues and here is what I plan to do:
- Remove `sample()` method in `qpd.py`,  instead  in `_johnson_quantile.py` use `.reshape(-1, 1)` instead of `.flatten()`, to ensure they are column vectors.
- When quantiles are symmetric, theta → 0.5, it causes arccosh(1/(2*0.5)) = arccosh(1.0) = 0 leading to divide by zero error. I plan to keep the clamping and safety margins to `qpd.py`

Yes, I did use AI to help debug due to time constraints. I understand all the changes and can explain the math.
If you think there's a better workaround, please let me know.

PR #638, opened 2025-11-20
[ENH] `pygam` regressor interface

Omswastik-11:
## Summary

This PR introduces a new **GAMRegressor** based on the `pygam` library, addressing **issue #634**.

---

### New files

* `gam.py` — main implementation of `GAMRegressor` with support for four distributions
* `test_gam.py` — tests for all major functionality

---

## Key Features

* Supports **Normal**, **Poisson**, **Gamma**, and **Binomial** models
* Allows flexible spline setup through the `terms` argument (auto terms or custom `pygam` terms)
* Maps `pygam` outputs to `skpro` distribution objects correctly
* Provides full probabilistic outputs via `predict_proba()`
* Integrates cleanly with the existing `BaseProbaRegressor` interface

---

## Implementation Notes

* Wraps `pygam.GAM`, choosing distribution and link functions automatically

* Parameter mapping:

  * **Normal:** σ² = scale
  * **Poisson:** μ = mean prediction
  * **Gamma:** α = 1/scale, β = 1 / (scale·μ)
  * **Binomial:** p = μ / n (default n = 1 unless specified)

* Follows the design style of `GLMRegressor` and `CyclicBoosting`

---

## Testing

* Six test functions covering:

  * Each supported distribution
  * Custom spline terms
  * Validation of distribution objects
* All tests pass
* Uses `@pytest.mark.skipif` for the optional `pygam` dependency

---
### API 

```python
from skpro.regression.gam import GAMRegressor
from pygam import s
import pandas as pd
import numpy as np

# dummy data
X = pd.DataFrame({
    "x1": np.linspace(0, 10, 200),
    "x2": np.random.randn(200),
})
y = pd.DataFrame({"y": 3 * np.sin(X["x1"]) + 0.5 * X["x2"] + np.random.randn(200)})

# fit a GAM with automatic spline terms
reg = GAMRegressor(distribution="normal", terms="auto")
reg.fit(X, y)

y_pred = reg.predict(X)

# probabilistic prediction (returns skpro distribution object)
y_proba = reg.predict_proba(X)
mean = y_proba.mean()
intervals = reg.predict_interval(X, coverage=0.9)

# custom spline terms
reg_custom = GAMRegressor(
    distribution="normal",
    terms=s(0, n_splines=10) + s(1, n_splines=5),
)
reg_custom.fit(X, y)
```

---

## Future Work

* Add **InverseGaussian** distribution support (planned in #635)
* Possible move towards an adapter-style design if maintainers prefer it

PR #658, opened 2025-11-26
[DOC] add missing API reference entry for `GAMRegressor`

neha222222:
Solves - [#634](https://github.com/sktime/skpro/issues/634)


fkiraly (changes requested):
Sorry for this, but https://github.com/sktime/skpro/pull/638 already has a complete implementation.

However, your PR adds the API docs reference which turns out was missing, so I will turn the PR into an API reference documentation one.

PR #667, opened 2025-12-01
[ENH] Add `interval_type` for configurable bounds in `TruncatedDistribution`

### LINKED ISSUE #666: [ENH] Extend `TruncatedDistribution` to support inclusive/exclusive bounds
Currently, `TruncatedDistribution`  has a specific truncation logic where the lower bound is exclusive and the upper bound is inclusive. This is not suitable for implementing distributions that require different boundary conditions. For example, when implementing Zero-Inflated distributions, we need to truncate a base distribution such that the lower bound (in this case `0`) is included.

Extend `TruncatedDistribution` to allow configuration of bound inclusivity for both limits. Add parameter like `inclusive_lower` and `inclusive_upper` with defaults set to preserve existing behaviour.

A possible workaround for example for a lower bound would be it to make it slightly below the target integer, by subtracting a small value `np.finfo(float).eps`.

**Additional context**
There has been some discussion on this during implementation of zero inflated models (#554)

Khushmagrawal:
#### Reference Issues/PRs
Fixes #666

This PR extends `TruncatedDistribution` to support an inclusive lower bound.

None

- Correctness of implementation

Yes, updated `get_test_params`  to include a test case with `inclusive_lower=True`.

#### Any other comments?
This PR currently only implements `inclusive_lower` (and not `inclusive_upper`) as this was the immediate blocker for Zero-Inflated models. However, the logic can be similarly applied to `inclusive_upper`. If this implementation pattern is correct, we can add inclusive_upper to this PR

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/truncated.py
+++ b/skpro/distributions/truncated.py
@@ -17,8 +17,9 @@ class TruncatedDistribution(BaseDistribution):
     .. math::
         Y \sim f(y \vert y \in I) = \frac{f(y)}{P(Y \in I)},
 
-    where :math:`I` is the interval defined by the bounds and ``interval_type``
-    and :math:`P(Y \in I)` is the total probability mass within that interval.
+    where :math:`I` is the interval defined by the bounds and ``interval_type``,
+    :math:`P(Y \in I)` is the total probability mass within that interval and
+    :math:`f(y)` is the probability mass/density function.
 
     Parameters
     ----------


skpro/distributions/truncated.py
```diff
@@ -98,9 +104,19 @@ def __init__(
             self.set_tags(**{"distr:paramtype": inner_paramtype})
 
     def _get_low_high_prob(self) -> Tuple[float, float]:
-        prob_at_lower = (
-            self.distribution.cdf(self.lower) if self.lower is not None else 0.0
-        )
+        if self.lower is not None:
```
fkiraly: this can be stated more easily by adding the correction from the pmf, can you please simplify the conditions, and leave the original line intact?

fkiraly (changes requested):
Great addition!

For easier use, may I instead suggest to add a single argument `interval_type` that can be one of the strings `"(]"` (default), `"[]"`, `"[)"`, and `"()"`? Indicating whether lower/upper bound are inclusive or not.

Khushmagrawal:
> For easier use, may I instead suggest to add a single argument `interval_type` that can be one of the strings `"(]"` (default), `"[]"`, `"[)"`, and `"()"`? Indicating whether lower/upper bound are inclusive or not.

Yes that will be more intuitive for user, I have updated the code

Khushmagrawal:
While updating `ZeroInflated` distribution to reflect this, I noticed that one-sided truncation creates a slight ambiguity. If only `lower` is defined (and `upper` is `None`), we would have to specify interval type as for example `[)` or `[]` even though the upper bound doesn't exist.

The code logic handles this already as it skips checking for upper bound bracket when `upper` is `None` but wanted to confirm if this is fine?

fkiraly:
> The code logic handles this already as it skips checking for upper bound bracket when upper is None but wanted to confirm if this is fine?

I think that is fine - if an inteval goes to infinity (no truncation), then the mathematical symbol used is the round bracket. But I think the best way to handle this is to ignore that bracket altogether, i.e., the one on the untruncated end.

Khushmagrawal:
> I think that is fine - if an inteval goes to infinity (no truncation), then the mathematical symbol used is the round bracket. But I think the best way to handle this is to ignore that bracket altogether, i.e., the one on the untruncated end.

Yes and the current code logic follows that.

skpro/distributions/truncated.py
```diff
@@ -7,29 +7,38 @@
...
     Mathematically, it can be expressed as:
 
     .. math::
-        Y \sim f(y \vert a \lt y \leq b) = \frac{f(y)}{F(b) - F(a)},
+        Y \sim f(y \vert y \in I) = \frac{f(y)}{P(Y \in I)},
 
-    where :math:`a` and :math:`b` is the lower and upper bound respectively, and
-    :math:`f(y)` is the probability mass/density function.
+    where :math:`I` is the interval defined by the bounds and ``interval_type``
+    and :math:`P(Y \in I)` is the total probability mass within that interval.
```
fkiraly: can you add a paragraph that explains briefly which interval types are supported, and how `I` relates to the parameters`?

It should also be said that the default is `(]`.

skpro/distributions/truncated.py
```diff
@@ -7,29 +7,38 @@
...
+        The lower bound below which values are truncated.
+        By default, this bound is exclusive (see ``interval_type``).
 
     upper : Union[float, int], optional
         The upper bound above which values are truncated.
+        By default, this bound is inclusive (see ``interval_type``).
+
+    interval_type : str, default="(]"
+        Defines the inclusivity of the bounds. Must be one of:
+        - "[]" : Closed interval (inclusive lower, inclusive upper).
```
fkiraly, line 40: to render correctly, an empty newline needs to precede the bullet point

skpro/distributions/truncated.py
```diff
@@ -105,12 +123,32 @@ def _get_low_high_prob(self) -> Tuple[float, float]:
             self.distribution.cdf(self.upper) if self.upper is not None else 1.0
         )
 
+        if self.get_tag("distr:measuretype") == "discrete":
```
fkiraly: I think this should be `!="continuous"`, since mixed distributions can also have a non-zero mass.

skpro/distributions/truncated.py
```diff
@@ -105,12 +123,32 @@ def _get_low_high_prob(self) -> Tuple[float, float]:
...
+            if self.upper is not None and not self._inclusive_upper:
+                prob_at_upper -= self.distribution.pmf(self.upper)
+
         return prob_at_lower, prob_at_upper
 
     def _calculate_density(self, x: np.ndarray, fun, as_log: bool):
         inf = float("inf")
 
-        is_valid = (x > (self.lower or -inf)) & (x <= (self.upper or inf))
+        lower_bound = self.lower if self.lower is not None else -inf
```
fkiraly, line 140: shorter but slightly less clear: `self.lower or -inf`

skpro/distributions/truncated.py
```diff
@@ -237,4 +277,13 @@ def get_test_params(cls, parameter_set="default"):  # noqa: D102
...
 
-        return [params1, params2, params3, params4]
+        # inclusive lower and exclusive upper
+        dist = Normal(mu=1.0, sigma=1.0)
+        params5 = {
+            "distribution": dist,
+            "lower": 0,
+            "upper": 5,
+            "interval_type": "[)",
+        }
```
fkiraly, line 289: can you please add one example for each interval type?

skpro/distributions/truncated.py
Khushmagrawal: Ah, are you suggesting to use 
`lower_bound = self.lower if self.lower else -inf`
In that case, it would evaluate to `-inf` when lower bound is zero (`self.lower = 0`), right?

skpro/distributions/truncated.py
Khushmagrawal: now corrected

skpro/distributions/truncated.py
Khushmagrawal: I have added the paragraph below explaining the parameter `interval_type`, would that be fine? or should I add something like (see `interval_type`) over here?

fkiraly: yes, this is great!

I am noticing a point of imprecision (which was there before): it should be said that `f` is the pdf.

skpro/distributions/truncated.py
fkiraly: no, I am suggesting to use `lower_bound = self.lower or -inf`

Khushmagrawal: Oh, but still this would evalute to `-inf` when the lower bound is at zero (`self.lower =  0`) which will be wrong, no?

fkiraly: ah, I see - yes! You are right.

fkiraly (approved):
Great!

Issue #670, opened 2025-12-03
[BUG] `XGBoostLSS` hardcodes stabilization, response and loss params


marrov:
**Describe the bug**
Current implementation of `XGBoostLSS` seem to hardcode and override some parameters in `_fit` () before passing it to the original implementation, which does not respect the user provided inputs in the init.

```python
xgblss = XGBoostLSS(
    xgblss_distr(
        stabilization="None",  # Should be self.stabilization
        response_fn="exp",      # Should be self.response_fn
        loss_fn="nll",          # Should be self.loss_fn
    )
)
```

Am I reading this wrong or was this done intentionally?

fkiraly:
looks like a bug that the reviewer (me) did not spot

PR #672, opened 2025-12-04
[ENH] in `XGBoostLSS`, expose underlying parameters when `n_trail=0`

### LINKED ISSUE #671: [BUG] `XGBoostLSS` wrapper should expose the underlying XGBoost params
**Describe the bug**

```python
        if self.n_trials == 0:
```

```python
cv = ...
estimator = XGBoostLSS()
```

joshdunnlime:
Allows passing of xgb_params to the xgboostlss model when `n_trail=0`.

Also allows the passthrough of `stabilization` and `response_fn`.

#### Reference Issues/PRs

Fixes #671 #670 

Not a major change. Just adds a new kwarg `xgb_params` and passes this through to xgboostlss.

No.

Style, test setup and docs.

Yes. I passed extra params to `get_test_params` and a `test_xgboostlss` standalone test.

#### Any other comments?

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -48,19 +48,19 @@ jobs:
           python -m build --wheel --sdist --outdir wheelhouse
 
       - name: Store wheels
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: wheels
           path: wheelhouse/*
 
-  test_unix_wheels:
+  test_wheels:
     needs: build_wheels
     name: Test wheels on ${{ matrix.os }} with ${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false  # to not fail all combinations if just one fail
       matrix:
-        os: [ubuntu-latest, macOS-latest]
+        os: [windows-latest, ubuntu-latest, macOS-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
 
     steps:
@@ -69,43 +69,18 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: wheels
           path: wheelhouse
 
-      - name: Get wheel filename
+      # Set wheel filename differently for Unix vs Windows
+      - name: Get wheel filename (Unix)
+        if: runner.os != 'Windows'
         run: echo "WHEELNAME=$(ls ./wheelhouse/skpro-*none-any.whl)" >> $GITHUB_ENV
 
-      - name: Install wheel and extras
-        run: python -m pip install "${{ env.WHEELNAME }}[all_extras,dev]"
-
-      - name: Run tests
-        run: |
-          python -m pytest
-
-  test_windows_wheels:
-    needs: build_wheels
-    name: Test wheels on ${{ matrix.os }} with ${{ matrix.python-version }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false  # to not fail all combinations if just one fail
-      matrix:
-        os: [windows-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-
-    steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-python@v6
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - uses: actions/download-artifact@v6
-        with:
-          name: wheels
-          path: wheelhouse
-
-      - name: Get wheel filename
+      - name: Get wheel filename (Windows)
+        if: runner.os == 'Windows'
         run: echo "WHEELNAME=$(ls ./wheelhouse/skpro-*none-any.whl)" >> $env:GITHUB_ENV
 
       - name: Install wheel and extras
@@ -118,13 +93,13 @@ jobs:
   upload_wheels:
     name: Upload wheels to PyPI
     runs-on: ubuntu-latest
-    needs: [build_wheels,test_unix_wheels,test_windows_wheels]
+    needs: [build_wheels,test_wheels]
 
     permissions:
       id-token: write
 
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: wheels
           path: wheelhouse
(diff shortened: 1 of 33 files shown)


marrov:
There are dependency failures and you forgot to mention this PR also fixes #670  - other than that LGTM

joshdunnlime:
@marrov I didn't see that other issue but noticed this when fixing the 2nd one. Thanks - added above for clarity.

The dependency issues are detailed here: #643. The test failures are also referenced in there by @fkiraly - alternative fix required.

> I would expose them explicitly, or the individual keys in `xgb_params` cannot be addressed by `set_params` and, in consequence, `sktime` or `skpro` tuners.

Agreed, this is an issue. However, XGBoostLSS itself doesn't expose this, which is somewhat understandable because of the number of there params that need to be exposed in XGBoost itself: https://github.com/dmlc/xgboost/blob/c3a5e3e66b1e46a191a479b791204a8ccd10ce60/python-package/xgboost/sklearn.py#L236
https://github.com/dmlc/xgboost/blob/c3a5e3e66b1e46a191a479b791204a8ccd10ce60/python-package/xgboost/sklearn.py#L808

What about passing kwargs and collecting them in self.xgb_params? This should work for skpro and possibly sklearn gridsearches. This means the API stays the same if we add explicit kwargs in the future but allows passthrough now.

We can open an issue on XGBoostLSS to explicitly add XGBoost parameters and one here to add XGBoostLSS after that.

fkiraly:
FYI, I merged from `main` and removed the overrides to `set_params` etc.

You need to make all individual parameters explicit in `__init__` - this is a requirement by the `sklearn`-like APIs.

(I assume you can just copy-paste from xgboost, also ensure you cover them via `get_test_params`)

joshdunnlime:
I saw the xgblss deps fix. I'll pull that.

I have started implementing the explicitly named kwargs fix but there are a lot of potential parameters and several edge-cases that are introduced.

XGBoostLSS leverages the standard XGBoost API, while the sklearn API uses different parameter names for well established sklearn parameters. Should skpro (as an extension of sklearn) adopt the sklearn parameter names or (as an extension of XGBLSS) adopt the default XGBoost parameters?
E.g. `n_estimators` vs `num_boost_round`, `learning_rate` vs `eta`. I am currently parsing both but might be overkill... Please advise.

Additionally, there was a new skpro `n_cpu` parameter introduced in the original skpro-XGBLSS implementation. This deviates from the `n_jobs` of sklearn and `nthreads` of XGBoost. I think this should be deprecated to conform with one of the others, in favour of `n_jobs` as it is used elsewhere in skpro.

joshdunnlime:
~I assume the failure is because we need to use `_xgb_params` for "fitted" prams?~

I was passing `xgb_params` in one of the tests still...

marrov:
Small comment: XGBoostLSS just added a new param for the distributions called "initialize". In my project, this seems quite important (and was the default behaviour) but it now defaults to `False` (not initialize). So, my recommendation would be to add this param to and update the dep boundary for xgboostlss. You can check how I added this in my fork [here](https://github.com/marrov/skpro/tree/fix-quantile-in-base).

fkiraly:
@statmixedml, did you make an API breaking change, or did this get accidentally merged with someone else's pull request?

fkiraly:
>XGBoostLSS leverages the standard XGBoost API, while the sklearn API uses different parameter names for well established sklearn parameters. Should skpro (as an extension of sklearn) adopt the sklearn parameter names or (as an extension of XGBLSS) adopt the default XGBoost parameters?
E.g. n_estimators vs num_boost_round, learning_rate vs eta. I am currently parsing both but might be overkill... Please advise.

I would not mind either way, as long as all the parameters are explicit, and as long as we have consistency between estimators from the @statmixedml stack.

skpro/regression/xgboostlss.py
```diff
@@ -37,22 +39,47 @@ class XGBoostLSS(BaseProbaRegressor):
...
+    n_jobs: int, optional, default=None
+        Number of CPUs to use for parallel processing of data in ``xgboostlss``.
+        If None, then the value of n_cpu is used.
+
     n_cpu: int or str, optional, default="auto"
         Number of CPUs to use for parallel processing of data in ``xgboostlss``.
-        Default is "auto" which uses all available CPUs.
+        Default is "auto" which uses all available CPUs. If an integer is given,
+        that number of CPUs is used. If both n_cpu and n_jobs are set, an error
+        is raised. `n_jobs` should be preferred to `n_cpu` for sklearn compatibility.
```
fkiraly, line 56: for proper rendering, code strings should be in double backticks, not single backticks.

joshdunnlime:
> Great!
> 
> Do I understand correctly that some parameters have two names, and only one of the two should be passed? I think this is confusing for the user. I would go with the `xgboost` (sklearn adapter) name only.
> 
> Further, could you please add all parameters explicitly in the docstring? It is a nightmarishly long list, but you should be able to copy paste it from xgboost code base?

The problem with XGBoost-sklearn params is that XGBLSS uses `xgb.train` interanlly, which means parsing them from XGBoost-sklearn specific params to XGB native params anyway (and potentially in a confusing matter if an error is raised referencing `eta`, when `learning_rate` was pass - maybe better to document them together?). Going with XGB native params means deviating from sklearn nomenclature. If we are doing some parsing already, my rational was to allow users to explicitly choose their convention.

The acceptance of multiple aliases is fairly well understood within the "boosting" user base. LightGBM also handles many aliases very nicely. It might be worth considering how we handle this ahead of LightGBMLSS being [implemented](https://github.com/sktime/skpro/pull/639).

In light of this, do you still wish to go with the XGB-sklearn convention, switch to XGB or allow both?

On the docstring, the slight problem here is that params are quite version specific. Most params are stable after 2.0 (except for Lambdarank and gpu specific kwargs). XGBLSS already pins to xgboost>=2.0, so as long as we're comfortable with these params being excluded (or documented and not available), then I can copy-and-paste that over.

fkiraly:
> In light of this, do you still wish to go with the XGB-sklearn convention, switch to XGB or allow both?

If you prefer to allow both, then I would not be against it.

However, I would document parameters differently:

```
paramname : abcdef, optional, default = 3
paramname 2: alias for paramname
    Description of parameter.
    More details
```

> XGBLSS already pins to xgboost>=2.0, so as long as we're comfortable with these params being excluded (or documented and not available), then I can copy-and-paste that over.

Yes, I think that is sufficient - if xgblss pins xgboost, then you can assume the parameter names are as in `xgboost>=2.0`.
Or something similar, that avoids user confusion by having multiple blocks.

fkiraly (approved):
Great!

I fixed downwards compatibility with 0.6.0 and below, taking care of the `initialize` parameter

joshdunnlime:
Thank for picking this up. It fell off my radar.

fkiraly:
also mine - I was shocked that I had not merged it when @marrov opened the fix for the same issues (because I thought I had), many apologies!

PR #674, opened 2025-12-06
[ENH] Sync skpro and sktime probabilistic metrics modules: sample_weight and output consistency

Omswastik-11:
## Summary

This PR brings the `skpro` probabilistic metrics module into alignment with `sktime`, as discussed in **issue #367**.
It adds **sample_weight** support across metrics, ensures **consistent output types**, and fixes **weighted multioutput** behavior.

---

### 1. **Sample Weight Support**

* Added a `sample_weight` parameter to:

  * `BaseProbaMetric.evaluate`
  * `BaseDistrMetric.evaluate`
* Updated input validation methods:

  * `_check_consistent_input`
  * `_check_ys`
    to correctly propagate `sample_weight`
* Implemented proper weighted averaging using `np.average` in:

  * `_evaluate` (Proba metrics)
  * `evaluate` (Distribution metrics)

---

### 2. **Output Consistency Improvements**

* `BaseDistrMetric.evaluate` now returns a **pd.Series** when:

  * `multioutput="raw_values"`
  * the metric is univariate (non-multivariate)
* This matches:

  * `BaseProbaMetric` behavior
  * `sktime` conventions
* Fixes the previous inconsistency where a 1-row DataFrame was returned.

---

### 3. **Weighted Multioutput Fixes**

* Improved handling of **array-like multioutput weights** in `BaseDistrMetric.evaluate`
* Ensures correct weighted aggregation across variables
* Fixes incorrect averaging when users pass custom weights.

---

## Tests Added / Updated

* **`test_sample_weight_pinball`** (in `test_probabilistic_metrics.py`)
  Verifies sample_weight support for quantile and interval metrics.

* **`test_sample_weight_logloss`** (in `test_distr_metrics.py`)
  Verifies sample_weight support for distribution-based metrics.

* **`test_multioutput_weights_logloss`** (in `test_distr_metrics.py`)
  Ensures weighted multioutput aggregation behaves correctly.

* **Updated `test_distr_evaluate`** to assert that raw-value outputs are returned as **pd.Series**.

Fixes #367

PR #675, opened 2025-12-07
[ENH] merge test_probabilistic_metrics into TestAllDistrMetrics

### LINKED ISSUE #398: [ENH] merge `test_probabilistic_metrics` into `TestAllDistrMetrics`
The module `test_probabilistic_metrics` contains tests for interval and quantile distributions with manual collection. These metrics are also automatically collected in `TestAllDistrMetrics`, which should be the location where these tests are carried out systematically.

Hence, we should move the existing tests in `test_probabilistic_metrics` to there.

Whether a metric is an interval or quantile metric can be inspectedby reading the `"scitype:y_pred"` tag.

Omswastik-11:
Fixes #398


Omswastik-11:
Hi @fkiraly !!

I think this will conflict with #674 . 
I think after merging #674 I need to rework on this to work according to your suggestion in #398 . Any thoughts ?

PR #699, opened 2026-01-12
[ENH] Fix `quantile()` in `BaseDistribution` and `XGBoostLSS` wrapper

### LINKED ISSUE #670: [BUG] `XGBoostLSS` hardcodes stabilization, response and loss params

### LINKED ISSUE #671: [BUG] `XGBoostLSS` wrapper should expose the underlying XGBoost params

### LINKED ISSUE #678: [BUG] `quantile` method of `BaseDistribution` fails for multi-index inputs

marrov:
#### Reference Issues/PRs
- Fixes #670 
- Fixes #671 
- Fixes #678

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -1643,7 +1643,6 @@ class BaseDistribution(BaseObject):
             with same ``columns`` as ``self``, and row ``MultiIndex`` that is product
             of ``RangeIndex(n_samples)`` and ``self.index``
         """
-
         return self._sample(n_samples=n_samples)
 
     def _sample(self, n_samples=None):


marrov:
> Hm, this seems to trip some tests - have you checked?
> 
> I would also suggest to split this PR into one for the base class index handling, and one for xgboost, that can be merged quickly.

Split into: #700  and #701 - closing this one now

PR #700, opened 2026-01-12
[BUG] Fix `XGBoostLSS` wrapper parameterization

### LINKED ISSUE #670: [BUG] `XGBoostLSS` hardcodes stabilization, response and loss params

### LINKED ISSUE #671: [BUG] `XGBoostLSS` wrapper should expose the underlying XGBoost params

marrov:
#### Reference Issues/PRs
- Fixes #670 
- Fixes #671

### DIFF
--- a/skpro/regression/xgboostlss.py
+++ b/skpro/regression/xgboostlss.py
@@ -61,6 +61,15 @@ class XGBoostLSS(BaseProbaRegressor):
         The number of trials in tuning.
         If this argument is set to None, there is no limitation on the number of trials.
         If set to 0, no tuning is done, and default parameters of XGBoostLSS are used.
+
+    xgb_params: dict, optional, default=None
+        Dictionary of XGBoost parameters to use when n_trials == 0.
+        If None and n_trials == 0, default XGBoost parameters are used.
+        Allows external hyperparameter tuning without breaking time series dependencies.
+
+    initialize: bool, optional, default=True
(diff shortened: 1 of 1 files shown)


marrov:
Current errors are due to the `initialize` parameter being a new addition in `0.6.1` of `xgboostlss`. This is a breaking interface change. Seems to be a big overlap between this PR and #672

fkiraly:
Thanks! I think we should merge https://github.com/sktime/skpro/pull/672 instead, to ensure the parameters are explicit.

I also added a fix for the `initialize` parameter there.

fkiraly:
I will merge this - for the credits re the `initialize` parameter and pass-on handling; and then https://github.com/sktime/skpro/pull/672 as the final fix.

PR #704, opened 2026-01-16
[ENH] Optimize `EnbpiRegressor` through faster sampling in `Empirical`

marrov:
#### Reference Issues/PRs

Addresses performance issues with EnbpiRegressor found when used with [Monte Carlo recursive probabilistic forecaster](https://github.com/sktime/sktime/pull/9242). This optimization enables practical usage of conformal prediction intervals in large-scale MC forecasting scenarios.

This PR updates ~~introduces `FastEnbpiRegressor`, a drop-in replacement for~~ `EnbpiRegressor` that provides ~10-100x speedup through two optimizations

~~1. Vectorized Aggregation in `_predict_proba()`: Replaces the O(n_train × n_test) nested loop with chunked vectorized numpy operations that process predictions in batches. This avoids the quadratic complexity of the original implementation.~~

~~2. Optimized `_FastEmpiricalEnbpi` Distribution Class: A specialized `Empirical` subclass that:~~
   - Skips expensive `_init_sorted()` initialization (~10x faster)
   - Implements custom `_sample()` using direct numpy indexing instead of pandas MultiIndex operations (~1000x faster sampling)
   - Stores raw prediction and error arrays for efficient resampling

Edit: optimizations mentioned above (strikethrough) still apply but are now used on `EnbpiRegressor` and `Empirical` directly.

**Performance Metrics:**
- Single batch prediction: 1.2-1.5x faster
- Sampling 100 from distribution: 2500x faster
- Small MC workflow (10 steps, 100 samples each): 22x faster (9.9s → 0.45s)
- Realistic-scale MC forecasting: ~100x faster (10+ minutes → 5 seconds)

No. The implementation uses only existing dependencies: NumPy, Pandas, and skpro's existing classes.

- ~~Correctness of numerical results: Verify that `_FastEmpiricalEnbpi` produces statistically equivalent samples and statistics as standard `Empirical`~~
- Appropriateness of skipping `_init_sorted()`: Confirm that quantiles computed on-demand from raw arrays are accurate
- Custom `_sample()` implementation: Review the NumPy indexing logic for correctness across different batch and sample sizes
- ~~Generalizability: he optimizations in the `_FastEmpiricalEnbpi` class could potentially benefit other estimators creating large empirical distributions (see "Any other comments" section)~~

No, but I should. Tested with throw-away scripts for this draft. If ok-ed will produce tests.

#### Any other comments?

~~The optimizations in `_FastEmpiricalEnbpi` are not EnbPI-specific. The performance findings suggest that:~~
- ~~Skipping `_init_sorted()` benefits any large Empirical distribution~~
- ~~Custom sampling using raw arrays is a general pattern applicable to distributions where parameters can be efficiently stored~~

~~A follow-up enhancement (separate PR) could add a `skip_init_sorted` parameter to the base `Empirical` class or create a general `FastEmpirical` class in the distributions module to benefit other estimators.~~

~~Note: the base branch here includes my fix for `quantile()` in the base distribution as I was unable to test without it. When #697 is merged, I will discard those changes and use main as the base branch.~~

Edit: base branch fixed and optimizations are now applied to original classes.

### DIFF
--- a/skpro/distributions/empirical.py
+++ b/skpro/distributions/empirical.py
@@ -41,7 +41,8 @@ class Empirical(BaseDistribution):
         if False, ``sample`` will sample entire instances from ``spl``
     skip_init_sorted : bool, optional, default=False
         if True, skips sorting of empirical samples on init and instead builds
-        sorted caches lazily when cdf/ppf/energy are first called
+        sorted caches lazily when cdf/ppf/energy are first called.
+        Performance flag only, does not change statistical behavior.
     index : pd.Index, optional, default = RangeIndex
     columns : pd.Index, optional, default = RangeIndex
 


fkiraly (changes requested):
Interesting! I am wondering whether this can be applied to the `Empirical` distribution directly?

Either way, I think you have accidentally included changes to the `xgboostlss` regressor, you need to update from `main` (and also update your fork)

marrov:
> Interesting! I am wondering whether this can be applied to the `Empirical` distribution directly?

Yes, I mentioned this in the PR OP under "**Any other comments?**". 

So, would you advocate for accepting this and then move to the base distribution or should I open a PR for these optimizations in `Empirical` first and then make this PR dependent on that one?

> Either way, I think you have accidentally included changes to the `xgboostlss` regressor, you need to update from `main` (and also update your fork)

@fkiraly - Indeed, my bad! Just the latest changes and rebased onto main. Now all tests are passing.

marrov:
> I am a bit apprehensive of duplicating the estimator as well as the empirical distribution.
> 
> Can we not just _replace_ both directly? Given that these are just faster versions of either?

@fkiraly - Sure thing! Just wanted your green-lighting this before changing the original ones. Updates are now applied to the original classes and the extra ones have been removed & all tests are passing.

skpro/distributions/empirical.py
```diff
@@ -41,6 +41,9 @@ class Empirical(BaseDistribution):
         if False, ``sample`` will sample entire instances from ``spl``
     index : pd.Index, optional, default = RangeIndex
     columns : pd.Index, optional, default = RangeIndex
+    skip_init_sorted : bool, optional, default=False
```
fkiraly, line 42: `index` and `columns` must always be the last args according to current implicit API contract of distributions.
(surprised this is not tested, we should add a test)

skpro/regression/enbpi.py
```diff
@@ -279,12 +278,41 @@ def _predict_proba(self, X) -> np.ndarray:
...
+            masked_preds = np.where(
+                bs_wo_ix_expanded, y_preds_chunk[np.newaxis, :, :, :], np.nan
+            )
+
+            y_preds_agg[:, start_idx:end_idx, :] = self._agg_fun(masked_preds, axis=1)
+
+        # Handle any NaN values in aggregated predictions, per training sample
+        nans = np.isnan(y_preds_agg)
+        if nans.any():
+            for i in range(n_train):
```
fkiraly: I am surprised these loops are more efficient? Is this really necessary, can we not use `numpy` array operations? Why is this more efficient?

skpro/regression/enbpi.py
```diff
@@ -302,7 +330,7 @@ def _predict_proba(self, X) -> np.ndarray:
         emp_ix = pd.MultiIndex.from_product([range(n_emp_spl), X.index])
 
         spl_df = pd.DataFrame(emp_df_vals, index=emp_ix, columns=cols)
-        y_proba = Empirical(spl_df)
+        y_proba = Empirical(spl_df, skip_init_sorted=True)
```
fkiraly, line 305: hmmm, should we add this as a parameter to the enbpi-regressor too?

Also, does this impact the row index order of the distribution returned, or not?

fkiraly (changes requested):
Great! I left some requests and some questions for my own clarification.

skpro/distributions/empirical.py
marrov: Fixed in `Empirical`. Also added a test a `test_all_distrs` that checks three cases:
* If both index and columns present: they must be the last two parameters (in either order)
* If only index present: it must be the last parameter
* If only columns present: if must be the last parameter
* If neither present: the test passes (nothing to check)

Can you confirm I read this correctly?

fkiraly: Not quite - all distributions must have `index` and `columns` as last parameters, in this precise order. Having just one or none should be an API violation.

skpro/regression/enbpi.py
marrov: I think we shouldn't expose it, no, because it will always be `True` for this class and it is really just an internal optimization, not really a meaningful user configuration. I made sure to add a test to ensure that the statistical behavior is unchanged. The only difference is that initialization is deferred. So if cdf/ppf are called, the "lazy initialization" handles the sorting.

skpro/regression/enbpi.py
marrov: Good intuition! I think I assumed these changes would have an effect and tested them alongside the changes to `Empirical` and saw an improvement. However, after dissecting the changes, this is did produce any major enhancements. I tried a different approach with a matrix multiplication `bs_vs_ix @ y_preds` using the identity $$Sum_LOO = Sum_Total - Sum_In_Bag$$ to replace the nested loops. This did improve the aggregation quite a bit (~50x speedup) but it did not matter too much when trying the whole class end to end (speedup dependent on the size of `n_train`, average at about 1.5x-2x). So I have just reverter it back to the original code.

fkiraly: ok

skpro/distributions/tests/test_all_distrs.py
```diff
@@ -274,6 +276,28 @@ def test_ppf_and_cdf(self, object_instance):
         else:
             assert np.allclose(x, x_approx)
 
+    def test_index_columns_last_args(self, object_class):
+        """Test that index and columns are the last arguments in __init__."""
+        sig = inspect.signature(object_class.__init__)
```
fkiraly: there is actually a method for this in `scikit-base` `BaseObject` - simply use `get_param_names(sort=False)`

Issue #416, opened 2024-06-30
[BUG] `plot` method of the `Binomial` distribution shows incorrect pmf. 


meraldoantonio:
**Describe the bug**
The `plot` method of the `Binomial` distribution shows incorrect pmf. 

**To Reproduce**
```python
from skpro.distributions import Binomial
n = Binomial(n=10, p=0.5)
n.plot('pmf')
```
<img width="609" alt="image" src="https://github.com/sktime/skpro/assets/37468543/15fbddff-fe5a-4c83-aafd-4aae97a7db6f">

**Expected behavior**
The `plot.("pmf")`  should show the correct Binomial plot, which looks like this:
![image](https://github.com/sktime/skpro/assets/37468543/cfe019b6-b420-4db0-af19-b50437fffb52)

**Additional context**
Currently, the plotting function samples a grid of x points, plots x, f(x) for the function and connects them. If the grid does not hit the mass points, the plot will look like this.

The only way to have the correct plot always is to know where the support is. That is currenty not inspectable from the object, and that's why I wanted the sets symbolism from https://github.com/sktime/skpro/issues/244.

This bug is also referenced in #415

PR #711, opened 2026-01-24
[ENH] `_pmf_support` method for `BaseDistribution` returning inspectable mass support

arnavk23:
## Reference Issues/PRs

Related to #416 (discrete PMF plotting improvements). This PR provides the foundational infrastructure that enables more robust discrete distribution plotting in follow-up PRs.

## What does this implement/fix? Explain your changes.

This PR introduces a new `_pmf_support(lower, upper, max_points)` method to the `BaseDistribution` class that provides a standardized, extensible way to obtain support points for discrete distributions within specified bounds.

**Key Changes:**
- **Base Implementation** (`skpro/distributions/base/_base.py`): Added `_pmf_support` method that returns empty array for continuous distributions and non-negative integers for discrete distributions by default
- **Empirical Distribution Override** (`skpro/distributions/empirical.py`): Custom implementation that returns actual empirical support points within bounds
- **Delta Distribution Override** (`skpro/distributions/delta.py`): Implementation that returns point mass locations within bounds  
- **Comprehensive Tests** (`skpro/distributions/tests/test_proba_basic.py`): Added `test_pmf_support_method()` covering all distribution types

This addresses the need for extensible support point detection across different discrete distribution types, replacing hardcoded assumptions with a clean, overridable API.

## Does your contribution introduce a new dependency? If yes, which one?

No, this change introduces no new dependencies.

## What should a reviewer concentrate their feedback on?

- [x] API design and method signature consistency
- [x] Extensibility for future distribution types
- [x] Test coverage completeness
- [x] Performance implications for large empirical distributions
- [x] Backward compatibility

## Did you add any tests for the change?

Yes, I added comprehensive tests in `test_support_method()` that cover:
- Continuous distributions (should return empty arrays)
- Discrete distributions with default integer support
- Empirical distributions with custom support points
- Delta distributions with point masses
- Boundary condition handling

## Any other comments?

This PR establishes the core infrastructure for proper discrete distribution support handling. It serves as a foundation for improved plotting capabilities (as demonstrated in related PRs) and future enhancements like full support API implementation (issue #244). The design prioritizes extensibility while maintaining backward compatibility.

### For all contributions

### For new estimators

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -1801,6 +1801,11 @@ class BaseDistribution(BaseObject):
         np.ndarray
             Array of support points within [lower, upper]
         """
+        if self.ndim > 0:
+            raise NotImplementedError(
+                "_pmf_support only applies to scalar (0-dimensional) distributions."
+            )
+
         # For continuous distributions, return empty array
         if self.get_tag("distr:measuretype", "mixed") == "continuous":
             return np.array([])
--- a/skpro/distributions/delta.py
+++ b/skpro/distributions/delta.py
@@ -206,22 +206,17 @@ class Delta(BaseDistribution):
         np.ndarray
             Array of support points within [lower, upper]
         """
+        if self.ndim > 0:
+            raise NotImplementedError(
+                "_pmf_support only applies to scalar (0-dimensional) distributions."
+            )
+
         c = self._bc_params["c"]
-        if self.ndim == 0:
-            c_val = c[()]
-            if lower <= c_val <= upper:
-                return np.array([c_val])
-            else:
-                return np.array([])
+        c_val = c[()]
+        if lower <= c_val <= upper:
+            return np.array([c_val])
         else:
-            # For array case, collect all c values within bounds
-            support_points = []
-            for i in range(c.shape[0]):
-                for j in range(c.shape[1]):
-                    c_val = c[i, j]
-                    if lower <= c_val <= upper:
-                        support_points.append(c_val)
-            return np.unique(np.array(support_points))[:max_points]
+            return np.array([])
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):
--- a/skpro/distributions/empirical.py
+++ b/skpro/distributions/empirical.py
@@ -495,29 +495,20 @@ class Empirical(BaseDistribution):
         np.ndarray
             Array of support points within [lower, upper]
         """
+        if self.ndim > 0:
+            raise NotImplementedError(
+                "_pmf_support only applies to scalar (0-dimensional) distributions."
+            )
+
         self._ensure_sorted()
 
-        if self.ndim == 0:
-            # Scalar case: filter sorted support points within bounds
-            support_points = self._sorted
-            mask = (support_points >= lower) & (support_points <= upper)
-            filtered_points = support_points[mask]
-            # Limit to max_points and return unique values
-            unique_points = np.unique(filtered_points)
-            return unique_points[:max_points]
-        else:
-            # Array case: for simplicity, return union of all support points
-            # across instances and columns within bounds
-            all_points = []
-            for t in self._instances:
-                for col in self.columns:
-                    points = self._sorted[t][col]
-                    mask = (points >= lower) & (points <= upper)
-                    filtered = points[mask]
-                    all_points.extend(filtered)
-
-            unique_points = np.unique(np.array(all_points))
-            return unique_points[:max_points]
+        # Scalar case: filter sorted support points within bounds
+        support_points = self._sorted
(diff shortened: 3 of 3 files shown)


skpro/distributions/base/_base.py
```diff
@@ -1749,6 +1749,28 @@ def get_ax(ax, i, j, shape):
...
+            Upper bound for support points
+        max_points : int, optional, default=100
+            Maximum number of support points to return
+
+        Returns
+        -------
+        np.ndarray
+            Array of support points within [lower, upper]
+        """
+        # Default implementation assumes integer support
```
fkiraly: I would say, this should apply to the *discrete part* in general.

That means, it should return an empty 1D array for continuous distributions. Having non-negative integers for everything else sounds like a solid default (since most common).

skpro/distributions/empirical.py
```diff
@@ -185,6 +185,47 @@ def _ensure_sorted(self):
...
+            Upper bound for support points
+        max_points : int, optional, default=100
+            Maximum number of support points to return
+
+        Returns
+        -------
+        np.ndarray
+            Array of support points within [lower, upper]
+        """
+        self._ensure_sorted()
```
fkiraly: how do we handle that the support of every entry will be different if the distribution is 2D?

skpro/distributions/tests/test_proba_basic.py
```diff
@@ -137,13 +138,34 @@ def test_proba_plotting(fun):
     assert all([isinstance(a, Axes) for a in ax.flatten()])
     assert all([a.get_figure() == fig for a in ax.flatten()])
 
-    # scalar case
```
fkiraly, line 140: why do we remove some of the test logic?

arnavk23:
> Thanks for the additional thoughts about _support, I think this is great.

Here - Q1: 2D Support Handling: The _support method is called on scalar distributions (after [iloc[i,j] extracts each entry), not on the full 2D distribution. This means each entry gets its own support calculation.
For distributions with uniform support across entries (like [Binomial]), the base implementation works fine since all entries share the same support characteristics.
For distributions with entry-specific support (like Empirical or Delta), subclasses override _support to compute the appropriate support points for that specific entry based on its parameters.

In new pr (title - Incorrect pmf plotting)  - Q2: Test Changes: The original test only checked that a stem plot was created, but didn't verify it contained actual data points (Added it back). The updated version checked for multiple support points to ensure the plotting worked correctly.

fkiraly:
> The _support method is called on scalar distributions (after [iloc[i,j] extracts each entry), not on the full 2D distribution. This means each entry gets its own support calculation.

Ah, I see! So it can be called only on 0D-array (i.e., scalar) distributions, right?

Although where all entries share the same support, it is a bit unfortunate that it cannot be called from the 2D one. Not sure how to solve this. Perhaps distinguish *potential* (class level) and *actual* (instance & entry level)? Anyway, that might be best discussed in a design issue and is not blocking for this PR.

arnavk23:
> Small change - to make the purpose clearer, could we rename to `_pmf_support`?

Made the required change.

skpro/distributions/delta.py
```diff
@@ -189,6 +189,40 @@ def _ppf(self, p):
...
+            Array of support points within [lower, upper]
+        """
+        c = self._bc_params["c"]
+        if self.ndim == 0:
+            c_val = c[()]
+            if lower <= c_val <= upper:
+                return np.array([c_val])
+            else:
+                return np.array([])
+        else:
```
fkiraly, line 218: question: we said `_pmf_support` should work only for the 0D case. But there is a case distinction now for 2D, array case?

skpro/distributions/empirical.py
```diff
@@ -478,6 +478,47 @@ def _ppf(self, p):
...
+
+        if self.ndim == 0:
+            # Scalar case: filter sorted support points within bounds
+            support_points = self._sorted
+            mask = (support_points >= lower) & (support_points <= upper)
+            filtered_points = support_points[mask]
+            # Limit to max_points and return unique values
+            unique_points = np.unique(filtered_points)
+            return unique_points[:max_points]
+        else:
```
fkiraly: same here

fkiraly:
can you explain why the 2D code was there and how you want to resolve it?

arnavk23:
> can you explain why the 2D code was there and how you want to resolve it?

The 2D logic in _pmf_support handled multi-dimensional distributions by aggregating support points across all elements. This contradicted the base class, which defines _pmf_support as 0D-only. I enforced this by raising NotImplementedError for ndim > 0 and removed redundant 2D handling. The code is now aligned with the docs, simpler, and all tests pass.

fkiraly (approved):
Ok, thanks!

PR #715, opened 2026-01-30
[ENH] Implement cast-and-warn for slow index types in broadcasting #597

### LINKED ISSUE #597: [BUG] Repeated calls to `to_numpy` in the Distribution energy method cause notable slow down
**Describe the bug**

There is a niche but significant performance issue when using some "non-standard" index type for input data (X, y). Specifically, this problem is most obvious when running an evaluation using CRPS score on distributions with no `numerically exact implementation of the 'energy' method` as repeated sampling of the distribution and calls to `to_numpy` within the `_get_bc_params_dict` add serious overhead. In the example below it creates a 4x slow down!

The `to_numpy` on a tz-aware datetime index gave HUGE slow down (100,000x vs non-tz-aware DatetimeIndex). This is a pandas -> numpy issue where numpy doesn't support timezone-aware datetimes (same with other non-native numpy dtypes). I raised an issue with pandas https://github.com/pandas-dev/pandas/issues/62309#issuecomment-3294093815 but, as numpy is unlikely to add tz-aware datetimes, pandas cannot improve this. They might consider raising a warning. As mentioned, this issue is magnified in skpro in many contexts due to repeated sampling of the distribution.

Index types affects:
- tz-aware DatetimeIndex (4x slowdown)
- PeriodIndex (4x slowdown)
- IntervalIndex (12x slowdown)
- Datetime IntervalIndex (4x slowdown)
- MultiIndex with any level containing the above types (>4x slowdown depending on combo)

Not affected
- RangeIndex
- CategoricalIndex
- DatatimeIndex (not tz-aware)
- MulitIndex of the above types only

**To Reproduce**

In a jupyter notebook for cell timings:

```python
import numpy as np
import pandas as pd
from sklearn.ensemble import RandomForestRegressor
from sklearn.linear_model import LinearRegression
from sklearn.preprocessing import StandardScaler
from skpro.benchmarking.evaluate import evaluate
from skpro.metrics import CRPS
from skpro.regression.residual import ResidualDouble

length = 8000
X = np.random.uniform(1, 3, length)
X_ = X + np.random.normal(0, 0.5, length)
y = (X_ ** 2) + X + 1
Xy = pd.DataFrame({"X": X, "y": y})

cv = KFold(n_splits=3)

model = TransformedTargetRegressor(
    ResidualDouble(
        LinearRegression(),
        RandomForestRegressor(),
    ),
    transformer=StandardScaler() 
    # res double failed without SS
    # also provides non-numerical energy method
)

Xy_dt = (
    Xy
    .assign(datetime=pd.date_range("2023-01-01", freq="h", periods=length))
    .set_index('datetime')
)

Xy_dt_utc = Xy_dt.tz_localize('UTC')
Xy_dt_uk = Xy_dt_utc.tz_convert('Europe/London')

# non-native numpy index type
Xy_dt_utc = Xy_dt.tz_localize('UTC')
```
```python
X = Xy[['X']]
y = Xy[['y']]
evaluate(model, cv, X, y, scoring=CRPS())
```
> 4.6s

```python
X = Xy_dt_uk[['X']]
y = Xy_dt_uk[['y']]
evaluate(model, cv, X, y, scoring=CRPS())
```
> 19.2

I have an extended notebook that I will share with all other index types but it is a bit long for here.

**Expected behavior**

To run a bit faster or to raise a warning for non-standard dtypes. There are several options I see that can improve the user experience here:

1) use `.values` instead of calling `to_numpy`. This makes tz-aware run the same speed as non-tz-aware. This has one con in that it is not recommended by pandas and behaviour could change in the future. However, this is very unlikely as `.values` is used extensively within the Pandas community For more info see the release notes for [to_numpy](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html#accessing-the-values-in-a-series-or-index). Locally, this is what I have been using for a 4x speed-up.

2) Just raise a warning for non-standard indices. Users can preprocess their index accordingly, e.g. `Xy.tz_localize(None)`.

3) Consider a more significant change to `_get_bc_params_dict` and how it is called

**Environment**
This is a numpy dtype issue so speeds shouldn't vary significantly across OS or package versions.

I am running this on MacOS with skpro version 2.9.2, Pandas 2.3, numpy 1.26.4.

**Additional context**

siddharthbaleja7:
#### Reference Issues/PRs
Fixes #597.

This PR addresses a significant performance bottleneck (4x to 12x slowdown) identified when using "non-standard" pandas index types (tz-aware `DatetimeIndex`, `PeriodIndex`, and `IntervalIndex`) in distribution broadcasting. 

The slowdown occurs because `to_numpy()` is repeatedly called on types that NumPy does not natively support, leading to expensive overhead within `_get_bc_params_dict`.

**Changes:**
* **New Utility:** Added `_get_fast_index_np` to handle efficient index-to-numpy conversion.
* **Automatic Casting:** Automatically casts complex indices (e.g., tz-aware to naive UTC) to native NumPy formats once, avoiding repeated conversion costs.
* **Performance Warning:** Implements a `PerformanceWarning` to alert users when these indices are detected, advising them on preprocessing for optimal speed.
* **Broadcasting Optimization:** Updated internal logic to pre-convert and store these indices so that broadcasting uses high-speed raw arrays.

No.

* The logic for casting `PeriodIndex` and `IntervalIndex` to `object` to bypass pandas ExtensionArray overhead.
* Ensuring the `PerformanceWarning` is informative and points correctly to the user's call site (using `stacklevel=2`).

Yes. I have verified that:
- [x] A `PerformanceWarning` is raised when a tz-aware index is provided.
- [x] The internal `_index_np` is correctly cast to a native, non-tz-aware NumPy array.
- [x] The code passes all existing distribution tests and `flake8` linting.

#### Any other comments?
This "cast-and-warn" approach was discussed with and approved by **@fkiraly**. It provides a 4-12x speed-up for time-series and interval-based distributions.

#### PR checklist
- [x] I've added myself to the list of contributors.
- [x] The PR title starts with [ENH].

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -2,21 +2,13 @@
 """Base classes for probability distribution objects."""
 
 __author__ = ["fkiraly"]
-
 __all__ = ["BaseDistribution"]
 
 from warnings import warn
-
 import numpy as np
 import pandas as pd
 
-try:
-    from pandas.errors import PerformanceWarning
-except ImportError:
-    PerformanceWarning = UserWarning
-
 from skbase.utils.dependencies import _check_estimator_deps, _check_soft_dependencies
-
 from skpro.base import BaseObject
 
 
@@ -45,7 +37,7 @@ def _get_fast_index_np(index):
             f"The index type {type(index)} is known to be slow with to_numpy(). "
             "For optimal performance, please cast to a native numpy-compatible "
             "constant, e.g., via .tz_localize(None) for DatetimeIndex.",
-            PerformanceWarning,
+            UserWarning,
             stacklevel=2,
         )
 
--- a/skpro/distributions/base/tests/test_performance_warning.py
+++ b/skpro/distributions/base/tests/test_performance_warning.py
@@ -1,0 +1,91 @@
+
+import pytest
+import pandas as pd
+import numpy as np
+import warnings
+from skpro.distributions.base._base import _get_fast_index_np
+
+
+def test_warn_on_tz_aware():
+    """Test that UserWarning is raised for tz-aware DatetimeIndex."""
+    dt_idx = pd.date_range("2020-01-01", periods=5, tz="UTC")
+    with pytest.warns(UserWarning, match="known to be slow"):
+        _get_fast_index_np(dt_idx)
+
+
+def test_warn_on_period():
+    """Test that UserWarning is raised for PeriodIndex."""
+    period_idx = pd.period_range("2020-01-01", periods=5, freq="D")
+    with pytest.warns(UserWarning, match="known to be slow"):
+        _get_fast_index_np(period_idx)
+
+
+def test_warn_on_interval():
+    """Test that UserWarning is raised for IntervalIndex."""
+    interval_idx = pd.interval_range(start=0, end=5)
+    with pytest.warns(UserWarning, match="known to be slow"):
+        _get_fast_index_np(interval_idx)
+
+
+def test_no_warn_on_standard():
+    """Test that no warning is raised for standard Index types."""
+    # Standard Int64 Index
+    std_idx = pd.Index([1, 2, 3, 4, 5])
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        _get_fast_index_np(std_idx)
+        relevant_warnings = [
+            x for x in w if "known to be slow" in str(x.message)
+        ]
+        msg = f"Unexpected warnings: {[str(x.message) for x in relevant_warnings]}"
+        assert len(relevant_warnings) == 0, msg
+
+    # RangeIndex
+    range_idx = pd.RangeIndex(start=0, stop=5)
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        _get_fast_index_np(range_idx)
+        relevant_warnings = [
+            x for x in w if "known to be slow" in str(x.message)
+        ]
+        assert len(relevant_warnings) == 0
+
+    # Naive DatetimeIndex
+    naive_dt_idx = pd.date_range("2020-01-01", periods=5)
+    with warnings.catch_warnings(record=True) as w:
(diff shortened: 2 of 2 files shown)


siddharthbaleja7:
Hi @fkiraly and @joshdunnlime! I've finalized the implementation using the 'cast-and-warn' approach we discussed.

The code now automatically handles slow index types (tz-aware, Period, Interval) by casting them to native NumPy formats, bypassing the overhead mentioned in #597. I've also verified that it passes flake8 and triggers the PerformanceWarning correctly.

Looking forward to your feedback when you have a moment!

skpro/distributions/base/_base.py
```diff
@@ -9,11 +9,54 @@
 
 import numpy as np
 import pandas as pd
+
+try:
```
fkiraly: why the try/except here - is this version dependent?

fkiraly: either way, we should avoid the `pandas` import if it is not version stable.

fkiraly (changes requested):
Nice, thanks!

* can we avoid the `pandas` import of the `PerformanceWarning` and just use a builtin warning? It is not version stable, so looks like a bug waiting to happen.
* we should add tests that (a) the warning is raised and not raised whenever we expect, and (b) no failure in the different subcases for indices.

siddharthbaleja7:
Thanks for the review @fkiraly! I've updated the PR to use the builtin `UserWarning` as suggested to avoid any pandas version dependency issues.

I've also added a comprehensive test suite (`skpro/distributions/base/tests/test_performance_warning.py`) that verifies:
1. The warning is correctly raised for `tz-aware DatetimeIndex`, `PeriodIndex`, and `IntervalIndex`.
2. No warning is raised for standard `Index`, `RangeIndex`, or naive `DatetimeIndex`.
3. The automatic casting logic works correctly for all these cases without errors.

I have also verified that the code is lint-free and passes all tests. Let me know if this looks good!

PR #721, opened 2026-02-15
[ENH] `KernelMixture` distribution for kernel density estimation

### LINKED ISSUE #324: [ENH] kernel mixture distribution
For distribution estimation, an important distribution type are kernel mixture distributions.

An open design question is how to represent these - `Mixture` allows mixture of arbitrary distribution objects, but:

* not all current kernels are distributions or can be currently represented as distirbutions
* there are no objects in `skpro` for kernels a the moment, so there is an open desigh question how to represent these
* parameterization is unclear - most common in literature is kernel choice, kernel parameters, and points of support

Compare `sklearn` `KernelDensity` - https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KernelDensity.html#sklearn.neighbors.KernelDensity

amaydixit11:
#### Reference Issues/PRs
Fixes #324

This PR implements `KernelMixture`, a new distribution class for nonparametric kernel density estimation. Key features:

**Core functionality:**
- Weighted mixture of kernel functions centered at support points
- 5 kernels implementation: Gaussian, Epanechnikov, Tophat, Cosine, Linear
- Exact implementations of pdf, cdf, mean, variance, and sampling
- Fully vectorized operations for efficiency (avoids per-component distribution objects)

**Bandwidth selection:**
- Manual specification via float parameter
- Automatic rules: Scott's and Silverman's bandwidth selectors

**Advanced features:**
- Custom kernels via `BaseDistribution` objects (experimental)
- Weighted support points with automatic normalization
- Reproducible sampling via `random_state` parameter
- Support for scalar and 2D array distributions

**Implementation notes:**
- Efficient `O(n*m)` evaluation for n support points and m query points
- Compatible with sklearn's `KernelDensity` (verified via parity tests)
- Comprehensive test coverage including edge cases and all kernel types

No new dependencies. Uses existing scipy.special.erf for Gaussian kernel CDF.

- [ ] Design of the custom kernel interface (`BaseDistribution` as kernel parameter)
- [ ] Handling of distribution-based kernels vs built-in string kernels
- [ ] Parameter validation and error messages
- [ ] Test coverage and edge cases

Yes, comprehensive test suite added:
- PDF normalization for all kernel types
- Mean/variance correctness with law of total variance
- CDF monotonicity and limits
- Sample shape for scalar and 2D distributions
- Parameter validation (bandwidth, weights, kernel type)
- Random state reproducibility
- Scott and Silverman bandwidth rules
- Custom distribution kernels (experimental feature)
- Sklearn parity test for Gaussian kernel
- Edge cases (single support point, zero bandwidth, etc.)

#### Any other comments?

This implementation provides a foundation for kernel density estimation in skpro, enabling nonparametric density estimation for distribution fitting tasks. The vectorized design ensures good performance for typical use cases (hundreds to thousands of support points).

Future enhancements could include:
- Additional kernel types (triangular, biweight, triweight)
- Adaptive bandwidth selection
- Multivariate kernel density estimation

### DIFF
--- a/skpro/distributions/tests/test_kernel_mixture.py
+++ b/skpro/distributions/tests/test_kernel_mixture.py
@@ -7,7 +7,6 @@ __author__ = ["amaydixit11"]
 import numpy as np
 import pandas as pd
 import pytest
-
 from skbase.utils.dependencies import _check_soft_dependencies
 
 from skpro.distributions.kernel_mixture import KernelMixture


fkiraly (changes requested):
Very nice! Do I understand though that in a 2D distribution, all distributions share the same support? This feels like a bit of a restriction. Would it be possible to allow different supports?

Also, an example with different weights per location would be great.

fkiraly:
PS: code formatting fails, please check

skpro/distributions/kernel_mixture.py
```diff
@@ -0,0 +1,528 @@
...
+    ... )
+
+    Using weighted support points:
+
+    >>> km_weighted = KernelMixture(
+    ...     support=[0.0, 1.0, 2.0],
+    ...     bandwidth=0.5,
+    ...     weights=[0.1, 0.3, 0.6],
+    ... )
+    >>> pdf_val = km_weighted.pdf(1.0)
```
amaydixit11, line 109: I’ve added an example using non-uniform weights to the docstring and ensured it’s covered by tests.

amaydixit11:
Hi, yes, currently here for ndim > 0 the support is shared across all marginals/locations. This was a deliberate initial restriction to keep the design simple and fully vectorized, and to stay close to the standard KDE setting (one empirical support, multiple evaluation locations).

Allowing different supports per location would essentially move this towards a “batch of independent KDEs” or a more general mixture-of-mixtures setup, which would require:

- storing support as an array/list per marginal,
- broadcasting weights/bandwidth accordingly,
- and more complex vectorization logic.

I agree this would be useful, especially for conditional or location-specific density estimation. My thought was to treat this as a natural follow-up extension once the core object stabilizes.

If you think supporting per-location supports should be part of the initial scope, I’m happy to explore a design for that (e.g., support as array-like of shape (n_locations, n_support)), but I wanted to avoid overcomplicating the first implementation.

Happy to iterate based on your preference here.

fkiraly:
I see, thanks. I would be happy with shared support in the first draft, if we think carefully about how the API could be naturally extended to also have different supports, e.g., a different type for the support param. We do not want to prevent the futuer extension

skpro/distributions/kernel_mixture.py
fkiraly: thanks

fkiraly:
doctests are failing because a float is returned - some versions of python will cause this to be printed as `2.0`, some as `np.float64(2.0)`. This is not easily solvable, and it will fail on one set of python versions or the other, so I would suggest, avoid printing out a float in the docstring tests.

amaydixit11:
Thanks for the confirmation, glad that shared support is fine for the first draft.

I've thought about the forward-compatibility story for the `support` parameter, and I think the cleanest extension path is type/shape-based dispatch:

- **Current (1D):** `support` is a 1D array-like of shape `(n_support,)`, this is shared across all marginals. This is what's implemented now.
- **Future (2D):** `support` could accept a 2D array of shape `(n_locations, n_support)` where each row is the support for a specific marginal, or a list of 1D arrays for variable-length per-location supports.

Also, the current implementation calls `np.asarray(support).ravel()` unconditionally, which would silently flatten a 2D input rather than dispatching on it. To enable the future extension, we'd need to remove the `ravel()` call and add an explicit shape check at init to branch on 1D vs 2D, but the external API contract would stay the same and the 1D path remains unchanged, so it's a clean and purely additive change.

For `weights`, the same pattern applies: currently 1D `(n_support,)`, future could be 2D `(n_locations, n_support)` or a list of weight arrays.

Does this design direction look right to you, or would you prefer a different type convention for the future case?

If it's okay I can raise another issue for the 2D support and work on it in a separate PR as a follow-up.

amaydixit11:
@fkiraly can you let me know if more changes are required?

fkiraly:
I am not sure if I get what you mean.

The current design for `skpro` only knows 0D and 2D distributions, for simplicity. Hence you do not need to think about, or support, 1D.

Whereas, if each entry in the distribution could have different support, you would be looking at a 2D structure of 1D arrays which can have different length, i.e., a 3D array that is ragged in the last dimension (not supported by numpy in this form).

amaydixit11:
Right, 
In that case, I agree, if we allowed different supports per entry in the 2D case, we’d effectively end up with a 2D grid where each cell carries its own support array, which becomes a ragged 3D structure. That wouldn’t sit well with NumPy.

Given that skpro only distinguishes 0D and 2D distributions, keeping the support shared across the 2D distribution feels consistent with the existing abstraction and keeps things clean.

I’m happy to stick with the shared-support design for this PR.
Let me know if any changes are needed

Issue #661, opened 2025-11-27
[ENH] `random_state` for `BaseDistribution.sample`


fkiraly:
It might be good to have a `random_state` parameter in the `sample` method of `BaseDistribution` and all children, to ensure that sampling with a random seed is possible.

Open questions:
* should the parameter be there, in `sample`? Alternative locations are the config via `set_config`, or the `__init__`.
* how to handle composite distributions? They would need to generate dependent `random_state` for components, how should that be done in a reliable way? Sample the `random_seed` uniformly from [0,1] or the full domain?

Before implementing, I would appreciate input on the design questions.

fkiraly:
FYI @joshdunnlime, @SimonBlanke, @tingiskhan

Aymuos22:
Hey @fkiraly , is it okay if I take this on? Plan would be:

- add random_state=None to BaseDistribution.sample (and every override), normalize it via a helper that returns a np.random.Generator, and swap all direct np.random calls over to that generator;

- extend stochastic helper methods (mean, var, energy, cdf, pdfnorm, etc.) with the same optional random_state, forwarding it to the internal sample calls;

- give composite/delegate distributions (mixtures, IID, transforms, TF adapters) a small utility to spawn child generators so component samples stay deterministic.

Please let me know if you’re happy with that approach, and I’ll put a PR together.

Aymuos22:
Hi @fkiraly 
I've tried to resolve the issue

if possible can you check:
https://github.com/sktime/skpro/pull/662

fkiraly:
> extend stochastic helper methods (mean, var, energy, cdf, pdfnorm, etc.) with the same optional random_state, forwarding it to the internal sample calls;

I see, the issue is that many methods have to be extended if we use a Monte Carlo approximation.

Would in this case a single `random_state` in the `__init__` not be a better design, since it avoids too many changes? This is also closer to how `random_state` is managed in `scikit-learn`.

Please note that above I said: "Before implementing, I would appreciate input on the design questions." - of course your prototype is appreciated, @Aymuos22, but we cannot guarantee that it would be the final design.

Aymuos22:
Hi @fkiraly 

Got that, adding random_state to every Monte Carlo call was getting noisy. I’m good with the sklearn-style approach: accept random_state in __init__, keep an internal generator, and allow optional per-call overrides. if you approve it, I’ll update the prototype, or hold off until we lock in the final design.

Aymuos22:
I've modified the prototype

[...]

PR #722, opened 2026-02-19
[ENH] Add random_state parameter to BaseDistribution.sample for reproducible sampling

moksha-hub:
## Summary
Adds a `random_state` parameter to `BaseDistribution.sample()` and the private [_sample()](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1637:4-1661:76) so sampling is reproducible, as requested in #661.

## Changes
- [sample(n_samples=None)](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1604:4-1635:75) → [sample(n_samples=None, random_state=None)](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1604:4-1635:75)
- [_sample(n_samples=None)](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1637:4-1661:76) → [_sample(n_samples=None, random_state=None)](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1637:4-1661:76)
- Uses `sklearn.utils.check_random_state(random_state)` internally
- Backward compatible — calling [sample()](cci:1://file:///c:/Users/moksh/OneDrive/Desktop/Esoc/skpro/skpro/distributions/base/_base.py:1604:4-1635:75) with no args unchanged

## Usage
```python
d = Normal(mu=0, sigma=1)
s1 = d.sample(100, random_state=42)
s2 = d.sample(100, random_state=42)
# s1 == s2 — reproducible!

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -1602,13 +1602,21 @@ class BaseDistribution(BaseObject):
         quantiles = qres.loc[:, cols]
         return quantiles
 
-    def sample(self, n_samples=None):
+    def sample(self, n_samples=None, random_state=None):
         """Sample from the distribution.
 
         Parameters
         ----------
         n_samples : int, optional, default = None
             number of samples to draw from the distribution
+        random_state : int, RandomState instance, or None, optional, default = None
+            Controls the random number generator used for sampling.
+
+            * If ``int``, the value is used to seed the ``RandomState``.
+            * If ``RandomState`` instance, it is used directly.
+            * If ``None``, the global ``numpy`` random state is used.
+
+            Pass an integer for reproducible results across multiple calls.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


arnavk23 (commented):
this changes the subclass interface. `sample()` now calls `_sample(..., random_state=...)`, which will **break any existing distribution subclass** that still defines `_sample(self, n_samples=None)` (TypeError for unexpected kwarg), so it’s not actually backward compatible.  

Could you adjust the design to avoid breaking subclasses (e.g., keep `_sample` signature stable or allow `**kwargs` / separate RNG helper), and also consider avoiding the new `sklearn` dependency/import in the sampling path (prefer NumPy RNG utilities if possible)?

ghost:
Hi @fkiraly and @moksha-hub,

Since #722 has been open for a few weeks with the backward-compatibility concern from @arnavk23 unresolved, I'd like to propose a fix and submit a fresh PR if that's welcome.

The core issue: adding `random_state` to `_sample()` breaks all existing subclasses that define `_sample(self, n_samples=None)` — they raise `TypeError` on the unexpected kwarg.

My proposed approach keeps `_sample()` signature **completely unchanged**:
- Add `random_state=None` to the public `sample()` only
- Normalize it to `np.random.Generator` via a small helper (pure numpy, no sklearn dependency)
- Temporarily inject it as `self._sample_rng` via a context manager before calling `_sample()`
- Base `_sample()` reads `self._sample_rng` if set, falls back to `np.random` otherwise
- `Empirical._sample()` gets a one-line update to respect `self._sample_rng`

This means zero changes to any existing subclass signature. Happy to submit a PR with full tests if this design looks good to you.

PR #724, opened 2026-02-19
[ENH] Fix 4-100x slowdown in _get_bc_params_dict for non-native index types

### LINKED ISSUE #597: [BUG] Repeated calls to `to_numpy` in the Distribution energy method cause notable slow down

moksha-hub:
## Summary
Fixes the 4–100x performance slowdown in `BaseDistribution._get_bc_params_dict` for tz-aware `DatetimeIndex`, `PeriodIndex`, `IntervalIndex` and other non-native numpy index types, as reported in #597.
## Root Cause
`pandas.Index.to_numpy()` is extremely slow for non-native numpy index types. This call happened on every distribution method call.
## Fix
Replace `.to_numpy()` with `np.array()` for index/columns conversion. This avoids the slow pandas conversion path while producing equivalent results.
## Speedup
| Index Type | Before | After |
|---|---|---|
| tz-aware DatetimeIndex | 19.2s | 4.6s (~4x) |
| IntervalIndex | ~12x slower | baseline |
Closes #597

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -619,9 +619,12 @@ class BaseDistribution(BaseObject):
         kwargs_as_np = {k: row_to_col(np.array(v)) for k, v in kwargs.items()}
 
         if hasattr(self, "index") and self.index is not None:
-            kwargs_as_np["index"] = self.index.to_numpy().reshape(-1, 1)
+            # Use .values instead of .to_numpy() to avoid 4-100x slowdown for
+            # non-native numpy index types (tz-aware DatetimeIndex, PeriodIndex,
+            # IntervalIndex, etc.). See skpro issue #597.
+            kwargs_as_np["index"] = np.array(self.index).reshape(-1, 1)
         if hasattr(self, "columns") and self.columns is not None:
-            kwargs_as_np["columns"] = self.columns.to_numpy()
+            kwargs_as_np["columns"] = np.array(self.columns)
 
         bc_params = self.get_tags()["broadcast_params"]
         if bc_params is None:


felipeangelimvieira:
Thank you for contibuting @moksha-hub! I see that there is already another developer working on that issue (#715). For that reason, we should focus on that PR, where you can also add comments and review, or search for another issue where no contributor is working on currently.

PR #741, opened 2026-02-21
ENH: Enforce non-negative weights in Empirical distribution

Divy6504:
This PR enforces non-negativity of weights in the Empirical distribution.

Changes:
- Added validation in Empirical.__init__ to raise ValueError if weights contain negative values.
- Added regression test to test_empirical.py.

Rationale:
Empirical distributions represent probability measures, and probability weights must be non-negative.

All existing tests pass locally (pytest -n 0 --no-cov -q).

### DIFF
--- a/skpro/distributions/empirical.py
+++ b/skpro/distributions/empirical.py
@@ -215,9 +215,11 @@ class Empirical(BaseDistribution):
 
         weights_values = []
         for spl_ix in self._spl_indices:
-            weights_i = self.weights.loc[spl_ix]
-            weights_i = weights_i.loc[self.index]
-            weights_values.append(np.asarray(weights_i))
+            weights_i = self.weights.loc[spl_ix].loc[self.index]
+            arr = np.asarray(weights_i)
+            if np.any(arr < 0):
+                raise ValueError("Weights cannot be negative.")
+            weights_values.append(arr)
 
         self._weights_array = np.stack(weights_values, axis=0)
         return self._weights_array
--- a/skpro/distributions/tests/test_empirical.py
+++ b/skpro/distributions/tests/test_empirical.py
@@ -8,6 +8,23 @@ from skpro.distributions.empirical import Empirical
 from skpro.tests.test_switch import run_test_module_changed
 
 
+def test_empirical_negative_weights():
+    """Negative weights should not be allowed in a probability distribution."""
+    spl_idx = pd.MultiIndex.from_product([[0, 1], [0, 1]], names=["sample", "time"])
+    spl = pd.DataFrame(
+        [[0], [100], [1], [200]],
+        index=spl_idx,
+        columns=["x"],
+    )
+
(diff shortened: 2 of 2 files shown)


Divy6504:
@fkiraly ,Thanks for your review .
I was reviewing the code again carefully after your comment to make sure I fully understood the execution paths.

In the estimator workflow, weights passed to Empirical are constructed via _get_empirical_weighted_sample(). Since alpha is validated in _check_alpha (0 < alpha < 1), the derived weights are mathematically guaranteed to be non-negative. So the internal usage path is safe and does not require additional validation.

However, Empirical is part of the public API and can also be constructed manually, e.g.:
Empirical(spl, weights=custom_weights)

I checked Empirical.__init__, _get_weights_array(), _sample(), and the base distribution classes — there is currently no validation preventing negative weights in this manual path.
While this is likely rare, it could happen in advanced workflows. 

For example:
A user computing importance weights numerically:
weights=np.array([0.3,-1e-10,0.7])
Empirical(spl,weights=weights)
Or weights coming from an optimization routine with solver tolerances:
weights = [0.3, -1e-9, 0.4, 0.3]

In such cases, the empirical distribution invariant
would technically be violated, and currently this would not be caught anywhere in the class.
That said, I agree that placing the check directly in __init__ may introduce unnecessary overhead, since Empirical can be constructed multiple times internally (e.g., via _iloc).

Would it make more sense to move the non-negativity validation into _get_weights_array() instead? That would:
1) Keep estimator workflows unaffected
2) Perform validation lazily only when weights are actually materialized
3) Execute once per instance due to caching

If you feel that the negative weights scenario is unlikely then i will revert the change.
Please let me know your preference.

Divy6504:
Following up on my previous comment—I’ve since verified the implementation logic locally to trace the execution and address the overhead concerns also as mentioned. 

I found that placing the validation inside _get_weights_array(self), specifically after the self.ndim == 0 check, successfully protects the public API while maintaining performance. 

This structure ensures that scalar distributions (the most common simple case) bypass the O(n) validation entirely via the early return. For vectorized distributions, the check runs exactly once before the stacking loop, with the result then stored in the cache.

Compared to putting the check in __init__, this approach significantly reduces performance overhead because it avoids running the validation every time an object is sliced or metadata is accessed via _iloc.

 The check only triggers when the weights are actually materialized for calculation. If this approach aligns with your view on balancing public API safety with efficiency, I can finalize the PR with the corresponding test cases and attach it for review.

Proposed Enhancement-
```
def _get_weights_array(self):
        """Return cached weights array of shape (n_samples, n_instances)."""
        if self.weights is None or self.ndim == 0:
            return None
        if self._weights_array is not None:
            return self._weights_array

        weights_values = []
        for spl_ix in self._spl_indices:
            weights_i = self.weights.loc[spl_ix].loc[self.index]
            arr = np.asarray(weights_i)
            if np.any(arr < 0):
                raise ValueError("Weights cannot be negative.")
            weights_values.append(arr)

        self._weights_array = np.stack(weights_values, axis=0)
        return self._weights_array

PR #742, opened 2026-02-21
[BUG] re-fit state leakage and missing censoring in GridSearch

ashnaaseth2325-oss:
## Summary

This PR fixes three related lifecycle and refit correctness issues in the probabilistic regression base classes and grid search logic.

The affected areas are:

* `skpro/regression/base/_base.py`

  * `BaseProbaRegressor.fit`
  * `BaseProbaRegressor._check_X`
* `skpro/model_selection/_tuning.py`

  * `BaseGridSearch._fit`

##Re-fit incorrectly validates feature names (predict-time guard running in fit)

`_check_X()` is shared between `fit()` and all predict variants. It uses a single sentinel:

```python
if hasattr(self, "feature_names_in_"):
    # validate columns
else:
    self.feature_names_in_ = X_feature_names
    self.n_features_in_ = len(X_feature_names)
```

After the first `fit()`, `feature_names_in_` is set.
On any subsequent `fit()` call:

* If columns differ → raises a `ValueError`
* If columns match → metadata is silently frozen to the first fit

The error message even says:

```
"X in predict methods must have same columns as X in fit"
```

…but this branch is executed during `fit()`.

This violates the expected sklearn contract where every `fit()` is a fresh training run.

---

##Converter stores are never reset on re-fit

The following stores are initialized in `__init__`:

```python
self._X_converter_store = {}
self._y_converter_store = {}
self._C_converter_store = {}
```

However, they are never cleared inside `fit()`.

The `convert()` function switches behavior based on whether a store is empty:

* Empty → `"update"` mode
* Non-empty → `"freeze"` mode (deep-copy, no updates)

After the first fit, stores become non-empty.
On second fit:

* Store is frozen
* New metadata (e.g., y column names) is not written back
* Predictions may restore stale round-trip metadata from the first fit

This can silently corrupt output formatting after re-fitting.

---

##Survival GridSearch refits without censoring (`C`)

Inside `BaseGridSearch._fit()`:

```python
# Correct
self.best_estimator_.fit(X, y, C=C)

# Incorrect (before fix)
estimator.fit(X, y)
```

When `return_n_best_estimators > 1`, the alternative estimators were refitted without `C`.

For survival estimators, this silently drops censoring information — producing incorrect models without any error.

---

# Fix

The fix follows sklearn’s estimator lifecycle contract: **each `fit()` must behave like a clean training run.**

##A + B — Reset state at the start of `fit()`

At the top of `BaseProbaRegressor.fit()`:

```python
# Reset feature metadata for clean re-fit
if hasattr(self, "feature_names_in_"):
    del self.feature_names_in_
if hasattr(self, "n_features_in_"):
    del self.n_features_in_

# Reset converter stores
self._X_converter_store = {}
self._y_converter_store = {}
self._C_converter_store = {}
```

Effects:

* `_check_X()` now properly sets feature names on every fit
* Re-fitting with different columns works as expected
* Converter stores update correctly on every fit
* No frozen metadata across training runs

---

##C — Pass censoring data in GridSearch refit loop

In `_tuning.py`:

```python
# Before
estimator.fit(X, y)

# After
estimator.fit(X, y, C=C)
```

Now:

* `best_estimator_`
* All `n_best_estimators_`

are refit consistently with censoring data.

---

# Verification

The following scenarios were tested:

### Re-fit with different feature columns

```python
reg.fit(X1, y1)
reg.fit(X2, y2)  # now succeeds
```

* No ValueError
* `feature_names_in_` reflects `X2`
* Metadata matches latest fit

---

### Converter store refresh

* First fit with numpy `y`
* Second fit with DataFrame `y`
* `_y_converter_store` correctly reflects second fit metadata
* Predictions restore correct column names

---

### Survival GridSearch correctness

With:

```python
gscv = GridSearchCV(..., return_n_best_estimators=2)
gscv.fit(X, y, C=C)
```

Now:

* `best_estimator_` uses `C`
* All `n_best_estimators_` use `C`
* No silent uncensored refits

---

# Impact

Before:

* Re-fitting could crash or freeze stale feature metadata
* Converter stores could silently restore wrong column names
* Survival grid search returned incorrectly refitted models

After:

* `fit()` is idempotent and lifecycle-safe
* Metadata always reflects the most recent training data
* GridSearch survival behavior is consistent and correct

---
This change is small, localized, and aligns estimator behavior with sklearn expectations while fixing silent correctness issues in both regression and survival workflows.


ashnaaseth2325-oss:
Hi @felipeangelimvieira @fkiraly , this PR resets fit-time state to ensure clean re-fitting (feature metadata and converter stores) and fixes missing `C` during `n_best_estimators_` refit in `GridSearch`. The changes are small and localized , happy to add tests or refine further if needed, thanks for reviewing!

fkiraly:
superseded by https://github.com/sktime/skpro/pull/747

PR #746, opened 2026-02-22
[BUG] `BaseGridSearch`: Ensure censoring data (C) is passed in all refit loops

ashnaaseth2325-oss:
## Summary

This PR fixes a refit inconsistency in `BaseGridSearch._fit_best` inside:

```
skpro/model_selection/_tuning.py
```

During refitting:

* `best_estimator_` was correctly refit using:

```python
estimator.fit(X, y, C=C)
```

* However, estimators stored in `n_best_estimators_` were refit using:

```python
estimator.fit(X, y)
```

This caused survival estimators inside `n_best_estimators_` to be refit **without censoring information**, producing silently incorrect models.

This PR ensures that censoring data (`C`) is forwarded consistently across all refit paths.

The fix applies to both:

* `GridSearchCV`
* `RandomizedSearchCV`

since both rely on `BaseGridSearch._fit_best`.

---

## Steps to Reproduce

1. Use a survival estimator that requires censoring data.
2. Run grid search with multiple returned estimators:

```python
gscv = GridSearchCV(
    estimator=survival_estimator,
    param_grid=...,
    return_n_best_estimators=2
)

gscv.fit(X, y, C=C)
```

3. Inspect:

   * `gscv.best_estimator_`
   * `gscv.n_best_estimators_[0][1]`

### Before Fix

* `best_estimator_` receives `C`
* `n_best_estimators_` models are refit **without `C`**
* No error is raised
* Predictions may differ silently due to missing censoring information

---

## Impact

This bug affects survival workflows where censoring data is required for correct likelihood estimation.

Consequences:

* Models in `n_best_estimators_` are incorrectly trained
* Downstream evaluation or ensembling may produce inconsistent results
* No exception is raised → silent correctness issue

This is particularly problematic in research or benchmarking scenarios where multiple top models are compared.

---

## Fix

The refit loop in `_fit_best` has been aligned so that all estimators receive `C`.

### Before

```python
if self.refit:
    estimator.fit(X, y)
```

### After

```python
if self.refit:
    # Pass C so survival estimators receive censoring data,
    # matching the best_estimator_ refit above
    estimator.fit(X, y, C=C)
```

This ensures:

* `best_estimator_`
* All entries in `n_best_estimators_`

are refit consistently with censoring data.

The change is localized and does not alter behavior for non-survival estimators.

---

## Result

After this fix:

* Refit behavior is consistent across all returned estimators
* Survival estimators always receive censoring data
* No silent loss of censoring information
* Behavior between `best_estimator_` and `n_best_estimators_` is now aligned

All existing tests pass, and no new dependencies were introduced.

---

## What reviewers should focus on

* Consistency between refit paths
* Correct forwarding of optional `C`
* Confirmation that `RandomizedSearchCV` is also covered
* No behavioral change for estimators that do not use `C`

### DIFF
--- a/skpro/model_selection/_tuning.py
+++ b/skpro/model_selection/_tuning.py
@@ -216,8 +216,6 @@ class BaseGridSearch(_DelegatedProbaRegressor):
             estimator = self.estimator.clone().set_params(**params)
             # Refit model with best parameters.
             if self.refit:
-                # Pass C so survival estimators receive censoring data, matching
-                # the best_estimator_ refit above (line 202).
                 estimator.fit(X, y, C=C)
             self.n_best_estimators_.append((rank, estimator))
             # Save score


ashnaaseth2325-oss:
Hi @fkiraly @felipeangelimvieira, this is the first split from #742 as suggested. It ensures all estimators in n_best_estimators_ receive C during refit, avoiding silent loss of censoring information in survival workflows.

PR #750, opened 2026-02-23
[BUG] Fix `_check_C` using wrong converter store `_y_converter_store` -> `_C_converter_store`

### LINKED ISSUE #749: [BUG] _check_C uses _y_converter_store instead of _C_converter_store
In **BaseProbaRegressor._check_C** (skpro/regression/base/_base.py, line 862), the censoring indicator C is converted using the wrong converter store.

```
C_inner = convert(
    obj=C,
    from_type=metadata["mtype"],
    to_type=C_inner_mtype,
    as_scitype="Table",
    store=self._y_converter_store,  # incorrect
)
```

__check_C_ writes to _self._y_converter_store_ instead of _self._C_converter_store_.

The comment above the line (# convert y to y_inner_mtype) also suggests this was copied from _check_y and not fully adapted.

**To Reproduce**

This occurs when fitting a survival regressor where y and C have different mtypes (e.g., y as pd.DataFrame and C as numpy.ndarray).

Example:

```
import pandas as pd
import numpy as np
from skpro.survival.base import BaseSurvReg

#Any survival regressor (e.g., CoxPH adapter)
#Using different mtypes for y and C triggers shared converter state
```

Because _check_C writes into _y_converter_store, both y and C share the same converter dictionary.

**Expected behavior**

_check_C should write to self._C_converter_store, consistent with _check_X and _check_y.

**Correct implementation:**

```
- store=self._y_converter_store,
+ store=self._C_converter_store,
```
**Environment**

```
skpro: current main branch
File: skpro/regression/base/_base.py
Affects: survival regressors (capability:survival=True)
```

**Additional context**

__init__ allocates three separate stores:

```
self._X_converter_store = {}
self._y_converter_store = {}
self._C_converter_store = {}

```
However,

 **_C_converter_store**

 is never written to.

This causes y and C to share inverse-transform state, which can silently produce incorrect round-trip conversions when their mtypes differ.

This is a one-line bug fix. I am happy to submit a PR.

kindler-king:
**Reference Issues/PRs**

Fixes #749

**What does this implement/fix? Explain your changes.**
This PR fixes a bug in BaseProbaRegressor._check_C.

Previously, the censoring indicator C was converted using self._y_converter_store instead of the dedicated self._C_converter_store. As a result, y and C could share the same converter dictionary, potentially corrupting inverse-transform state when their mtypes differ (e.g., pd.DataFrame vs ndarray). This could lead to silent round-trip conversion errors in survival regressors.

The fix changes the converter store used in _check_C to self._C_converter_store.

Additionally, this PR corrects a stale copy-paste comment inside _check_C that incorrectly referenced converting y.

**Does your contribution introduce a new dependency? If yes, which one?**

- Correctness of the store replacement in _check_C
- Consistency with _check_X and _check_y
- Confirmation that survival estimators behave correctly when y and C have different mtypes

**Did you add any tests for the change?**
No new tests were added. This is a one-line bug fix aligning _check_C with the existing design pattern used in _check_X and _check_y.

**Any other comments?**
This is a minimal, targeted bug fix with no API changes and no behavioral changes outside survival regression.

### DIFF
--- a/skpro/regression/base/_base.py
+++ b/skpro/regression/base/_base.py
@@ -852,14 +852,14 @@ class BaseProbaRegressor(BaseEstimator):
         if not valid:
             check_is_error_msg(msg, var_name="C", raise_exception=True)
 
-        # convert y to y_inner_mtype
+        # convert C to C_inner_mtype
         C_inner_mtype = self.get_tag("C_inner_mtype")
         C_inner = convert(
             obj=C,
             from_type=metadata["mtype"],
             to_type=C_inner_mtype,
             as_scitype="Table",
-            store=self._y_converter_store,
+            store=self._C_converter_store,
         )
 
         return C_inner, metadata


fkiraly:
Correct bugfix but likely AI generated.

kindler-king:
> Correct bugfix but likely AI generated.

Hey @fkiraly the PR request was written using AI for sure, but the bug fix was a 1 line fix that I randomly found while going through the code. Hope the context helps :D
I am working on identifying more issues and will definitely make sure not to rely on AI tools.

PR #754, opened 2026-02-23
[ENH] Cauchy probability distribution

patelchaitany:
#### Reference Issues/PRs

#22 

This PR adds the Cauchy probability distribution.

PDF :  $$f(x) = \frac{1}{\pi \gamma \left[1 + \left(\frac{x - x_0}{\gamma}\right)^2\right]}$$

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -35,6 +35,7 @@ Continuous support - full reals
     :toctree: auto_generated/
     :template: class.rst
 
+    Cauchy
     Laplace
     Logistic
     Normal
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -9,6 +9,7 @@ __all__ = [
     "Binomial",
     "BurrIII",
     "BurrXII",
+    "Cauchy",
     "ChiSquared",
     "Delta",
     "Empirical",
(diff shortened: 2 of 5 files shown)


marrov:
@patelchaitany - CI test suite is failing, which shows your PR has issues at the moment. Make sure to debug the test failures locally and apply the fixes. If you are unfamiliar with this approach, you can read up on it here: https://www.sktime.net/en/stable/developer_guide/continuous_integration.html

patelchaitany:
@marrov Tests are failing for the Cauchy distribution because it has no defined mean or variance and It get NaN from both skpro and scipy.

The line np.allclose(res, scipy_res) fails since NaN != NaN in NumPy by default.

Quick fix: change it to (test_scipy_adapter.py : test_method_no_params)

    np.allclose(res, scipy_res, equal_nan=True)

Let me know what you think about.

fkiraly:
would this be solved by setting mean and variance to `np.inf`? Technically that is the correct value, since the integrals are diverging to plus infinity.

patelchaitany:
@fkiraly Cauchy mean, the integral doesn't diverge to +inf as the positive and negative tails both diverge to +inf and -inf , giving inf - inf, which is undefined, not infinite. for variance i agree as it diverge to infinity.

fkiraly:
Yes, you are right, I meant the variance, which could be `np.inf`.

For mean, indeed, the integral is undefined. I see two workarounds:

* Allow `np.nan` and update the tests. This could be done programmatically with a tag (e.g., a list of tags that are undefined, similar to the lists about exact/approximate)
* set the `mean` to be the same as location, as a hacky workaround. This is not entirely wrong if you define the mean a certain way (but it is wrong in the strit sense)

patelchaitany:
@fkiraly Updated the test and _tags.py to include the capabilities:undefined tag.

Also extended the inf vs nan guard in the scipy adapter test to handle arrays  the existing check only covers scalars, but Cauchy has array shaped test params so we need the same logic for DataFrame and array results too.

PR #760, opened 2026-02-25
[BUG] Fix ValueError in _check_list_of_dict_table due to NumPy 2.0 generator handling

kabirvashisht4-glitch:
Reference Issues/PRs
Fixes a crash in list-of-dict validation caused by NumPy 2.0 handling of generators.

The validator _check_list_of_dict_table was crashing with a ValueError: Calling nonzero on 0d arrays is not allowed.

This happened because np.where was receiving a generator expression. In modern NumPy (2.0+), this results in a 0-dimensional object instead of an array-like object.

Changes:

Replaced the generator expression with a list comprehension [...] to ensure an array is passed to np.where.

Added [0] to the np.where call to correctly extract the array of indices from the returned tuple.

No.

Verify if the use of list comprehension within the f-string is the preferred style for index extraction in this specific check.

I verified the fix locally using a list-of-dict data structure that previously triggered the error.

Any other comments?
This is my first contribution to skpro as part of my GSoC 2026 preparation!

### DIFF
--- a/skpro/datatypes/_table/_check.py
+++ b/skpro/datatypes/_table/_check.py
@@ -459,7 +459,8 @@ def _check_list_of_dict_table(obj, return_metadata=False, var_name="obj"):
     if not np.all([isinstance(x, dict) for x in obj]):
         msg = (
             f"{var_name} must be a list of dict, but elements at following "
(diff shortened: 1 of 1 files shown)


kabirvashisht4-glitch:
@fkiraly , could you please review this PR

PR #761, opened 2026-02-25
Fix dummy variance formula and make survival adapter reflection safe by using vars() instead of dir()

Divy6504:
Earlier i had opened a pull request regarding the variance formula change but it was failing certain CI tests due to an issue in scikit-survival package. 
So i have closed the previous pr opened a new PR which fixes two issues. First, it corrects the variance computation in DummyProbaRegressor to ensure that the returned variance corresponds to σ² rather than σ which was opened in the prevoious pr as well. Second, it resolves a CI failure affecting survival estimators (SurvivalForestSkSurv and SurvivalForestXtraSkSurv) under scikit-survival >= 0.27.0 and scikit-learn >= 1.8.0. The failure was caused by unsafe reflection in _get_fitted_params_default_safe, which previously relied on dir(obj). Under newer versions of scikit-learn, BaseEstimator.__dir__() internally calls hasattr() and getattr(), which in turn evaluate properties such as feature_importances_ from scikit-survival.

This property raises NotImplementedError, causing the VM test to fail before attribute filtering occurs.I have updated the implementation to use vars(obj) instead, which directly accesses obj.__dict__ and avoids invoking __dir__, hasattr, or property evaluation. This preserves the intended behavior of collecting fitted attributes (which follow the sklearn convention of ending with _) while preventing the crash. The fix is backward-compatible, as fitted parameters are stored in the instance dictionary in both older and newer dependency versions. Also,I have validtaed the changes locally under scikit-learn 1.7.x with scikit-survival 0.26.x as well as scikit-learn 1.8.x with scikit-survival 0.27.0, including full pytest execution and the VM test harness, all of which passed successfully along with pre commit formatting

### DIFF
--- a/skpro/regression/dummy.py
+++ b/skpro/regression/dummy.py
@@ -126,7 +126,7 @@ class DummyProbaRegressor(BaseProbaRegressor):
         X_ind = X.index
         X_n_rows = X.shape[0]
         y_pred = pd.DataFrame(
-            np.ones(X_n_rows) * self._sigma, index=X_ind, columns=self._y_columns
+            np.ones(X_n_rows) * (self._sigma**2), index=X_ind, columns=self._y_columns
         )
         return y_pred
 
--- a/skpro/regression/tests/test_all_regressors.py
+++ b/skpro/regression/tests/test_all_regressors.py
@@ -1,10 +1,12 @@
 """Automated tests based on the skbase test suite template."""
+import numpy as np
 import pandas as pd
 import pytest
 from skbase.testing import QuickTester
 
 from skpro.datatypes import check_is_mtype, check_raise
 from skpro.distributions.base import BaseDistribution
+from skpro.regression.dummy import DummyProbaRegressor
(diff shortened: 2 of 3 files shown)


Divy6504:
I also noticed that some failing jobs appear related to optional dependency installation under Python 3.14 (e.g., cyclic_boosting, ondil version constraints, and a temporary glum download issue).

Since this PR does not modify registry logic or dependency guards, I suspect these may be environment-specific CI issues. However, I'm happy to further investigate compatibility guards for Python 3.14 if needed.

Also do you think the CI failures were primarily triggered by the reflection behavior change, or could the new variance test also be interacting with the registry in an unexpected way?
Also, if you feel the variance test would be better placed elsewhere in the test suite (e.g., a more estimator-specific test file), I’d be happy to restructure it.

fkiraly:
this PR seems to combine unrelated, random changes in different modules, without a corresponding issue. Likely AI generated without user oversight.

Issue #322, opened 2024-05-13
[ENH] Implement Histogram Conditional Density Estimation


ShreeshaM07:
**Describe the solution you'd like**

Histogram estimation is not present in skpro. Implement them from scratch using the conditional density estimate finding the optimal binwidth(h) and find the function that fits the histograms most aptly without over smoothing nor undersmoothing.

**Additional context**

Useful resources
- <a href=https://egrcc.github.io/docs/math/all-of-statistics.pdf>All of statistics(chapter 20.2)</a>
- https://www.youtube.com/watch?v=SUvPJ4URYGA

ShreeshaM07:
@fkiraly Do you recommend any other resources to refer for implementing this?

fkiraly:
Sure!

Some classical ones:
* Hyndman et al - [Estimating and visualizing conditional densities](https://www.tandfonline.com/doi/abs/10.1080/10618600.1996.10474715)
* Efromovich - [Conditional density estimation in a regression setting](https://projecteuclid.org/journals/annals-of-statistics/volume-35/issue-6/Conditional-density-estimation-in-a-regression-setting/10.1214/009053607000000253.short)
* Bashtannyk, Hyndman - [Bandwidth selection for kernel conditional density estimation](https://www.sciencedirect.com/science/article/pii/S0167947300000463?casa_token=msybNuEOctUAAAAA:W-Xt3V3Txqk3ZN06WTpQ7WLtKA5UA4zRwx--7Z5a4zeNFgOeUIUrrs6nwTsHvO0FusMbfZCU)

This is mostly kernel based.

Also, what is `tophat`?

ShreeshaM07:
> Also, what is `tophat`?

Its a type of kernel in `sklearn` KDE
the K(x,h) is proportional to 1 for x<h.

It mostly resembles a bin itself.

fkiraly:
Oh, I see, the "top-hat kernel", which is the same as a box kernel.
It corresponds to a uniform distribution.

Here are a few things I noticed:

* for histogram CDE, we need a histogram distribution - these are not available in `skpro` yet. That might be a simpler implementation item, as it does not hinge on design decisions? Issue here: https://github.com/sktime/skpro/issues/323
* there is also a subtask of representing kernel mixtures, not all of these are distribution mixtures. There are a few design decisions to think about, this is a more difficult issue. Issue opened here: https://github.com/sktime/skpro/issues/324
[...]

PR #735, opened 2026-02-20
[ENH] Add initial Histogram Conditional Density Estimator (CDE)

amaydixit11:
#### Reference Issues/PRs

Fixes #322

---

#### What does this implement/fix? Explain your changes.

This PR introduces an initial implementation of a histogram-based conditional density estimator via `HistogramCDERegressor`.

The estimator performs conditional density estimation by:
- Identifying local neighborhoods in feature space using k-nearest neighbors
- Constructing histograms over target values within each neighborhood
- Returning a `Histogram` distribution at prediction time

The PR adds:
- A new nonparametric regressor in `skpro.regression.nonparametric`
- Integration with the existing `Histogram` distribution
- Basic configuration options (`n_bins_x`, `n_bins_y`, `strategy`)
- A unit test validating output type and normalization

This provides a baseline implementation for histogram-based CDE.

---

#### Does your contribution introduce a new dependency? If yes, which one?

No. This implementation relies only on existing dependencies already used within the project (NumPy, pandas, scikit-learn).

---

#### What should a reviewer concentrate their feedback on?

- Correctness of the locality estimation using KNN
- Compatibility with `BaseProbaRegressor` interface
- Design alignment with planned kernel-based CDE estimators
- API consistency and naming conventions

---

#### Did you add any tests for the change?

Yes.

A unit test (`test_histogram_cde_output`) has been added to verify:
- The returned object is a `Histogram` distribution
- Output shape is correct
- Bin masses are properly normalized

---

#### Any other comments?

This implementation is intended as a baseline for histogram-based conditional density estimation. It is designed to be extended in future work towards kernel-based and adaptive bandwidth methods.

I’m happy to iterate on the design based on reviewer feedback.

---

##### For all contributions
- [x] I've added myself to the list of contributors
- [x] The PR title starts with [ENH]

##### For new estimators
- [x] I've added the estimator to the API reference
- [x] I've added illustrative usage examples to the docstring
- [ ] I've set the `python_dependencies` tag if applicable

PR #772, opened 2026-02-26
[ENH] Nadaraya Watson CDE regressor

patelchaitany:
#### Reference Issues/PRs

Implementing CDE from this road map - #7 

and the Converstion #322 

In this PR I have Implemented the nadaraya watson method. 

Here is High Level Overview of the How dose this is doing point prediction, CDE and weight calculation .

**Weight calculation**: For each test point x, compute the Euclidean distance to every training point, scale by the bandwidth $(u_i = d_i / h)$, apply the kernel function $(u_i)$ to convert distances into similarity scores, and normalize so the weights sum to 1: $w_i(x) = K(u_i) / sum_j K(u_j)$. If all kernel values are zero (no neighbors within support), fall back to uniform weights $w_i = 1/n$.

**Point prediction**: The predicted value is the weighted average of training targets: y_hat(x) = $sum_i w_i(x) * y_i.$ In matrix form this is y_hat = W @ y, where W is the weight matrix from the step above. Training points close to x dominate the sum; distant points contribute little or nothing.

**CDE (Conditional Density Estimation)**: Instead of collapsing to a single mean, the full conditional distribution is returned as a weighted empirical distribution: F_hat(y | x) = $sum_i w_i(x) * delta(y - y_i)$. Each training target y_i is a possible outcome with probability $w_i(x)$. This is wrapped in an Empirical distribution object, from which you can extract the mean, variance, quantiles, CDF, or any other distributional summary.

$delta(y-y_i)$ is dirac delta.

NO

##### For all contributions

##### For new estimators
- [X] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.

### DIFF
--- a/docs/source/api_reference/regression.rst
+++ b/docs/source/api_reference/regression.rst
@@ -234,7 +234,7 @@ Gaussian process and kernel regression
 
     GaussianProcess
 
-Non-parametric regressors
+Non-parametric conditional density estimation
 -------------------------
 
 .. currentmodule:: skpro.regression.nonparametric
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ from skpro.regression.residual import ResidualDouble
 
 # step 1: data specification
 X, y = load_diabetes(return_X_y=True, as_frame=True)
-X_train, X_new, y_train, _ = train_test_split(X, y)
+X_train, X_new, y_train, y_test = train_test_split(X, y)
 
 # step 2: specifying the regressor - any compatible regressor is valid!
 # example - "squaring residuals" regressor
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -4,13 +4,13 @@
 Probability distributions
 =========================
 
-The :mod:`sktime.distributions` module contains
+The :mod:`skpro.distributions` module contains
 probability distributions which combine a ``pandas.DataFrame``-like API
 with a ``scikit-base`` compatible object interface.
 
 All distributions in ``skpro`` can be listed using the ``skpro.registry.all_objects`` utility,
 using ``object_types="distribution"``, optionally filtered by tags.
-Valid tags can be listed using ``sktime.registry.all_tags``.
+Valid tags can be listed using ``skpro.registry.all_tags``.
 
 Base
 ----
@@ -77,6 +77,7 @@ Continuous support - non-negative reals
     InverseGaussian
     LogLaplace
     Pareto
+    Rayleigh
     TruncatedPareto
     Weibull
 
--- a/docs/source/api_reference/metrics.rst
+++ b/docs/source/api_reference/metrics.rst
@@ -9,7 +9,7 @@ probabilistic predictions, including survival and time-to-event predictions.
 
 All metrics in ``skpro`` can be listed using the ``skpro.registry.all_objects`` utility,
 using ``object_types="metric"``, optionally filtered by tags.
-Valid tags can be listed using ``sktime.registry.all_tags``.
+Valid tags can be listed using ``skpro.registry.all_tags``.
 
 Survival/time-to-event specific metrics in ``skpro`` can be listed
 by filtering by ``capability:survival`` being ``True``.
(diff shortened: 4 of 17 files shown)


arnavk23:
@patelchaitany Loved the work done here by you. One suggestion try passing flake+black on your branch or just try precommit files instead on it. It will help you as when we pass workflows generally what is causing problems for distributions is code quality checks.

patelchaitany:
@arnavk23 Thanks a lot for the feedback!

I've taken it seriously and made sure to follow the contribution guidelines more closely this time.

Specifically, I installed and activated the pre-commit hooks (pre-commit install) and Ran the full test suite locally before pushing (make test) and Followed the recommendations in the (https://www.sktime.net/en/stable/developer_guide/continuous_integration.html).

This helped catch a few things early and should make the CI flow smoother.

If I'm still missing any guideline, have overlooked something important, or if there's anything else I should adjust (code style, tests, documentation, etc.), please let me know  I'm happy to fix it right away.

patelchaitany:
@fkiraly @felipeangelimvieira can you pls review it. if any changes needed pls let me know.

docs/source/api_reference/regression.rst
```diff
@@ -223,6 +223,17 @@ Gaussian process and kernel regression
 
     GaussianProcess
 
+Non-parametric regressors
```
fkiraly: I would say "non-parametric conditional density estimation"?

patelchaitany: I saw in this PR #735 it is using Non-parametric regressors so I thought i also follow same naming.

patelchaitany:
@fkiraly I think kernel mixture distribution still dose not exists in the Repo it exists in this PR #721 and For Now I am going with Mixture (which might be inefficient) or wait for other PR to merge ?

docs/source/api_reference/regression.rst
fkiraly: #735 is also CDE, so the same would apply

fkiraly:
ah, thanks for reminding me, @patelchaitany!

I finished https://github.com/sktime/skpro/pull/721 and will merge it soon, then you can use it.

For consistency, I would use `h` instead of `bandwidth`.

fkiraly:
#721 is merged now

patelchaitany:
@fkiraly I have one question, When Calculating the CDE for the more than One Data Point Can I return the Array With the KernelMixture as Object.

fkiraly:
not sure what you mean, could you elaborate and give an example?

patelchaitany:
@fkiraly. Currently, KernelMixture accepts a 1D weights array that is shared across all rows. For example:

KernelMixture(support=[3.0, 5.0, 4.5], weights=[0.5, 0.3, 0.2], h=0.5)

For example, with 3 training targets and 2 test points:

  Test point 0 (close to y=3.0): weights = [0.6, 0.1, 0.3]
  Test point 1 (close to y=5.0): weights = [0.05, 0.8, 0.15]

I need to return a single distribution with index of length 2, where row 0 uses [0.6, 0.1, 0.3] and row 1 uses [0.05, 0.8, 0.15]. Similar to how Normal(mu=[1.0, 2.0], sigma=[0.5, 0.5]) gives a 2 row distribution with different parameters per row.

Currently this isn't possible with KernelMixture because weights is shared across all rows.
Would it make sense to extend KernelMixture to support 2D weights (one weight vector per row)? Or is there another existing approach I should use?

fkiraly:
Ah, that is true - yes, I would recommend to extend `KernelMixture` to support 2D weights. Should not be too hard?

patelchaitany:
@fkiraly I have created PR for extending the KernelMixture to 2D weights #841 pls give it look.

fkiraly:
merged, so you can now use it

docs/source/api_reference/regression.rst
fkiraly: I think just "non-parametric" is too general. I would add a CDE reference.

patelchaitany:
@fkiraly now using the kernel mixture density and also allow user to pass kernel in Y direction.

PR #780, opened 2026-03-01
[BUG] fix transformer chaining in regression pipeline

ashnaaseth2325-oss:
## **Summary**

Fixes a mismatch between `Pipeline._fit` and `_transform` in:

```
skpro/regression/compose/_pipeline.py
```

During `fit`, transformers are chained correctly:

```python
X = t.fit_transform(X=X, y=y)
```

But in `_transform`, `X` was never updated inside the loop. Each transformer received the **original X**, and only the last output was returned.

So a pipeline:

```
T1 → T2 → Regressor
```

was trained on:

```
T2(T1(X_train))
```

but predicted on:

```
T2(X_test)
```

---

## **Impact**

* Multi-transformer pipelines produced incorrect predictions.
* CV and backtesting results could be silently corrupted.
* No error was raised — shapes remained valid.
* Single-transformer pipelines were unaffected (hard to detect).

---

## **Fix**

Add one line inside `_transform`:

```python
X = Xt
```

This properly chains transformer outputs.

---

## **Result**

* Fit and predict now use the same feature representation.
* Multi-step pipelines behave correctly.
* No API changes.

### DIFF
--- a/skpro/regression/tests/test_pipeline_transform_chain.py
+++ b/skpro/regression/tests/test_pipeline_transform_chain.py
@@ -34,4 +34,3 @@ def test_transformer_chaining_in_predict():
 
     assert np.allclose(Xt.values, expected.values)
     assert len(y_pred) == len(y)
-    


ashnaaseth2325-oss:
Hello @fkiraly , @felipeangelimvieira 
This PR fixes an inconsistency where `_transform` wasn’t chaining transformer outputs like `_fit`, causing multi-step pipelines to use incorrect features at predict time.
Added `X = Xt` inside the loop so prediction now mirrors training . Happy to adjust if needed.

ashnaaseth2325-oss:
Hello @fkiraly,
Thanks for the suggestion.
I've updated the loop to use Xt = transformer.transform(Xt) for improved readability and consistency with fit.
All checks are now passing. Happy to adjust further if needed.

ashnaaseth2325-oss:
Hello @fkiraly 
Thanks for the suggestion! I’ve added a test chaining two FunctionTransformer(np.exp) steps with a deterministic estimator and checking that the result matches exp(exp(X)).
 CI is green now. Let me know if anything else should be adjusted.

PR #784, opened 2026-03-02
[ENH] exact `_log_pdf` for `Uniform` distribution

Ashish-Kumar-Dash:
## Reference Issues/PRs

Fixes #782 

## What does this implement/fix? Explain your changes.

Adds the missing `_log_pdf` method to the `Uniform` distribution. The class already declared `"log_pdf"` in its `capabilities:exact` tag but never implemented the method, causing the base class fallback (`np.log(self.pdf(x))`) to be used at runtime. This fallback is numerically unstable for out-of-bounds points (`log(0)` triggers numpy warnings) and contradicts the exact tag.

The new implementation directly computes:

- `-np.log(upper - lower)` for in-bounds points
- `-np.inf` for out-of-bounds points

using `np.where`, which is both exact and numerically clean.

## Does your contribution introduce a new dependency? If yes, which one?

No.

## What should a reviewer concentrate their feedback on?

The `_log_pdf` implementation in uniform.py — single method addition following the same pattern as `Laplace._log_pdf` and `Normal._log_pdf`.

## Did you add any tests for the change?

No new tests were added. The existing `test_log_pdf_and_pdf` test in `test_all_distrs.py` already validates consistency between `pdf` and `log_pdf` for all distributions, including `Uniform`. All 97 existing Uniform tests pass.

## Any other comments?

This is a small, low-risk change — one method addition (~18 lines including docstring) with no changes to existing logic.

### For all contributions

- [x] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
- [x] The PR title starts with either `[ENH]`, `[MNT]`, `[DOC]`, or `[BUG]`. `[ENH]` - adding or improving code.

PR #787, opened 2026-03-03
[ENH] exact `_log_pdf` for `Uniform` distribution

### LINKED ISSUE #782: [ENH] add _log_pdf implementation to Uniform distribution

The `Uniform` distribution declares `"log_pdf"` in its `capabilities:exact` tag, but does not implement the `_log_pdf` method. This causes the base class fallback to be used at runtime, which computes `np.log(self.pdf(x))`. This is numerically unstable for out-of-bounds points (evaluating `log(0)` triggers numpy warnings) and contradicts the `exact` capability tag.

Add a direct `_log_pdf` implementation to the `Uniform` class that returns `-np.log(upper - lower)` for in-bounds points and `-np.inf` for out-of-bounds points, using `np.where`. This is mathematically exact, avoids `log(0)` warnings, and fulfills the already-declared `exact` capability.

- Removing `"log_pdf"` from `capabilities:exact` — this would be accurate but leaves performance and numerical stability on the table for a trivially implementable method.
- Keeping the base class fallback (`np.log(self.pdf(x))`) — this works functionally but produces numpy `RuntimeWarning` for out-of-bounds evaluations and is slower due to the extra division in `_pdf` followed by a `log` call.

**Additional context**

The fix is a single method addition (~18 lines including docstring), following the same pattern used by other distributions (e.g., `Laplace`, `Normal`). All 97 existing `Uniform` tests pass after the change, including `test_log_pdf_and_pdf` which validates consistency between `pdf` and `log_pdf`.

doraem-on:
#### Reference Issues/PRs
Fixes #782 

The `Uniform` distribution declares `"log_pdf"` in its `capabilities:exact` tag, but it didn't actually implement the `_log_pdf` method. It was relying on the base class fallback, which computes `np.log(self.pdf(x))`. 

This caused two problems:
1. It triggers a numpy `RuntimeWarning` when evaluating `log(0)` for out-of-bounds points.
2. It's computationally inefficient because it does a division followed by a log call.

This PR adds a direct `_log_pdf` implementation using `np.where`. It returns exactly `-np.log(upper - lower)` for in-bounds points and `-np.inf` for out-of-bounds points. This completely avoids the warning, fulfills the declared `exact` capability, and is mathematically sound.

No new dependencies.

- [x] Just verifying that the `np.where` logic and `self._bc_params` usage perfectly align with `skpro`'s array broadcasting expectations.

I didn't need to add new tests because the existing test suite already catches this! I ran `pytest skpro/distributions/ -k Uniform` locally, and all 97 tests pass successfully, including the `test_log_pdf_and_pdf` consistency checks.

#### Any other comments?
Excited to make this contribution! Let me know if anything needs tweaking.

##### For all contributions

##### For new estimators
  dependency isolation.

### DIFF
--- a/skpro/distributions/uniform.py
+++ b/skpro/distributions/uniform.py
@@ -86,6 +86,34 @@ class Uniform(BaseDistribution):
         in_bounds = np.logical_and(x >= lower, x <= upper)
         pdf_arr = in_bounds / (upper - lower)
         return pdf_arr
+    
+    def _log_pdf(self, x):
+        """Logarithmic probability density function.
+
+        Parameters
+        ----------
+        x : 2D np.ndarray, same shape as ``self.loc``
+            Values to evaluate the log-pdf at.
+
+        Returns
+        -------
+        2D np.ndarray, same shape as ``self.loc``
+            Log-probabilities of x.
+        """
+        import numpy as np
+
+        # Retrieve parameters (skpro typically broadcast these to match x's shape)
+        lower = self.lower
+        upper = self.upper
+
+        in_bounds = (x >= lower) & (x <= upper)
+
+        # Calculate exact log pdf and handle out-of-bounds without RuntimeWarning
+        return np.where(
+            in_bounds,
+            -np.log(upper - lower),
+            -np.inf
+        )
 
     def _cdf(self, x):
         """Cumulative distribution function.


fkiraly (changes requested):
Is this not exactly the same implementation as from the default `_log_pdf`? Can you explain why this is better than just falling back on the default?

fkiraly (approved):
Ok, I read the PR description carefully. Makes sense.

fkiraly:
This is a copy of https://github.com/sktime/skpro/pull/784

PR #796, opened 2026-03-05
[ENH] MDNRegressor (Mixture Density Network)

joshdunnlime:
#### Reference Issues/PRs

No issue opened.

A new regressor implementation of Mixture Density Network (MDN) as per [Bishop 1994](https://publications.aston.ac.uk/id/eprint/373/1/NCRG_94_004.pdf) with noise regularisation as per [Rothfuss 2019](https://arxiv.org/abs/1907.08982).
It also includes optional passing of pytorch activation functions and optimizers.

It implements a fully vectorised NormalMixture distribution where each rows weights are individually learnt and applied. It also implements a custom vectorised bisection for fast `_ppf` method calls.

Yes, soft deps: [Pytorch](https://docs.pytorch.org/docs/stable/index.html) and [pytorch-optimizer](https://pytorch-optimizers.readthedocs.io/en/latest/).

NormalMixture distribution. This adds slightly opinionated design choice to the API.

Yes. Standard library param tests for dist and est. Additional test for est coming.

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference/regression.rst
+++ b/docs/source/api_reference/regression.rst
@@ -171,8 +171,8 @@ Distributional boosting
 
     XGBoostLSS
 
-Neural conditional density estimation
--------------------------------------
+Deep learning based regression
+------------------------------
 
 .. currentmodule:: skpro.regression.mdn
 
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -240,5 +240,14 @@
         "code",
       ]
     },
+    {
+      "login": "amaydixit11",
+      "name": "Amay Dixit",
+      "profile": "https://github.com/amaydixit11",
+      "contributions": [
+        "code",
+        "bug"
+      ]
+    },
   ]
 }
--- a/docs/source/api_reference/distributions.rst
+++ b/docs/source/api_reference/distributions.rst
@@ -35,6 +35,7 @@ Continuous support - full reals
     :toctree: auto_generated/
     :template: class.rst
 
+    Cauchy
     Laplace
     Logistic
     Normal
@@ -151,6 +152,7 @@ Mixture composition
     :toctree: auto_generated/
     :template: class.rst
 
+    KernelMixture
     Mixture
     NormalMixture
 
--- a/skpro/distributions/__init__.py
+++ b/skpro/distributions/__init__.py
@@ -9,6 +9,7 @@ __all__ = [
     "Binomial",
     "BurrIII",
     "BurrXII",
+    "Cauchy",
     "ChiSquared",
     "Delta",
     "Empirical",
@@ -26,6 +27,7 @@ __all__ = [
     "HalfNormal",
     "Hurdle",
     "IID",
+    "KernelMixture",
     "InverseGamma",
     "InverseGaussian",
     "Histogram",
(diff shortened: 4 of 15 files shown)


pyproject.toml
```diff
@@ -54,9 +54,11 @@ all_extras = [
     "mapie; python_version < '3.13'",
     "matplotlib>=3.3.2",
     "ngboost<0.6.0; python_version < '3.13'",
+    "pytorch-optimizer>=3.2.0; python_version < '3.14'",
```
fkiraly: I think for now we should not add dependencies to the `all_extras` depset - instead, let´s test the estimator on its own VM via the `tests:vm` tag set to `True`.

joshdunnlime:
My understanding of `Mixture` is that it only has global weights for each sub-dist. MDNR _could_ use this, however, it would mean that for each prediction row, it would use a new instance of `Mixture` (computationally this would be crippling). `MixtureNormal` has performant, vectorised implementation of most of the class methods so is actually pretty quick!

You could argue that we could call it `Normal(Row|Instance|HeteroWeighted)Mixture` but I think it is documented clearly enough to avoid confusion.

joshdunnlime:
I did consider the following:
1) Extending `Mixture` to include "instance" and "global" weights.
2) Create a generic `InstanceWeightedMixture` class.
3) A base class `BaseInstanceWeightedMixture`.

The downsides were:
1) Far too much work. IMHO these are two quite different classes to the point I think this would just be a wrapped around a `Mixture` and `InstanceWeightedMixture`.

2) `InstanceWeightedMixture` which accepts a mixture of different dists comes with a lot of complications. The primary one is that the `NormalMixture` has very well defined (and mostly exact) methods for all of the standard methods. My understanding of weighted mixtures of distributions stops a normals so I have no idea what is feasible here. The vectorisation is clearly very well defined for normal distributions in scipy, and assumptions on the support for root-finding in `_ppf` were fairly straightforward. Implemenation-wise, I believe it serves to keep `NormalMixture` even if we have 3) as the speed-ups and exactness are significant.

3) Happy to open a Issue. My knowledge of anything beyond normal mixtures of dists is effectively zero so I doubt I'd be up to a PR.

joshdunnlime:
Just as a side note: The natural extension to MDNs is probably something Flow/Kernel based, as opposed to extending this with non-Normal dists.

docs/source/api_reference/regression.rst
```diff
@@ -171,6 +171,17 @@ Distributional boosting
 
     XGBoostLSS
 
+Neural conditional density estimation
```
fkiraly: how about "deep learning based regressors" instead?

joshdunnlime:
Thanks for updating. I'll do that optimizer package stuff.
I'm also adding some further improvements to the model which seem to help with training speed and stability.

fkiraly:
btw, other topic, did we want to have the discussion on the differentiable transformers sometime? Have not seen you in the meetups, and I really think we need a sync meet on that (at least once). If the meetup times are not convenient, feel free to start a scheduling discussion on the discord.

joshdunnlime:
I've added ngem loss: https://arxiv.org/html/2602.10602v1 to the bellow PR on my fork. It seems to show significant benefits at lower `lr`, particularly for CRPS. I am happy to merge into this `mdn` branch but further tidying is required. See comments in PR:
https://github.com/joshdunnlime/skpro/pull/1

It would be best to hold off and merge this all together to avoid minor breaking changes but I am unclear on how to handle the `ngem_lr` defaults as it really requires lower `lr` to perform well. Comments very much welcome.

Yes, apologies there. I will reach out on Discord regarding the Diff Transform class.

PR #797, opened 2026-03-05
[BUG] fix wrong converter stores in `predict` and `_check_C`

WHOIM1205:
Fix converter store misuse in BaseProbaRegressor

BaseProbaRegressor.predict was using _X_converter_store while converting y_pred
back to the user mtype. This incorrectly writes target column metadata into the
X converter store which corrupts feature name resolution and causes repeated
predict or predict_proba calls to fail with feature name mismatch errors.

The fix switches the store to _y_converter_store so target metadata is cached
in the correct location.

Also fixes a similar issue in _check_C where _y_converter_store was used instead
of _C_converter_store during C conversion.

This prevents converter store contamination and ensures stable repeated
predict and predict_proba calls.


WHOIM1205:
hey @fkiraly 

This PR fixes a small converter store mixup in BaseProbaRegressor.predict where _X_converter_store was used while converting y_pred which ends up writing target column metadata into the X store. After the first predict call this corrupts feature name resolution and can cause subsequent predict or predict_proba calls to fail with feature name mismatch.

The change switches it to _y_converter_store so y metadata is stored correctly.

I also fixed a similar store reference in _check_C where _y_converter_store was used instead of _C_converter_store.

Issue #389, opened 2024-06-14
[ENH] Design of a BayesianEstimator Class for sktime and skpro


meraldoantonio:
**Is your feature request related to a problem? Please describe.**
Not a problem per se, but we're planning to create a few Bayesian estimators in `sktime` and `skpro`.
Bayesian estimators uniquely integrate prior knowledge through prior distributions and update this knowledge based on observed data to form posterior distributions. To accommodate the distinctive workflow of Bayesian estimators, we propose designing a new BayesianEstimator class.

This issue relates to the following two PRs:
- [The design on the `enhancement proposals` repo](https://github.com/sktime/enhancement-proposals/pull/35)
- [The development of `BayesianLinearRegressor` estimator on `skpro`, on which this template will be applied](https://github.com/sktime/skpro/pull/358)

RishavIITK22:
Hi @meraldoantonio @fkiraly !..i have been doing research projects in probabilistic ML for a while now, and I would like to take up the issue. Could you please guide me more on this?

orbaps:
hey can I work on this

orbaps:
Here is a detailed step-by-step plan for the architecture and implementation:

1. Define the BaseBayesianRegressor Architecture
Create a new module (e.g., skpro.regression.base_bayesian.py) containing the BaseBayesianRegressor class. This class should encapsulate the PyMC backend logic so that individual Bayesian estimators (like a Bayesian Ridge or a Bayesian Neural Network) only need to define the probabilistic model structure.

Key responsibilities of the base class:

Manage the PyMC Model context.
Handle MCMC sampling parameters (draws, tune, chains, target_accept).
Store the resulting arviz.InferenceData trace.
2. Standardize Initialization (__init__)
The __init__ method of the base class should accept the MCMC configuration parameters. Individual estimators inheriting from this base class should accept prior specifications in their __init__.

Priors Configuration: As discussed in EP #35, we should support passing priors as a dictionary or by leveraging pymc-marketing's Prior class to avoid re-implementing prior specification logic.
Example: BayesianMyModel(priors={"intercept": pm.Normal.dist(mu=0, sigma=1)})
MCMC Params: The base class should accept standard PyMC pm.sample() kwargs: draws=1000, tune=1000, chains=None, cores=None, random_seed=None.
3. Standardize Fitting (_fit)
The _fit method in the base class should handle the heavy lifting of the MCMC sampling. The inheriting class will define a _build_model(self, X, y) method.

[...]

PR #802, opened 2026-03-06
 [ENH] Add `BaseBayesianRegressor` for Bayesian model interface

arnavk23:
## Reference Issues/PRs
Towards #389

## What does this implement/fix? Explain your changes.

This PR implements the Bayesian model interface design proposed in issue #389. It introduces a `BaseBayesianRegressor` class that encapsulates PyMC backend logic for MCMC sampling and posterior predictive inference, making it easier to create new Bayesian estimators.

The base class handles all PyMC complexity, allowing subclasses to focus only on defining their probabilistic models via the `_build_model()` method.

## Does your contribution introduce a new dependency? If yes, which one?

Yes, adds `arviz` as an optional dependency for Bayesian posterior analysis and diagnostics.

## What should a reviewer concentrate their feedback on?

- API design and consistency with existing skpro patterns
- Proper handling of PyMC model context and MCMC sampling

## Did you add any tests for the change?

The existing tests for `BayesianLinearRegressor` should continue to work with the refactored implementation. Additional tests for the base class could be added in a follow-up PR.

## Any other comments?

This implementation follows the detailed design specification from the issue #389. The base class provides a clean abstraction that will make it easier to implement additional Bayesian estimators in the future.

### For all contributions

### For new estimators
- [x] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured dependency isolation, see the [estimator dependencies guide](https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).


fkiraly (changes requested):
Thanks!

This seems to be primarily MC-based. I think a base Bayesian regressor should be able to accommodate non-MC-paradigms too.

arnavk23:
@fkiraly Thanks for the review. I refactored `BaseBayesianRegressor` by introducing posterior hooks while keeping the existing PyMC/MCMC behavior as the default path.

I also added `BayesianLinearClosedFormRegressor` as a non-MC example to show how conjugate/closed-form methods can now plug into the same base class.

arnavk23:
@fkiraly @marrov @joshdunnlime Needed some help. 
ModuleNotFoundError for pymc and arviz occurs in CI even though these dependencies are present in pyproject.toml. This suggests the CI is not installing all_extras correctly or the environment is not picking up the extras. 
How to fix this, currently the tests are being skipped if pymc or arviz are not installed, preventing ModuleNotFoundError.

joshdunnlime:
> @fkiraly @marrov @joshdunnlime Needed some help. ModuleNotFoundError for pymc and arviz occurs in CI even though these dependencies are present in pyproject.toml. This suggests the CI is not installing all_extras correctly or the environment is not picking up the extras. How to fix this, currently the tests are being skipped if pymc or arviz are not installed, preventing ModuleNotFoundError.

Here is a tags example from [XGBoostLSS implementation](https://github.com/sktime/skpro/blob/main/skpro/regression/xgboostlss.py#L106):

```python
    _tags = {
        # packaging info
        # --------------
        "python_dependencies": ["xgboostlss"],  # PEP 440 python dependencies specifier,
        # CI and test flags
        # -----------------
        "tests:vm": True,  # requires its own test VM to run
        "tests:python_dependencies": ["optuna", "optuna-integration"],
    }
```

In your case, you will want `pymc` as `python_dependencies` (like xgboostlss). My understanding is the python_deps get checked at class init, where as you probably want to use the soft_check function for arviz when plotting/summary is called (the model can still init, fit, predict without arvis). If you want to run tests with arviz calls, use the `test:python_dependencies`.
None of these need to/should be added to the `all_extra` (pymc was already there so I guess leave it).

Here is some more examples in the MDNRegressor: https://github.com/sktime/skpro/pull/796/changes/e6fbe9fa1f08532dca661e679f013c72646f0cfb

I'm still learning so Franz might correct some of this but hopefully this gets you unstuck :)

PR #807, opened 2026-03-06
[BUG] Fix missing raise before TypeError in Pipeline._check_steps

Sanchit2662:
## Description 
---                                                                                                                                                                                                                                                                                                            I found a missing raise keyword in the pipeline validation logic that was silently allowing invalid configurations to pass through.     
                                                                                                                                                                        
In `skpro/regression/compose/_pipeline.py, line 111` was creating a TypeError but never raising it. This meant if someone built a pipeline with a transformer as the last step (when allow_postproc=False), the validation would just skip over the error instead of               failing.

## Fix
  Added raise before the TypeError on line 111. Now invalid pipelines properly raise an exception instead of silently succeeding and breaking later during fit/predict.

  It's a one-word fix that aligns with the two other validation checks in the same method which already use raise correctly. Pretty straightforward catch.


Sanchit2662:
Hi @fkiraly @felipeangelimvieira , please look review the changes.

PR #815, opened 2026-03-07
fix: use correct converter store for y conversion in predict()

ashnaaseth2325-oss:
# Summary

This PR fixes a bug in `predict()` where the wrong converter store was used.

In `BaseProbaRegressor.predict()`, the output conversion of `y_pred` was writing to `_X_converter_store`. This is incorrect because the conversion is related to `y`, not `X`.

Because of this, the first `predict()` call could store the target column name inside `_X_converter_store`. On the next `predict()` call, this store could be used when reconstructing `X`, which may change the column name and cause a sklearn feature name mismatch error.

# Steps to Reproduce

Example:

```python
X_train = np.array([[1.0],[2.0],[3.0]])
y_train = pd.Series([1.1,2.0,3.1], name="target")

reg = ResidualDouble(LinearRegression())
reg.fit(X_train, y_train)

reg.predict(X_test)  # works
reg.predict(X_test)  # may crash
```

The second call can raise a feature name mismatch error.

# Fix

Changed the store used during output conversion.

Before:

```python
store=self._X_converter_store
```

After:

```python
store=self._y_converter_store
```

A similar fix was applied in `_check_C()` so the correct converter store is used there as well.

# Result

After the fix:

* `predict()` works correctly when called multiple times
* converter stores are kept separate
* feature name errors no longer happen

# Test

Added a regression test that calls `predict()` twice and checks that both results are the same.


ashnaaseth2325-oss:
Hello @fkiraly @felipeangelimvieira 
It looks like the failing CI checks are all from the `SurvivalForestSkSurv` and `SurvivalForestXtraSkSurv` estimator tests.
From the logs, the failure is caused by a `NotImplementedError` raised by `feature_importances_` inside `sksurv` when `_get_fitted_params_default_safe()` iterates over `dir(obj)`. This happens during `test_fit_does_not_overwrite_hyper_params`.
Since this occurs during estimator inspection in the survival adapter and the change in this PR only affects the converter store used in `predict()`, it seems unrelated to the fix here. Happy to investigate further if needed.

ashnaaseth2325-oss:
Hello @fkiraly @felipeangelimvieira 
Hi, just checking if you had a chance to look at the CI failures. Let me know if I should investigate or try a fix. Thanks!

PR #818, opened 2026-03-08
[MNT] Fix permission error in update_contributors workflow (#816)

### LINKED ISSUE #816: [BUG] `generate-markdown-and-create-pr` CI job is broken
The `generate-markdown-and-create-pr` CI job is broken - it seems to fail with a generic error.

This should be investigated and fixed.

kabirvashisht4-glitch:
**Reference Issues/PRs**
  Fixes #816.

**What does this implement/fix? Explain your changes.**
I noticed the update_contributors workflow was consistently failing with a generic error during the PR creation step. After looking into it, I found two main issues:

Permissions: The default GITHUB_TOKEN didn't have write access to push the new branch or open the PR.

Tooling: The existing script used gh pr create with a manual login, which felt a bit fragile in a CI environment.

**My changes**:

Added a permissions block at the top of the YAML to grant contents: write and pull-requests: write.

Replaced the manual gh CLI commands with the peter-evans/create-pull-request action. I think this is a more robust way to handle the "commit-push-PR" sequence because it manages branch naming and PR metadata more cleanly.

**Does your contribution introduce a new dependency? If yes, which one?
It adds a dependency on the peter-evans/create-pull-request@v5 GitHub Action.

Please check if the permissions scope is too broad or if you prefer sticking to the gh CLI. From my testing, the action approach seems to be the standard way to fix these "generic" permission failures.

**Did you add any tests for the change?
I haven't added new tests as this is a CI/CD workflow change. However, I’ve verified the YAML syntax and the logic for the if: env.BRANCH != '' condition to ensure we don't trigger empty PRs.


kabirvashisht4-glitch:
@fkiraly could you please review this PR

Issue #376, opened 2024-06-07
[ENH] feature importance interface and tag for regression models


fkiraly:
Some native and interfaced models have the ability to return feature importances, such as various tree models, or GAM and cyclic-boosting.

We should consider how to best design this and implement, possibly:

* a feature importance interface point
* a `capability:feature_importance` tag that indicates whether the model return this

amanpandey2587:
Hi @fkiraly , I have started working on this issue . My approach would be like this 
1. Register the tag (skpro/registry/_tags.py)
Add capability:feature_importance (type bool, default False) to OBJECT_TAG_REGISTER, consistent with existing capability tags like capability:update and capability:missing.
2. Add the public interface (skpro/regression/base/_base.py)
Add a feature_importances() method to BaseProbaRegressor that:
Raises NotImplementedError if the tag is False
Prefers a _feature_importances() private hook over the feature_importances_ attribute (following skpro conventions)
Validates length against feature_names_in_ and returns a pd.Series indexed by feature names
3. Concrete estimators
skpro/survival/ensemble/_grad_boost_sksurv.py : set tag True for SurvGradBoostSkSurv and SurvGradBoostCompSkSurv, both of which natively expose feature_importances_ via sksurv. RandomSurvivalForest will be left out as sksurv does not expose feature_importances_ for it.
skpro/regression/gam/_gam.py : set tag True, derive importances from per-term pseudo-R² (deviance explained per smooth term, normalised to sum to 1)
4. Tests
Add tests covering tag gating, not-fitted guard, return type/shape, index alignment, both implementation paths, error conditions, and per-estimator integration smoke tests.
Please guide me on this whether my approach is correct or requires some refinement.
Thanks , I have raised a pull request for the same .

PR #842, opened 2026-03-09
ENH adding feature importances support to skpro probabilistic regressors

amanpandey2587:
#### Reference Issues/PRs
Fixes Issue : #376 

I have added a feature_importances() public method to BaseProbaRegressor in _base.py, along with a new capability:feature_importance tag (default False) in _tags. The method returns a pd.Series of feature importances indexed by feature names seen in fit.
- If the tag is False, calling feature_importances() raises NotImplementedError
- If the tag is True, the method prefers the _feature_importances() private hook over the feature_importances_ attribute,    following existing skpro conventions
- Length of importances is validated against feature_names_in_ and raises ValueError on mismatch

I have also set capability:feature_importance=True in GradientBoostingSurvivalRegressor (_grad_boost_sksurv.py) and implemented the _feature_importances() hook, delegating to the underlying sksurv estimator's feature_importances_ attribute.

No new dependencies were introduced 

- The design of the fallback chain in feature_importances() , here the hook is preferred over attribute so that it is consistent with skpro conventions.
- to check whether the tag name capablility:feature_importance fites the existing tag naming pattern or not
- to check the placement of feature_importances() in _base.py relative to predict()
- test coverage in test_feature_importances,py

Yes ,I have added skpro/regression/tests/test_feature_importances.py 
- tag gating :  NotImplementedError when tag is False
- not-fitted guard : RuntimeError before fit
- return type, shape, name, and index alignment with feature_names_in_
- both implementation paths (attribute and _feature_importances() hook)
- error conditions : valueError on length mismatch, AttributeError when tag is True but nothing is implemented
- GradientBoostingSurvivalRegressor integration smoke tests
I am also adding a screenshot of passing of these tests as well 
image

##### For all contributions

### DIFF
--- a/skpro/regression/ensemble/_ngboost.py
+++ b/skpro/regression/ensemble/_ngboost.py
@@ -92,7 +92,7 @@ class NGBoostRegressor(BaseProbaRegressor, NGBoostAdapter):
         "authors": ["ShreeshaM07"],
         "maintainers": ["ShreeshaM07"],
         "python_dependencies": "ngboost",
-        "capability:feature_importance": True,
+        "capability:feature_importance": False,
     }
 
     def _feature_importances(self):
@@ -228,6 +228,9 @@ class NGBoostRegressor(BaseProbaRegressor, NGBoostAdapter):
 
         self.ngb_ = clone(self.ngb)
         self.ngb_.fit(X, y)
+        # set tag dynamically based on whether base learner exposes feature importances
+        supports_importance = self.ngb_.feature_importances_ is not None
+        self.set_tags(**{"capability:feature_importance": supports_importance})
         return self
 
     def _predict(self, X):


amanpandey2587:
Hi @fkiraly , I have added feature_importances along with tag to NGBoost, GAM, XGBOostLSS , sksurv gradient boosting and CyclicBoosting , also the BaseProbaRefressor.feature_importances now enforces a pd.Series named feature_importance which is indexed be feature_names_in with length checks and docstring explains index ,name and value expectations. and one more thing now the feature_importances supply the standardized Series(hook-first design) so that base now prefers the hook, I have added stub for detection and each model hook returns the standardized pd.seriese which fallback  to feature_importances only if hook is absent .
I have also implemented test for each of them which are passing .
Please check these changes and tell me if anything else is required.
Thanks

skpro/regression/gam/_gam.py
```diff
@@ -209,6 +210,27 @@ def _predict(self, X):
...
+            Index: feature names from ``fit``. Name: ``"feature_importance"``.
+            Values: range (max - min) of partial dependence per term.
+        """
+        gam = self.estimator_
+        names = self.feature_names_in_
+        n = len(names)
+        imp = np.zeros(n, dtype=float)
+        for i in range(n):
+            try:
+                pd_val = gam.partial_dependence(term=i)
```
fkiraly: is this correct if terms are not `"auto"`?

FYI @dswah, your comment would be appreciated.

skpro/regression/ensemble/_ngboost.py
```diff
@@ -92,8 +92,42 @@ class NGBoostRegressor(BaseProbaRegressor, NGBoostAdapter):
...
+        pd.Series
+            Index: feature names from ``fit``. Name: ``"feature_importance"``.
+            Values: from the underlying NGBoost model when base learner supports it.
+        """
+        import numpy as np
+        import pandas as pd
+
+        names = self.feature_names_in_
+        n_features = len(names)
+        if hasattr(self.ngb_, "feature_importances_"):
```
fkiraly: could you explain when this exists, and point to the exact documentation in ngboost for that?

skpro/regression/base/_base.py
```diff
@@ -204,6 +205,80 @@ def _update(self, X, y, C=None):
...
+        above. Library-specific handling (e.g. mapping from scikit-survival or
+        xgboost) belongs in that override, so the public method always returns
+        the same structure.
+        """
+        self.check_is_fitted()
+
+        if not self.get_tag("capability:feature_importance", False):
+            raise NotImplementedError(
+                f"{type(self).__name__} does not provide feature importances. "
+                "Set the tag 'capability:feature_importance' to True and implement "
```
fkiraly: I think that is the wrong recommendation for the user. The user should be told that this simply does not exist.
Then call an inner private method.

skpro/regression/base/_base.py
```diff
@@ -204,6 +205,80 @@ def _update(self, X, y, C=None):
...
+                "`_feature_importances` to return a pd.Series in the documented format."
+            )
+
+        hook = getattr(type(self), "_feature_importances", None)
```
fkiraly: not sure what you are doing and why. I think it should just call a private method, after input checks and conversions. Like every other method.

skpro/regression/gam/_gam.py
amanpandey2587: I actually missed that case. The original implementation was using partial_dependence iterating over term indices assuming a 1 to 1 mapping with features.I didn't accounted for cases like interaction terms or explictly passed term structures where teh index i no longer corresponds to feature i.The fix now inspects gam.terms directly, maps each term to its feature indicies via the feature/features attributes and distributes the partial dependence range across all involved features for interaction terms.If term metadata is unavailable it falls back to one term per features. This should handles correctly both cases.

skpro/regression/base/_base.py
amanpandey2587: I have implemented it as per your saying now.

skpro/regression/base/_base.py
amanpandey2587: Same as above, it is now implemented as per your saying

skpro/regression/ensemble/_ngboost.py
amanpandey2587: I had previously added hasattr check as a defensive guard because I had assumed feature_importances_ might not exist on all NGBoost instances ,particularly when using non tree base learners like LinearRegression. My resasoning was based on the sklearn convention where feature_importances_ is only set as an instance attribute when the model supports it. I had also referenced the Base Learners section of the NGBoost usage docs (https://stanfordmlgroup.github.io/ngboost/1-useage.html#base-learners) which documents that any sklearn regressor can be passed as Base but does not describe the internal behaviour of feature_importances_. 
After checking the NGBoost source code at https://github.com/stanfordmlgroup/ngboost/blob/master/ngboost/ngboost.py ,  I found out that feature_importances_ is actually a @property defined directly on the NGBoost base class so it is always present on every fitted instance regardless of the base learner. The property return None when the base leaner is not a tree based regressor. So the hasattr guard was based on a wrong assumption, I should have checked the source code along with docs and sklearn conventions. I had also made the capability:feature_importance to be updated dynamically in _fit after fitting which makes the tag truthful at runtime ( True) only when the fitted base learner actually exposes feature importances

PR #858, opened 2026-03-10
[ENH] add ConstantVarianceRegressor baseline wrapper

NandiniDhanrale:
Related to #7

## Summary
- add `ConstantVarianceRegressor`, a probabilistic regression baseline that uses a deterministic regressor for the mean and the training sample variance as a constant predictive variance
- support both `Normal` and `Laplace` predictive distributions with variance calibrated to the training sample
- document the new estimator in the regression API reference and changelog
- add targeted estimator tests

## Details
- adds `skpro.regression.constantvariance.ConstantVarianceRegressor`
- wraps a deterministic sklearn-compatible regressor and uses its point predictions as the predictive mean
- uses the training target variance as the predictive variance for every future instance
- implements explicit `predict_proba` for `Normal` and `Laplace` so the wrapper is not restricted to the framework default Gaussian fallback

## Tested
- `py -m pytest -o addopts="" skpro/regression/tests/test_constantvariance.py -q`
- `py -m pytest -o addopts="" skpro/regression/tests/test_all_regressors.py -k ConstantVarianceRegressor -q`

### DIFF
--- a/skpro/regression/constantvariance.py
+++ b/skpro/regression/constantvariance.py
@@ -12,6 +12,8 @@ from skpro.distributions.normal import Normal
 from skpro.regression.base import BaseProbaRegressor
 from skpro.utils.sklearn import prep_skl_df
 
+_MIN_VARIANCE = 1e-12
+
 
 class ConstantVarianceRegressor(BaseProbaRegressor):
     """Probabilistic regressor with mean from a deterministic regressor.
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Thanks!

* please do not change the changelog, this is a past changelog
* please add the new regressor after the dummy regressor

NandiniDhanrale:
@fkiraly 
Addressed the requested doc changes and fixed the constant-target edge case.

Updates:
- kept the past changelog unchanged
- moved `ConstantVarianceRegressor` to appear after `DummyProbaRegressor` in the regression API reference
- stabilized `ConstantVarianceRegressor` for zero-variance training targets by applying a small positive variance floor so `Normal`/`Laplace` predictions remain numerically valid
- added a regression test covering constant targets for both supported distributions

Tested:
- `py -m pytest -o addopts="" skpro/regression/tests/test_constantvariance.py -q`
- `py -m pytest -o addopts="" skpro/regression/tests/test_all_regressors.py -k ConstantVarianceRegressor -q`

PR #859, opened 2026-03-10
[ENH] Add plot_calibration function to utils/plotting #162

### LINKED ISSUE #162: [ENH] Calibration plots for probabilistic predictions
@benHeid made nice calibration plots here: https://github.com/sktime/sktime/pull/5632

So I wonder:

* should these also be added in `skpro` for probabilistic tabular regressors? Same principle works.
* any reasons why it could not be shifted via copy/paste and import change?

Mid-term, it might be best to have them only in `skpro`, and make `sktime` forecasting rely more on `skpro` distributions machinery and utilities.

FYI @benHeid, any thoughts?

MBKKHAN:
#### Reference Issues/PRs
Fixes #162

Adds `plot_calibration` function to `skpro/utils/plotting.py`.

The function plots a calibration curve for probabilistic predictions by comparing predicted quantile levels to observed coverage fractions. A well-calibrated model should have its curve close to the diagonal.

The implementation follows the same pattern as the existing plotting functions in the same file.

No. Uses matplotlib which is already used in the same file.

- Correctness of the calibration curve calculation
- Whether the function signature matches skpro conventions

Not yet. Happy to add tests if requested.

##### For all contributions
- [x] The PR title starts with [ENH].

### DIFF
--- a/skpro/tests/test_plot_calibration.py
+++ b/skpro/tests/test_plot_calibration.py
@@ -15,7 +15,8 @@ def test_plot_calibration_runs():
             (0, 0.9): [3, 4, 4, 5],
         }
     )
-    ax = plot_calibration(y_true, y_pred)
+    fig, ax = plot_calibration(y_true, y_pred)
+    assert fig is not None
     assert ax is not None
 
 
--- a/skpro/utils/plotting.py
+++ b/skpro/utils/plotting.py
@@ -245,25 +245,57 @@ def plot_crossplot_loss(y_true, y_pred, metric, ax=None):
 
 
 def plot_calibration(y_true, y_pred, ax=None):
-    """Plot calibration curve for probabilistic predictions.
+    r"""Plot the calibration curve for a sample of quantile predictions.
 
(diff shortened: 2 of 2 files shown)


MBKKHAN:
Thank you for the feedback @fkiraly! I will look at the sktime plot_calibration implementation more carefully and update the function to also handle BaseDistribution instances. Working on it now.

MBKKHAN:
@fkiraly I have now carefully reviewed the sktime plot_calibration implementation and am updating the function to align more closely with it  including the figure/ax return pattern and DataFrame-based plotting. I have also added BaseDistribution support. I Will push the update shortly. Thanks for your support and understanding.

MBKKHAN:
@fkiraly I have completed the update. The function now aligns with the sktime implementation  wide figure size, DataFrame-based plotting, and fig/ax return pattern. BaseDistribution support has also been added so the function accepts both predict_quantiles and predict_proba output directly. Please let me know if further changes are needed.

PR #863, opened 2026-03-11
[BUG] Fix TypeError in BaseProbaMetric._evaluate_by_index

### LINKED ISSUE #809: [BUG] TypeError in BaseProbaMetric._evaluate_by_index — too many positional arguments passed to evaluate()
**Describe the bug**
In `skpro/metrics/base.py`, the default jackknife fallback in `_evaluate_by_index` (lines 257 and 259) calls `self.evaluate(y_true, y_pred, self.multioutput, **kwargs)`. The `evaluate()` method signature only accepts two positional arguments (`y_true`, `y_pred`), so passing `self.multioutput` as a third positional argument raises a `TypeError`.

**To Reproduce**
Any custom `BaseProbaMetric` subclass that implements only `_evaluate` (not `_evaluate_by_index`) will trigger the default jackknife path in `_evaluate_by_index`, which crashes with: `TypeError: evaluate() takes 3 positional arguments but 4 were given`.

**Expected behavior**
The jackknife fallback should call `self.evaluate(y_true, y_pred, **kwargs)` without the extra `self.multioutput` argument, since `evaluate()` already accesses `self.multioutput` internally.

**Additional context**
The fix is to remove `self.multioutput` from lines 257 and 262 in `skpro/metrics/base.py`. I'd like to work on this — please assign this issue to me.

Mahaveerjain-18:
#### Reference Issues/PRs
Fixes #809

This PR fixes a bug in `BaseProbaMetric._evaluate_by_index` where the method `evaluate()` was called with an extra positional argument `self.multioutput`.

The `evaluate()` method signature only accepts `y_true` and `y_pred` as positional arguments, with additional parameters passed through `**kwargs`. Passing `self.multioutput` as a positional argument caused the following error:

TypeError: evaluate() takes 3 positional arguments but 4 were given

To resolve this issue, the unnecessary positional argument `self.multioutput` was removed from the `evaluate()` calls in `_evaluate_by_index`.

This aligns the method calls with the correct `evaluate()` function signature and prevents the TypeError.

No. This change does not introduce any new dependencies.

- Correctness of the fix in `_evaluate_by_index`
- Ensuring the behavior of the jackknife fallback remains unchanged
- Consistency with existing code style

No new tests were added. The change corrects a method call without altering the expected behavior of the metric. All existing tests pass locally.

#### Any other comments?
Local test results after applying the fix:

653 passed, 317 skipped

##### For all contributions
- [ ] I've added myself to the list of contributors
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]

##### For new estimators
- [ ] Not applicable

### DIFF
--- a/skpro/metrics/base.py
+++ b/skpro/metrics/base.py
@@ -251,14 +251,21 @@ class BaseProbaMetric(BaseObject):
             must have same index and columns as ``y_true``.
             Predicted values, i-th row is prediction for i-th row of ``y_true``.
         """
+        multioutput = self.multioutput
         n = y_true.shape[0]
-        out_series = pd.Series(index=y_pred.index)
+        if isinstance(multioutput, str) and multioutput == "raw_values":
+            out_series = pd.DataFrame(
+                index=y_true.index, columns=y_true.columns, dtype="float64"
+            )
+        else:
+            out_series = pd.Series(index=y_true.index, dtype="float64")
         try:
             x_bar = self.evaluate(y_true, y_pred, **kwargs)
             for i in range(n):
-                out_series[i] = n * x_bar - (n - 1) * self.evaluate(
-                    np.vstack((y_true[:i, :], y_true[i + 1 :, :])),  # noqa
-                    np.vstack((y_pred[:i, :], y_pred[i + 1 :, :])),  # noqa
+                idx = y_true.index[i]
+                out_series.loc[idx] = n * x_bar - (n - 1) * self.evaluate(
+                    y_true.drop(idx),
+                    y_pred.drop(idx),
                     **kwargs,
                 )
             return out_series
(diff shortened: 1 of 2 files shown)


Mahaveerjain-18:
Hi! I’ve implemented the fix for issue #809 by removing the extra positional argument `self.multioutput` from the `evaluate()` calls in `_evaluate_by_index`.
All tests pass locally. Please let me know if any changes are needed.

Mahaveerjain-18:
Thanks for the suggestion!

I updated `_evaluate_by_index` to remove the extra positional argument passed to `evaluate()` and replaced the previous `np.vstack` logic with pandas `drop()` so the DataFrame structure is preserved.

I also added a regression test in `test_probabilistic_metrics.py` to ensure that `evaluate_by_index` runs without raising a TypeError for all probabilistic metrics.

I compared the implementation with `sktime`’s `BaseForecastingErrorMetric._evaluate_by_index`. In `sktime`, the implementation operates directly on forecasting data structures, whereas in `skpro` the predictions are probabilistic DataFrames with additional column levels. Using `drop()` instead of numpy stacking preserves the DataFrame structure expected by `evaluate()` and avoids breaking the validation logic.

All tests pass locally (661 passed, 317 skipped).

Mahaveerjain-18 (commented):
changes made

Mahaveerjain-18:
@fkiraly @felipeangelimvieira Hi! Just checking if any additional changes or tests are needed for this PR. Happy to update it if required. Thanks!

PR #872, opened 2026-03-11
[ENH] Update `TestAllRegressors` to check against parameter mutation in non-state changing methods

Ahmed-Zahran02:
#### Reference Issues/PRs

Fixes https://github.com/sktime/skpro/issues/867

The  problem with predict_proba is that it can mutate the parameters dict, and the original test cases were using the function once so the change couldn't be triggered. So we need to check them after using that function to avoid value errors or failures. See https://github.com/sktime/skpro/pull/861
My suggested solution adds a scenario where we use the `predict_proba` function twice:
```python
        params_before = copy.deepcopy(regressor.get_params())

        regressor.predict_proba(X_test)
        regressor.predict_proba(X_test[:5])

        params_after = regressor.get_params()
```
, then we check they are the same.

### DIFF
--- a/skpro/regression/residual.py
+++ b/skpro/regression/residual.py
@@ -351,7 +351,7 @@ class ResidualDouble(BaseProbaRegressor):
 
         # collate all parameters for the distribution constructor
         # distribution params, if passed
-        params = distr_params
+        params = dict(distr_params)
         # row/column index
         ix = {"index": X.index, "columns": self._y_cols}
         params.update(ix)


Ahmed-Zahran02:
Ok, I'm working on it.

Ahmed-Zahran02:
I noticed that some of my fails are fixed with the change in #861.
Should I apply it in my branch?

fkiraly (approved):
Thanks!

PR #876, opened 2026-03-11
fix: get_test_params uses `self` instead of `cls` in several metric classes

### LINKED ISSUE #875: [BUG]: get_test_params uses `self` instead of `cls` in several metric classes
In `skpro/metrics/_classes.py`, several metric classes define 
`get_test_params` as a `@classmethod` but incorrectly use `self` 
as the first argument instead of `cls`:

- `EmpiricalCoverage.get_test_params(self)`
- `IntervalWidth.get_test_params(self)`
- `ConstraintViolation.get_test_params(self)`
- `LinearizedLogLoss.get_test_params(self)`

`@classmethod` methods receive the class as first argument, 
which by convention must be named `cls`. Using `self` implies 
an instance method and is incorrect. Compare with `PinballLoss`, 
`LogLoss`, `CRPS`, and `AUCalibration` in the same file which 
all correctly use `cls`.

**Expected behavior**
All `get_test_params` classmethods should use `cls` as the first argument.

**Proposed Fix**
Change all four occurrences from:
```python
@classmethod
def get_test_params(self):
```
to:
```python
@classmethod
def get_test_params(cls, parameter_set="default"):
```

**File**
`skpro/metrics/_classes.py`

OmAmbole009:
Fixes #875

## What
Replaced `self` with `cls` and added `parameter_set="default"` argument 
in `get_test_params` classmethods for four metric classes in 
`skpro/metrics/_classes.py`.

## Why
`@classmethod` methods receive the class as first argument, which by 
convention must be named `cls`, not `self`. Using `self` implies an 
instance method and is incorrect. This was inconsistent with 
`PinballLoss`, `LogLoss`, `CRPS`, `SquaredDistrLoss` and `AUCalibration` 
in the same file, which all correctly use `cls`.

## Classes fixed
- `EmpiricalCoverage`
- `IntervalWidth`
- `ConstraintViolation`
- `LinearizedLogLoss`

## Change
- `skpro/metrics/_classes.py`: 4 occurrences of `get_test_params(self)` 
  replaced with `get_test_params(cls, parameter_set="default")`

### DIFF
--- a/skpro/metrics/_classes.py
+++ b/skpro/metrics/_classes.py
@@ -267,7 +267,7 @@ class EmpiricalCoverage(BaseProbaMetric):
         return out_df
 
     @classmethod
-    def get_test_params(self):
+    def get_test_params(cls, parameter_set="default"):
         """Retrieve test parameters."""
         params1 = {}
         params2 = {"coverage": 0.5}
@@ -367,7 +367,7 @@ class IntervalWidth(BaseProbaMetric):
         return out_df
 
     @classmethod
-    def get_test_params(self):
+    def get_test_params(cls, parameter_set="default"):
         """Retrieve test parameters."""
         params1 = {}
         params2 = {"coverage": 0.5}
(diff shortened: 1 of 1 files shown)


OmAmbole009:
Hi @fkiraly @felipeangelimvieira Review this Pr.

fkiraly (approved):
That's right, thanks!

How did this not get detected by tests?

PR #878, opened 2026-03-12
[MNT] Replace raw assertions with explicit ValueError in coercion utilities

Ayushigitgithub:
#### Reference Issues/PRs
None

#### What does this implement/fix?
Replace raw `assert` statements in `skpro/metrics/_coerce.py` with
explicit `ValueError` exceptions so invalid inputs give clearer
error messages instead of an empty `AssertionError`.

No


Ayushigitgithub:
Hi @fkiraly 
Thanks for the review
I updated the targeted coercion helpers to replace the raw asserts with helper specific ValueError messages, restored the original file structure, and ran pre-commit on the file.

PR #880, opened 2026-03-12
[BUG] Fix `Levy` distribution to conform with `scipy` interface

### LINKED ISSUE #879: [BUG] Levy distribution: __dict__ manipulation breaks sklearn.clone() after set_params
**Describe the bug**

The `Levy` distribution (`skpro/distributions/levy.py`) manually writes to 
`self.__dict__` inside `__init__` *after* the property setters have already run:
```python
self.__dict__["scale"] = scale
self.__dict__["loc"] = loc
```

This creates a split-state bug: `self.loc` (via property → `_loc`) and 
`self.__dict__["loc"]` (raw dict entry) diverge after any `set_params()` call. 
Since `sklearn.clone()` reads parameters via `get_params()` and reconstructs 
the object, the cloned instance silently gets the **original constructor values** 
instead of the updated ones — no error is raised, making this a silent data corruption bug.

**To Reproduce**
```python
from skpro.distributions.levy import Levy
from sklearn.base import clone

dist = Levy(loc=0.0, scale=1.0)
dist.set_params(loc=5.0)

print(dist.loc)               # → 5.0  (property reads _loc correctly)
print(dist.__dict__["loc"])   # → 0.0  ← STALE, never updated by set_params

cloned = clone(dist)
print(cloned.loc)             # → 0.0  ← Wrong! Expected 5.0
```

**Expected behavior**

After `set_params(loc=5.0)`, both `dist.loc` and any cloned/pickled copy should 
return `5.0`. The `__dict__` entries should never shadow the property values.

The fix is to remove the `__dict__` manipulation and the unnecessary 
property/setter boilerplate entirely, and refactor `Levy` to inherit from 
`_ScipyAdapter` — exactly like the structurally identical `FatigueLife` 
distribution already does. This removes ~40 lines of redundant overrides 
(`_pdf`, `_cdf`, `_ppf`, `_mean`, `_var`) that the adapter handles automatically.

**Environment**

- OS: Windows 11
- Python: 3.11
- skpro: latest `main` branch
- scipy: 1.11+
- sklearn: 1.3+

**Additional context**

Every other scipy-backed distribution in the codebase (`FatigueLife`, `Burr12`, 
`GenPareto`, etc.) correctly uses `_ScipyAdapter` with no `__dict__` hacks. 
`Levy` is the only one that deviates from this pattern, introduced in the 
original PR that added the distribution. The proposed refactor brings it in 
line with the rest of the codebase and eliminates the bug entirely.

direkkakkar319-ops:
**Description:**
Fixes #879 

**What was broken:**
`Levy.__init__` manually wrote to `self.__dict__["loc"]` and `self.__dict__["scale"]`
after the property setters already ran. This caused split-state — after `set_params()`,
the property returned the updated value but `__dict__` retained the stale constructor
value. `sklearn.clone()` silently returned the wrong parameters.

**What this PR does:**
- Removes the `__dict__` manipulation entirely
- Removes the unnecessary `@property` / setter boilerplate for `loc` and `scale`  
- Renames `loc` → `mu` to avoid conflict with `BaseDistribution`'s reserved `loc` property
- Follows the same pattern as `Normal` and `Laplace` in the codebase

**Tests:**
- 65 Levy-specific tests: all passed 
- 89 full distribution tests: all passed

### DIFF
--- a/skpro/distributions/levy.py
+++ b/skpro/distributions/levy.py
@@ -2,10 +2,10 @@
 
 from scipy.stats import levy
 
-from skpro.distributions.base import BaseDistribution
+from skpro.distributions.adapters.scipy import _ScipyAdapter
 
 
-class Levy(BaseDistribution):
+class Levy(_ScipyAdapter):
     r"""Levy probability distribution.
 
     The Levy distribution is parametrized by location :math:`\mu` and


fkiraly:
There were a lot of spurious methods. I removed them, the distribution is hopefully fixed now.

direkkakkar319-ops:
Hi @fkiraly! Thank you for jumping in and cleaning that up .
Glad the bug was worth fixing. I've learned a lot about how BaseDistribution handles parameters from this.

direkkakkar319-ops:
Is there anything else you'd like me to address before this PR is ready . Happy to update tests, docs, or anything else needed!

direkkakkar319-ops:
Hi @fkiraly! I noticed after your cleanup that the base class was still BaseDistribution — changed it to _ScipyAdapter so that _get_scipy_object and _get_scipy_param work correctly.

All 65 Levy tests passing now 

image

direkkakkar319-ops:
Changed this 

`Examples
    --------
    >>> from skpro.distributions.levy import Levy
    >>> dist = Levy(mu=0.0, scale=1.0)
    >>> dist.mean()`

to this

`Examples
    --------
    >>> from skpro.distributions.levy import Levy
    >>> dist = Levy(mu=0.0, scale=1.0)
    >>> dist.mean()
    inf`

in this file skpro/distributions/levy.py

direkkakkar319-ops:
Hi @fkiraly soon will fix to pass the test case. Inform you when is done .

fkiraly (approved):
Finished it.

PR #886, opened 2026-03-13
Fix RuntimeWarning on out-of-bounds evaluation in QPD_S._cdf()

### LINKED ISSUE #885: [BUG] RuntimeWarning: invalid value encountered in log when evaluating QPD_S.cdf at or below bounds
### Description
Calling **cdf()** on **QPD_S** (the semi-bounded Johnson distribution) with values at or below the lower bound raises a `RuntimeWarning: invalid value encountered in log`. This happens because `np.log` is called before any boundary masking is applied.

### To Reproduce

```
import numpy as np
import warnings
from skpro.distributions.qpd import QPD_S
qpd = QPD_S(alpha=0.2, qv_low=[-0.3], qv_median=[0.0], qv_high=[0.3], lower=-0.5)
x = np.array([-1.0, -0.5])  # out of bounds and exactly at bound
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    res = qpd.cdf(x)
    for warn in w:
        print(f"Warning caught: {warn.message}")
```

Expected behavior **QPD_S.cdf()** should safely and silently return **0.0** for any **x <= lower** without throwing internal math warnings.

Environment is the latest main branch of skpro.

Additional context The fix is pretty straightforward (just applying a boolean mask before np.log like _pdf does).

kindler-king:
**Description:**
This fixes a bug in the **QPD_S** (Johnson Quantile-Parameterized Distribution, semi-bounded mode) distribution where `np.log()` was being blindly called across the entire input array, even for values outside the valid support bound ($X \le \text{lower}$).

Evaluating the CDF at or below the theoretical support bound previously triggered` RuntimeWarning: invalid value encountered in log` because negative values and 0 were being passed straight into `np.log`.

The fix conditionally masks the np.log and np.arcsinh operations so they only apply to the valid support. Out-of-bounds indices now safely and immediately return 0.0 without complaining.

**Changes made:**

- Updated `QPD_S._cdf(x)` in **qpd.py** to use a valid_mask (x > lower), bypassing negative/zero calculations entirely.
- Added  `test_qpd_s_bounds()` in **test_qpd.py** to explicitly test bounds and ensure no warnings are thrown.

**Testing**
Ran the skpro/distributions test suites. test_qpd_s_bounds ensures no mathematically unstable operations are called on the boundaries.

Fixes #885

### DIFF
--- a/skpro/distributions/qpd.py
+++ b/skpro/distributions/qpd.py
@@ -324,9 +324,21 @@ class QPD_S(BaseDistribution):
 
         phi = self.phi
 
-        in_arcsinh = np.log((x - lower) / theta) / kappa
+        # Only compute cdf for x > lower
+        valid_mask = x > lower
+        
+        cdf_arr = np.zeros_like(x, dtype=float)
+
+        if not np.any(valid_mask):
+            return cdf_arr
+
+        x_valid = np.where(valid_mask, (x - lower) / theta, 1.0)
+        
+        in_arcsinh = np.log(x_valid) / kappa
         in_sinh = np.arcsinh(in_arcsinh) - np.arcsinh(n * c * delta)
-        cdf_arr = phi.cdf(np.sinh(in_sinh) / delta)
+        cdf_valid = phi.cdf(np.sinh(in_sinh) / delta)
+        
+        cdf_arr = np.where(valid_mask, cdf_valid, 0.0)
 
         return cdf_arr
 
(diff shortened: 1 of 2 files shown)


arnavk23:
@kindler-king  Loved the work done here by you. One suggestion try passing flake+black on your branch or just try precommit files instead on it. It will help you as when we pass workflows generally what is causing problems for distributions is code quality checks.

PR #893, opened 2026-03-13
[BUG] Fix GeneralizedPareto __dict__ hack causing sklearn.clone() to return wrong parametersc

### LINKED ISSUE #889: [BUG] GeneralizedPareto distribution has __dict__ hack causing sklearn.clone() to return wrong parameters
**Describe the bug**

`GeneralizedPareto.__init__` in `skpro/distributions/gen_pareto.py` has a 
`__dict__` hack that causes `sklearn.clone()` to silently return wrong parameters. 
The same bug was previously found and fixed in the Levy distribution (#879).

**To Reproduce**

from skpro.distributions.gen_pareto import GeneralizedPareto
from sklearn.base import clone

dist = GeneralizedPareto(c=0.5, scale=1.0, loc=0.0)
dist.set_params(scale=2.0)

cloned = clone(dist)
print(cloned.scale)  # Returns 1.0 instead of 2.0 — wrong!

**Expected behavior**

`sklearn.clone()` should return a distribution with the updated parameters. 
`cloned.scale` should return `2.0` after `set_params(scale=2.0)`.

The root causes are:
1. `self.__dict__["scale"] = scale` and `self.__dict__["loc"] = loc` create 
a stale copy that never updates when `set_params` is called.
2. Unnecessary `@property`/setter boilerplate (~20 lines of dead code).
3. `loc` conflicts with `BaseDistribution`'s reserved read-only `loc` property.

**Environment**
- 
- Python 3.14
- skpro main branch
- scipy, sklearn latest

**Additional context**

Same `__dict__` hack pattern as Levy distribution fixed in #879. 
Fix: remove `__dict__` hack, remove property boilerplate, rename `loc → mu` 
(same as `Normal`, `Laplace`, `Levy`), pass `loc=mu` to all scipy calls.

direkkakkar319-ops:
#### Reference Issues/PRs
Fixes #889 the `__dict__` hack bug in `GeneralizedPareto`, same pattern as #879 (Levy fix).

`GeneralizedPareto.__init__` in `skpro/distributions/gen_pareto.py` had three problems:
1. `self.__dict__["scale"] = scale` and `self.__dict__["loc"] = loc` created a stale 
copy — after `set_params(scale=2.0)`, the property returns `2.0` but `__dict__["scale"]` 
stays `1.0`, so `sklearn.clone()` silently returns wrong parameters.
2. Unnecessary `@property`/setter boilerplate (~20 lines of dead code).
3. `loc` conflicts with `BaseDistribution`'s reserved read-only `loc` property.

**Fix Attempt 1 — Failed**
Renamed `loc → mu`, removed `__dict__` hack and property boilerplate. But in all 
scipy calls accidentally passed `mu=mu` instead of `loc=mu`:
```python
genpareto.ppf(p, c, mu=mu, scale=scale)  # WRONG
```
Screenshot 2026-03-13 235606

48 tests failed.

Scipy doesn't know a `mu` keyword — it always expects `loc=`. This caused:

TypeError: _parse_args() got an unexpected keyword argument 'mu'

**Fix Attempt 2 — Passed **
Kept `mu` as the internal parameter name but passed it to scipy as `loc=`:
```python
genpareto.ppf(p, c, loc=mu, scale=scale)  # CORRECT
```
Screenshot 2026-03-13 235617

The variable is named `mu` to avoid the `BaseDistribution` conflict, but scipy 
still receives it as `loc=`. 69 tests passed, 0 failed.

No new dependencies.

- The renaming of `loc → mu` and whether this is consistent with the rest of the codebase (same pattern as `Normal`, `Laplace`, `Levy`)
- The scipy calls passing `loc=mu` correctly

No new tests added — the existing 69 tests in `skpro/distributions/tests/test_all_distrs.py` 
all pass after the fix (previously broken due to the `__dict__` hack).

#### Any other comments?
This bug was discovered while scanning for the same `__dict__` pattern that was 
fixed in the Levy distribution (#879), using:
```bash
grep -rn "__dict__\[" skpro/distributions/ --include="*.py"
```

##### For all contributions
- [ ] I've added myself to the list of contributors with `bug` and `code` badges.
- [x] The PR title starts with [BUG].


fkiraly (changes requested):
Please look at what I did in #879 and replicate that - instead of, presumably, throwing an AI at this.

direkkakkar319-ops:
Hi @fkiraly! Understood -I will look at what you did in #879 and replicate the _ScipyAdapter pattern manually. 
I appreciate the feedback and will make sure I understand each change before pushing.

direkkakkar319-ops:
Hi @fkiraly! Here's what I changed and why:

1. Base class → _ScipyAdapter 
2. Added _get_scipy_object() and _get_scipy_param() — the two methods _ScipyAdapter requires
3. c gets a default value of 0.5 so the class can be instantiated without arguments
4. Added _var override to return np.inf instead of nan when variance is undefined (c >= 0.5)
5. Improved docstring with math formula, parameter descriptions and doctest example

I used AI as a tool to understand the pattern from #879, but I reviewed and understood each change before applying it.

All the test cases are passed for gen_pareto.py

image

fkiraly (approved):
Looks good now.

PR #924, opened 2026-03-14
[ENH] distribution fitters - framework and first examples

### LINKED ISSUE #897: [ENH] distribution fitters - framework and first examples
With scalar distributions being mature, it is now possible to introduce distribution fitters as a new object class, `"distfit"`.

A distribution fitter produces a single scalar distribution for a tabular dataset. It has the following methods:

* `fit(X, C=None)`, writes to `self`
* `proba()`, returns a scalar distribution fitted to `X` seen in `fit`, with optional censoring indicator `C`
* all other methods inherited from `skpro` `BaseEstimator` (not `sklearn` `BaseEstimator`

Four things should be added for the start:

* a base class, similar to `BaseProbaRegressor`, but mirroring this simpler design, in `distfitter`
* two example classes: `NormalFitter`, which fits a normal with sample mean and variance; and `MOMfitter(dist_cls, mean_name="mu", std_name=None)`, which computes mean and std and substitutes in the two variables. If `None`, checks `std_name` to be `"sigma"`, `"scale"`, if exists uses these.
    * optional: also interfacing the `distfit` package as a distribution fitter
* a test class, `TestAllDistFitters`, similar to the other classes; it also needs to be registered in `get_test_class_registry`
* an extension template

Good first issue but a hard one.

patelchaitany:
#### Reference Issues/PRs

Closes #897 

Below is the major changes which this PR introduce.

1. _base.py : Implemented BaseDistFitter which follows the same template-method pattern as BaseProbaRegressor. The public fit() method handles input validation and mytype conversion via _check_X/_check_C ,delegates to _fit() same for the proba() it also delegates to _proba(). Support for the censoring  via the "capability:survival" tag.

2. Two example of fitters : 

    1. _normalfitter.py :  NormalFitter, always fits a Normal distribution using sample mean and std
    2. _momfitter.py : MOMFitter, generic method-of-moments fitter. Accepts any distribution class (dist_cls), auto-detects "sigma" or "scale" parameter names, and supports a dist_params dict for passing extra fixed parameters (like truncation bounds or degrees of freedom.)

3. Tags : _tags.py added "distfitter" to the capability:survival tag's parent types, added X_inner_mtype and C_inner_mtype tag classes scoped to the "distfitter" scitype

No

Yes

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distfitter/_momfitter.py
+++ b/skpro/distfitter/_momfitter.py
@@ -183,6 +183,10 @@ class MOMFitter(BaseDistFitter):
             "std_name": "sigma",
             "dist_params": {"df": 5},
         }
-        params5 = {"dist": Laplace(mu=0.0, scale=1.0)}
-        params6 = {"dist": TDistribution(df=5), "mean_name": "mu", "std_name": "sigma"}
+        params5 = {"dist": Laplace(mu=0.0, scale=1.0), "mean_name": "mu"}
+        params6 = {
+            "dist": TDistribution(df=5, mu=0.0, sigma=1.0),
+            "mean_name": "mu",
+            "std_name": "sigma",
+        }
         return [params1, params2, params3, params4, params5, params6]
--- a/skpro/distfitter/_momfitter.py
+++ b/skpro/distfitter/_momfitter.py
@@ -182,8 +182,8 @@ class MOMFitter(BaseDistFitter):
             "mean_name": "mu",
             "std_name": "sigma",
             "dist_params": {"df": 5},
-        }
-        params5 = {"dist": Laplace(mu=0.0, scale=1.0), "mean_name": "mu"}
+        },
+        params5 = {"dist": Laplace(mu=0.0, scale=1.0), "mean_name": "mu"},
         params6 = {
             "dist": TDistribution(df=5, mu=0.0, sigma=1.0),
             "mean_name": "mu",
--- a/skpro/distfitter/_momfitter.py
+++ b/skpro/distfitter/_momfitter.py
@@ -18,8 +18,8 @@ class MOMFitter(BaseDistFitter):
 
     Parameters
     ----------
-    dist_cls : class
-        A distribution class from ``skpro.distributions``.
+    dist : skpro distribution instance or class
+        A distribution instance or class from ``skpro.distributions``.
         Must accept keyword arguments for location and scale parameters.
     mean_name : str, optional (default="mu")
         Name of the distribution parameter that corresponds to the mean.
@@ -27,7 +27,7 @@ class MOMFitter(BaseDistFitter):
         Name of the distribution parameter that corresponds to the
         standard deviation. If None, auto-detects by looking for
         ``"sigma"`` or ``"scale"`` in the ``__init__`` signature of
-        ``dist_cls``.
+        ``dist``.
     dist_params : dict or None, optional (default=None)
         Additional fixed parameters to pass to the distribution constructor.
         These are merged with the estimated mean and std parameters when
@@ -42,7 +42,7 @@ class MOMFitter(BaseDistFitter):
     >>> from skpro.distfitter import MOMFitter
     >>> from skpro.distributions.normal import Normal
     >>> X = pd.DataFrame([1.0, 2.0, 3.0, 4.0, 5.0])
-    >>> fitter = MOMFitter(dist_cls=Normal, mean_name="mu", std_name="sigma")
+    >>> fitter = MOMFitter(dist=Normal, mean_name="mu", std_name="sigma")
     >>> fitter.fit(X)
     MOMFitter(...)
     >>> dist = fitter.proba()
@@ -50,7 +50,7 @@ class MOMFitter(BaseDistFitter):
     Using Laplace distribution (uses "scale" instead of "sigma"):
 
     >>> from skpro.distributions.laplace import Laplace
-    >>> fitter = MOMFitter(dist_cls=Laplace, mean_name="mu", std_name="scale")
+    >>> fitter = MOMFitter(dist=Laplace, mean_name="mu", std_name="scale")
     >>> fitter.fit(X)
     MOMFitter(...)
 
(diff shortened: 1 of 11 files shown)


patelchaitany:
@fkiraly  could you pls review this pr.

fkiraly (changes requested):
Good start!

No new tags need to be added for `X_inner_mtype`, the range can just be extended to the new object type.

patelchaitany:
@fkiraly, Done - extended parent_type of the existing X_inner_mtype and C_inner_mtype tags to include "distfitter" instead. Removed the duplicate tag classes

patelchaitany:
@fkiraly, Done - extended parent_type of the existing X_inner_mtype and C_inner_mtype tags to include "distfitter" instead. Removed the duplicate tag classes

skpro/distfitter/_momfitter.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Returns
+        -------
+        params : dict
+            Parameter names mapped to their values.
+        """
+        return super().get_params(deep=False)
+
+    def _repr_html_(self):
+        """HTML representation, overridden to avoid recursion into dist_cls."""
```
fkiraly: I think we should not do this.

Can you explain why there would be recursion?

skpro/distfitter/_momfitter.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict
+            Parameters to create testing instances of the class.
+        """
+        from skpro.distributions.normal import Normal
+
+        params1 = {"dist_cls": Normal, "mean_name": "mu", "std_name": "sigma"}
+        params2 = {"dist_cls": Normal, "mean_name": "mu"}
```
fkiraly: include examples where the names are non-standard, or distributions other than `Normal`.

skpro/distfitter/_normalfitter.py
```diff
@@ -0,0 +1,79 @@
...
+from skpro.distfitter.base import BaseDistFitter
+
+
+class NormalFitter(BaseDistFitter):
+    r"""Fit a Normal distribution using sample mean and standard deviation.
+
+    Estimates the parameters :math:`\mu` and :math:`\sigma` of a Normal
+    distribution from the sample mean and sample standard deviation of the
+    data passed to ``fit``.
+
```
fkiraly, line 15: can you include at least a few standard strategies? I would add a parameter `method`, which can be

* `"unbiased"`, with denominator `N-1` as current (default)
* `"MLE"` - with denominator `N`
* optional: shrinkage, with a shrinkage target

fkiraly (approved):
Very nice! I left a few comments above, small ones.

skpro/distfitter/_momfitter.py
patelchaitany: You're right, "recursion" is the wrong term. The actual issue is a TypeError.

dist_cls is a class (e.g., Normal), not an instance. Since Normal inherits from BaseObject, hasattr(Normal, "get_params") is True. The base class get_params(deep=True) then calls Normal.get_params() on the class itself, which fails with TypeError: missing required positional argument 'self'.

The same crash occurs through the _repr_html_ path - _get_visual_block calls get_params() and then tries to render dist_cls as a nested component.

fkiraly: oh, ok, this is a general issue then with the html repor of `scikit-base` classes, is it? I would suggest to try addressing this in `scikit-base` then - not necessarily with a PR; but reporting the problem as an issue there.

fkiraly: FYI @mateuszkasprowicz

fkiraly:
From discussion on 2026-05-29:

* please report the issue with the report on `scikit-base`
* please also give an explanation of what exactly happens if we do not override `__repr__` - optimally also write some example code for `scikit-base`, e.g., a simple class inheriting `BaseObject`, where the problem occurs

patelchaitany:
@fkiraly — following up on the May 29 comment.

I’ll open an issue on scikit-base for this. The bug is in BaseObject.get_params(deep=True): it treats any __init__ parameter that has a get_params attribute as a nested component and calls get_params() on it. That is wrong when the parameter is a class (a BaseObject subclass), not an instance. In that case scikit-base ends up calling SomeClass.get_params() without self, which raises TypeError: missing 1 required positional argument: 'self'. The same pattern appears in the HTML diagram repr path (_object_html_repr / _get_visual_block), which also calls get_params() with deep=True by default.

On this PR, MOMFitter stores dist_cls as a distribution class (e.g. Normal) in _momfitter.py. Normal inherits from BaseObject, so it triggers exactly that logic. Without the overrides in _momfitter.py (get_params forcing deep=False and the custom _repr_html_), fitter.get_params(deep=True) fails, generic estimator tests that call get_params() can fail, and Jupyter’s diagram display can fail. Plain text repr(fitter) usually still works because it uses get_params(deep=False). The overrides are a temporary workaround until scikit-base only descends into instances, not classes.

Minimal repro for the upstream issue and a skpro-specific repro are below.
```
from skbase import BaseObject

class Child(BaseObject):
    def __init__(self, x=1):
        self.x = x
        super().__init__()

class Parent(BaseObject):
    def __init__(self, nested_cls):
        self.nested_cls = nested_cls
        super().__init__()

Parent(nested_cls=Child).get_params(deep=True)
```

patelchaitany:
@fkiraly I’ve opened a fix upstream in scikit-base: [sktime/skbase#559](https://github.com/sktime/skbase/pull/559). Could you take a look when you have a chance and let me know if any changes are needed there?

Once that PR is merged and released (or we bump the scikit-base dependency here), I’ll update this PR to drop the local workarounds in MOMFitter (get_params / _repr_html_ overrides in skpro/distfitter/_momfitter.py) and rely on the upstream fix instead.

patelchaitany:
@fkiraly , Is there any changes are required.

fkiraly (approved):
Not at all - I think this is a brilliant PR, but it took time to review - it adds a new object type so we need to get it right, so I had to give it an extensive review.

I made only small changes - allowing object instances in `MOMFitter`, which:
* is more coherent design-wise, generally having  `sth_class` and `sth_kwargs` argument in the constructor indicates that instead the user may want to pass an instance.
* allows us to revert the `scikit-base` lower bound increase.

PR #926, opened 2026-03-15
fix: add check_is_fitted() to predict and predict_proba

ashnaaseth2325-oss:
## SUMMARY

This PR adds a missing fitted state check to `predict()` and `predict_proba()` in `skpro/regression/base/_base.py`. Without this check, calling these methods on an unfitted estimator raises an `AttributeError` instead of the expected `NotFittedError`.
The change makes their behavior consistent with other prediction methods.

---

## FIX

**Before**

```python
def predict(self, X):
    X = self._check_X(X)
```

**After**

```python
def predict(self, X):
    self.check_is_fitted()
    X = self._check_X(X)
```

---

## VERIFICATION

**Steps**

```python
reg = ResidualDouble(LinearRegression())
reg.predict(X)
```

**Expected**

Raises `NotFittedError` instead of `AttributeError`.

### DIFF
--- a/skpro/regression/base/_base.py
+++ b/skpro/regression/base/_base.py
@@ -223,6 +223,8 @@ class BaseProbaRegressor(BaseEstimator):
         y : pandas DataFrame, same length as `X`
             labels predicted for `X`
         """
+        self.check_is_fitted()
+
         X = self._check_X(X)
 
         y_pred = self._predict(X)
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @fkiraly @felipeangelimvieira,
Can you please review this PR.
Thanks!

ashnaaseth2325-oss:
Hello @fkiraly 
Just a gentle follow-up on this PR. 
Thanks!

PR #930, opened 2026-03-15
[BUG] Fix discrete PMF plotting to use stem plots

### LINKED ISSUE #918: [BUG] `test_discrete_pmf_plotting` is failing
The `test_discrete_pmf_plotting` test for the new `pmf` plotting method introduced in https://github.com/sktime/skpro/pull/714 is failing.

The reason is unclear, this should be diagnosed and fixed.
Here is the printout:

```
FAILED skpro/distributions/tests/test_proba_basic.py::test_discrete_pmf_plotting - AssertionError: Should plot at multiple support points
assert 3 > 5
 +  where 3 = len([<matplotlib.lines.Line2D object at 0x7fcc852781a0>, <matplotlib.collections.LineCollection object at 0x7fcc852a8410>, <matplotlib.lines.Line2D object at 0x7fcc852782f0>])
 +    where [<matplotlib.lines.Line2D object at 0x7fcc852781a0>, <matplotlib.collections.LineCollection object at 0x7fcc852a8410>, <matplotlib.lines.Line2D object at 0x7fcc852782f0>] = get_children()
 +      where get_children = <StemContainer object of 3 artists>.get_children
```

kabirvashisht4-glitch:
Reference Issues/PRs
Fixes #918

What does this implement/fix?

After investigating the root cause of the failing test, we found that 
the real issue was never in the test itself but in _plot_single in 
_base.py. The method always calls ax.plot() regardless of whether the 
distribution is discrete or continuous. This means for discrete 
distributions the plot output goes into ax.lines not ax.containers, 
so ax.containers is always empty and the test throws an IndexError 
before any assertion is even reached.

We also found that the default x bounds from ppf(0.001) and ppf(0.999) 
cut off the tail support points for Binomial(n=10), returning [1..9] 
instead of the full [0..10] support.

Changes made:
- Modified _plot_single in _base.py to detect discrete distributions 
  and use ax.stem() via _pmf_support() for PMF plots
- Continuous distributions and all other plot types keep using ax.plot()
- Added test_discrete_pmf_plotting in test_proba_basic.py with explicit 
  x_bounds to verify all 11 support points are plotted for Binomial(n=10)

Does your contribution introduce a new dependency?
No.

Did you add any tests?
Yes.

### DIFF
--- a/skpro/survival/adapters/_common.py
+++ b/skpro/survival/adapters/_common.py
@@ -104,17 +104,11 @@ def _get_fitted_params_default_safe(obj=None):
         attr for attr in dir(obj) if attr.endswith("_") and not attr.startswith("_")
     ]
 
-    def hasattr_safe(obj, attr):
+    fitted_param_dict = {}
+    for p in fitted_params:
         try:
-            if hasattr(obj, attr):
-                getattr(obj, attr)
-                return True
+            fitted_param_dict[p[:-1]] = getattr(obj, p)
         except Exception:
-            return False
-
-    # remove the "_" at the end
-    fitted_param_dict = {
-        p[:-1]: getattr(obj, p) for p in fitted_params if hasattr_safe(obj, p)
-    }
+            continue
 
     return fitted_param_dict


kabirvashisht4-glitch:
The test-est (CoxNet) failure is a pre-existing issue unrelated to this PR — No module named 'skpro.utils.validation' appears on other PRs as well. All checks related to my changes pass.

PR #932, opened 2026-03-15
[ENH] HalfT (Half Student-t) distribution

OmAmbole009:
#### What does this implement/fix?

Implements the Half Student-t (`HalfT`) distribution, towards #22.

The Half-t distribution is a Student-t distribution truncated to 
non-negative values. It generalizes both:
- Half-Normal (as `df` -> infinity)
- Half-Cauchy (when `df = 1`)

#### Parameters
- `df` : degrees of freedom (must be positive)
- `sigma` : scale parameter (must be positive)

#### Methods implemented
- `mean`, `var`
- `pdf`, `log_pdf`
- `cdf`, `ppf`

#### Testing
All 103 tests pass locally.

#### References
Towards #22


fkiraly (changes requested):
Thanks! Can you please make sure you do not delete exports for the `Rayleigh` distribution?

PR #940, opened 2026-03-16
[BUG] fix HalfNormal, HalfCauchy, HalfLogistic passing scale as positional arg

### LINKED ISSUE #939: [BUG] HalfNormal, HalfCauchy, HalfLogistic return wrong values for non-default scale
**Describe the bug**

Three distributions — HalfNormal, HalfCauchy, and HalfLogistic — return mathematically incorrect results for pdf(), cdf(), ppf(), mean(), and var() whenever their scale parameter is set to anything other than 1.0.
The root cause is in _get_scipy_param() in all three files — they return the scale parameter as a positional argument instead of a keyword argument. The _ScipyAdapter base class unpacks these as obj.pdf(x, *args, **kwds). Since scipy.stats.halfnorm, halfcauchy, and halflogistic have zero shape parameters, scipy maps the first positional argument after x to loc, not scale. So HalfNormal(sigma=2.0).pdf(x) silently becomes halfnorm.pdf(x, loc=2.0, scale=1.0) — shifting the distribution instead of scaling it. For any x < loc, the pdf returns 0.0 with no error or warning. The bug went unnoticed because the default value is 1.0, which does not produce obviously wrong results in basic tests.

**To Reproduce**

import pandas as pd
from skpro.distributions.halfnormal import HalfNormal
from skpro.distributions.halfcauchy import HalfCauchy
from skpro.distributions.halflogistic import HalfLogistic
from scipy.stats import halfnorm, halfcauchy, halflogistic

x_df = pd.DataFrame({"y": [1.0]})
idx, col = pd.Index([0]), pd.Index(["y"])

HalfNormal(sigma=2.0, index=idx, columns=col).pdf(x_df).iloc[0, 0]
skpro → 0.0  |  scipy halfnorm.pdf(1.0, scale=2.0) → 0.352065

HalfCauchy(beta=2.0, index=idx, columns=col).pdf(x_df).iloc[0, 0]
skpro → 0.0  |  scipy halfcauchy.pdf(1.0, scale=2.0) → 0.254648

HalfLogistic(beta=2.0, index=idx, columns=col).pdf(x_df).iloc[0, 0]
skpro → 0.0  |  scipy halflogistic.pdf(1.0, scale=2.0) → 0.235004

A screenshot of the reproduction output is attached below.
Image
All three skpro distributions return 0.0. All three scipy direct calls return the correct positive value.

**Expected behavior**

HalfNormal(sigma=2.0).pdf(1.0) should return ~0.352, not 0.0
HalfCauchy(beta=2.0).pdf(1.0) should return ~0.255, not 0.0
HalfLogistic(beta=2.0).pdf(1.0) should return ~0.235, not 0.0

**Environment**
OS: macOS
Python: 3.x
skpro: main branch (latest)
scipy: latest

**Additional context**

Affected files and lines:

skpro/distributions/halfnormal.py line 66
skpro/distributions/halfcauchy.py line 70
skpro/distributions/halflogistic.py line 71

The silent failure (returning 0.0 instead of raising an error) makes this particularly dangerous — users relying on these distributions with non-default scale values get wrong results with no indication anything is broken.
The fix is a one-line change in each of the three files — pass the scale parameter as a keyword argument ({"scale": sigma}) instead of a positional argument ([sigma]) in _get_scipy_param(). This is consistent with how Cauchy in cauchy.py already handles it correctly.
If confirmed, I will be opening a PR to fix all three files.

ANANYA542:
##Description
These three distributions passed their scale parameter as a positional argument in _get_scipy_param(). Since scipy.stats.halfnorm, halfcauchy, and halflogistic have no shape parameters, scipy interpreted the positional arg as loc instead of scale. This caused pdf, cdf, ppf, mean, and var to return wrong values for any non-default parameter value.
Fix: pass the parameter as scale keyword argument instead.

#### Reference Issues/PRs
 fixes #939 

HalfNormal, HalfCauchy, and HalfLogistic were passing their scale parameter as a positional argument in _get_scipy_param(). Since scipy.stats.halfnorm, halfcauchy, and halflogistic have no shape parameters, scipy was interpreting that positional argument as loc instead of scale.
This meant all three distributions silently returned wrong values for pdf, cdf, ppf, mean, and var whenever the parameter was anything other than 1.0. For example, HalfNormal(sigma=2.0).pdf(1.0) returned 0.0 instead of ~0.352.
The fix changes _get_scipy_param() in all three files from:
pythonreturn [param], {}   # param incorrectly treated as loc
to:
pythonreturn [], {"scale": param}   # param correctly passed as scale
This is consistent with how Cauchy and other distributions in skpro already handle this correctly.

No new tests were added. The distributions are already covered by the automated test suite via their get_test_params() methods. The existing tests were passing because they primarily check API contracts (types/shapes) rather than mathematical equivalence to scipy's output. The fix corrects the parameter mapping without breaking any existing tests.

#### Any other comments?
Checked all distributions using _ScipyDist — only these three are affected because they are the only ones whose scipy counterparts have zero shape parameters, making the positional-vs-keyword distinction critical.
A screenshot confirming the fix — showing the correct values being returned after the change — is attached below.

##### For all contributions

##### For new estimators
- [ ] If the estimator relies on a 
soft dependency, I've set the `python_dependencies` tag and ensured

Screenshot 2026-03-16 at 1 15 10 PM


ANANYA542:
@fkiraly could u just review this pr under issue #939 and let me know if some changes or explanation required

fkiraly (approved):
Nice, thanks.

Regression tests would have been nice, but there is another PR adding that, so I am merging this.

PR #956, opened 2026-03-16
Prevent zero-variance instability in BaseProbaRegressor.predict_proba

### LINKED ISSUE #955: [BUG] Zero predicted variance causes instability in BaseProbaRegressor.predict_proba
### Description

When `BaseProbaRegressor.predict_proba` falls back to constructing a **Normal distributio**n using **predict_var**, a predicted variance of **0** leads to **sigma = 0** in the Normal distribution.

This causes divide-by-zero warnings and NaN values when evaluating pdf or log_pdf.

**Example flow:**
```
pred_var = self.predict_var(X)
pred_std = np.sqrt(pred_var)
Normal(mu=pred_mean, sigma=pred_std)
```

If `pred_var == 0`, downstream probability evaluations become unstable.

### To Reproduce

This can happen with deterministic regressors where _predict_var returns 0.

**Example scenario:**

```
class MockZeroVarRegressor(BaseProbaRegressor):

    def _predict(self, X):
        return np.zeros(len(X))

    def _predict_var(self, X):
        return np.zeros(len(X))
```

**Calling:**

model.predict_proba(X).pdf(0)

can produce warnings and NaN values.

kindler-king:
**Reference Issues/PRs**
Fixes #955 

**What does this implement/fix?** 

This PR fixes a numerical instability in `BaseProbaRegressor.predict_proba`.

When `predict_var` returns 0, the fallback Normal distribution is constructed with `sigma=0`, which leads to divide-by-zero warnings and **NaN** values when evaluating pdf or log_pdf.

To prevent this, the predicted variance is clipped to machine epsilon before computing the standard deviation:

`pred_var = np.clip(pred_var, np.finfo(float).eps, None)
`
This ensures the resulting Normal distribution always has a strictly positive scale while leaving normal model outputs effectively unchanged.

**Does your contribution introduce a new dependency?**
No.

- Whether clipping variance at machine epsilon is the appropriate safeguard.
- Consistency with the existing probabilistic regression design.

Yes.

A regression test was added that uses a mock regressor returning zero variance and verifies that:
`predict_proba().pdf()` and `log_pdf()` remain finite
no numerical warnings are raised

### DIFF
--- a/skpro/regression/base/_base.py
+++ b/skpro/regression/base/_base.py
@@ -333,18 +333,37 @@ class BaseProbaRegressor(BaseEstimator):
 
         # we use predict_var to get scale, and predict to get location
         pred_var = self.predict_var(X=X)
-        pred_std = np.sqrt(pred_var)
         pred_mean = self.predict(X=X)
-
-        from skpro.distributions.normal import Normal
 
         if hasattr(X, "index"):
             index = X.index
         else:
             index = pd.RangeIndex(start=0, stop=len(X), step=1)
         columns = self._get_columns(method="predict")
-        pred_dist = Normal(mu=pred_mean, sigma=pred_std, index=index, columns=columns)
 
+        eps = np.finfo(float).eps
+
+        # if all predicted variances are below machine epsilon,
+        # predictions are deterministic - return a Delta distribution
+        if (pred_var.values < eps).all():
+            from skpro.distributions.delta import Delta
+
+            return Delta(c=pred_mean, index=index, columns=columns)
+
+        # for non-degenerate predictions, construct a Normal
+        # clamp any individual near-zero variance entries to eps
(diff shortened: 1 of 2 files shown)


kindler-king:
Hello @fkiraly , thanks for the suggestion, I’ve updated the implementation.

- All-zero variance now returns a `Delta` distribution.
- Non-zero variance continues to return `Normal`.

For the mixed case, I explored returning per-row heterogeneous distributions, but according to my understanding, skpro doesn’t support that (no concat in `BaseDistribution`, and `Mixture` uses global weights).

So for now, I fall back to` Normal` with `eps` clamping for zero-variance entries to maintain numerical stability.

Also fixed formatting issues using pre-commit.

Would love your feedback on this approach.

PR #973, opened 2026-03-19
[BUG] Fix incorrect variance calculation and capability tagging in Hurdle

### LINKED ISSUE #972: [BUG] Incorrect variance computation and misleading capability tags in Hurdle
**Describe the bug**
The Hurdle distribution contains two independent but related issues:

1. Incorrect variance formula in _var The current implementation computes: p * var_positive + p * (1 - p) * mean_positive However, the correct variance for a hurdle distribution is:
$$
\mathrm{Var}(Y) = p\sigma^2 + p(1-p)\mu^2
$$
The mean term is not squared in the implementation, violating the law of total variance and leading to severely underestimated variance.

2. Incorrect capability tagging Hurdle determines whether mean and var are exact by inspecting the base distribution (e.g., Normal). However, internally it wraps this distribution in LeftTruncated, which: 
does not implement exact mean/var instead relies on numerical integration (PPF-based approximation)
As a result: Hurdle advertises exact capabilities but actually produces approximate results and triggers runtime warnings This leads to misleading API behavior and inconsistent capability metadata.

**To Reproduce**
```python
import numpy as np
from skpro.distributions.hurdle import Hurdle

base = Normal(mu=10.0, sigma=1.0)
hurdle = Hurdle(p=0.5, distribution=base)

np.random.seed(42)
samples = hurdle.sample(100_000)
empirical_var = float(samples.var().iloc[0])

skpro_var = float(hurdle.var().iloc[0, 0])

print("Empirical variance:", empirical_var)
print("skpro variance:", skpro_var)
```

screenshot of the issue is attached below:

Image

**Expected behavior**
Variance should follow:

$$
\mathrm{Var}(Y) = p\sigma^2 + p(1-p)\mu^2
$$

Empirical and analytical variance should match within numerical tolerance Hurdle should correctly advertise mean and var as approximate, not exact

**Environment**
OS: macOS
Python: 3.x
skpro: latest (main branch)
NumPy / Pandas: standard versions

**Additional context**
The discrepancy is structural, not due to numerical approximation The related ZeroInflated distribution correctly implements this variance formula, indicating inconsistency
The issue impacts: uncertainty estimation ,probabilistic metrics ,downstream model evaluation

**Proposed fix**
Correct variance formula:
return ( self.p * var_positive+ (mean_positive ** 2) * self.p * (1.0 - self.p))
Fix capability tagging. Inspect LeftTruncated(distribution) instead of the raw base distribution Downgrade mean and var to approximate capabilities

ANANYA542:
#### Reference Issues/PRs
fixes #972 

This PR fixes an issue in the `Hurdle` distribution where the variance was being computed incorrectly.  
The current implementation uses `p * var_positive + p * (1 - p) * mean_positive`, but the mean term should be squared. This was leading to significant underestimation of the variance. The formula has been corrected to use `(mean_positive ** 2)` in line with the law of total variance.

I also updated how capability tags are determined. Previously, `Hurdle` checked the base distribution to decide whether `mean`/`var` are exact, but since it internally wraps the distribution using `LeftTruncated`, those values are actually computed via numerical approximation. The tags now reflect this correctly by checking the truncated distribution instead.

No

- Whether the updated variance formula aligns with expected behavior across different base distributions  
- Whether capability tagging should consistently be based on the internally used truncated distribution  

I verified the fix using Monte Carlo simulation (100k samples), where the empirical variance now matches the analytical result within tolerance. No new formal unit tests have been added yet.

#### Any other comments?

A screenshot showing the before/after behavior has also been added for clarity.

Screenshot 2026-03-20 at 2 23 44 AM

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/hurdle.py
+++ b/skpro/distributions/hurdle.py
@@ -85,7 +85,10 @@ class Hurdle(BaseDistribution):
         # if and only if the capabilities of the inner distribution are exact
         self_exact_capas = self.get_tag("capabilities:exact", []).copy()
         self_approx_capas = self.get_tag("capabilities:approx", []).copy()
-        distr_exact_capas = distribution.get_tag("capabilities:exact", []).copy()
+        
+        # We must inherit exactness from the truncated distribution, NOT the unbound base
+        distr_exact_capas = self._truncated_distribution.get_tag("capabilities:exact", []).copy()
+        
         for capa in self_exact_capas:
             if capa not in distr_exact_capas:
                 self_exact_capas.remove(capa)
@@ -181,7 +184,7 @@ class Hurdle(BaseDistribution):
         mean_positive = self._truncated_distribution.mean()
         var_positive = self._truncated_distribution.var()
 
-        return self.p * var_positive + mean_positive * self.p * (1.0 - self.p)
+        return self.p * var_positive + (mean_positive ** 2) * self.p * (1.0 - self.p)
 
     def _ppf(self, p):
         prob_zero = 1.0 - self.p


ANANYA542:
Hi @fkiraly  — just following up on this PR. If you could take some time to review it and let me know if any changes or explanation is required, I’d really appreciate it.

PR #979, opened 2026-03-20
[BUG] Fix backward independence routing in Mixture sample

### LINKED ISSUE #974: [BUG] Mixture._sample handles indep_cols and indep_rows backward, destroying independence assumptions
**Describe the bug**
In skpro.distributions.mixture.Mixture, the arguments indep_rows and indep_cols are intended to control whether rows and columns are sampled independently from mixture components (defaulting to True).
However, in the _sample method (around line 224), the conditional logic handling these flags is inverted:
```python
if indep_rows:
    rd_size[0] = 1
if indep_cols:
    rd_size[1] = 1
```
When indep_rows=True, this sets the sampling size along the row dimension to 1. This forces np.random.choice to draw a single component index, which is then broadcast across all rows — making them fully dependent.
Conversely, when indep_rows=False, the full dimension is retained, allowing independent sampling across rows when dependence was intended.

**To Reproduce**
```python
**To Reproduce**
```python
import numpy as np
import pandas as pd
from skpro.distributions.mixture import Mixture

mu1 = np.full((N_ROWS, 2), 100.0)
mu2 = np.full((N_ROWS, 2), -100.0)

n1 = Normal(mu=mu1, sigma=0.01)
n2 = Normal(mu=mu2, sigma=0.01)

# TEST 1: Request Independent Rows, Dependent Cols
mixture_1 = Mixture(
    distributions=[n1, n2], weights=[0.5, 0.5], indep_rows=True, indep_cols=False
)
np.random.seed(42)
sample_1 = np.round(mixture_1.sample(1).values)
unique_rows_1 = len(np.unique(sample_1, axis=0))

print("--- TEST 1: indep_rows=True, indep_cols=False ---")
print(f"EXPECTED: ~20 unique rows | ACTUAL: {unique_rows_1}")
print(sample_1[:5])

# TEST 2: Request Dependent Rows, Independent Cols
mixture_2 = Mixture(
    distributions=[n1, n2], weights=[0.5, 0.5], indep_rows=False, indep_cols=True
)
np.random.seed(13)
sample_2 = np.round(mixture_2.sample(1).values)
unique_rows_2 = len(np.unique(sample_2, axis=0))

print("\n--- TEST 2: indep_rows=False, indep_cols=True ---")
print(f"EXPECTED: 1 unique row | ACTUAL: {unique_rows_2}")
print(sample_2[:5])

```
Observed behavior
SKPRO OUTPUT (indep_rows=True):
[[ 100. -100.]
 [ 100. -100.]
 [ 100. -100.]
 [ 100. -100.]
 [ 100. -100.]]

SKPRO OUTPUT (indep_rows=False):
[[-100. -100.]
 [ 100.  100.]
 [-100. -100.]
 [-100. -100.]
 [-100. -100.]]

indep_rows=True → rows are identical (dependent) 
indep_rows=False → rows are independent 
This indicates the independence logic is reversed.
A screenshot showing the same behavior has been attached below for reference.

Image

**Expected behavior**
1. When indep_rows=True, rows should sample mixture components independently 
2. When indep_rows=False, rows should share a single component (broadcast behavior)

**Environment**
- OS: macOS  
- Python: 3.x  
- skpro: latest (main branch)  
- NumPy: standard version  

**Additional context**
**Proposed solution**
Invert the condition in _sample:
```python 
if not indep_rows:
    rd_size[0] = 1
if not indep_cols:
    rd_size[1] = 1
```
This ensures: independence → full sampling dimension ; dependence → broadcasted single component

ANANYA542:
#### Reference Issues/PRs

fixes #974 

This PR fixes an issue in `Mixture._sample` where the `indep_rows` and `indep_cols` flags were handled in reverse.

Previously, the implementation set `rd_size[...] = 1` when the independence flags were `True`, which caused `np.random.choice` to draw a single component and broadcast it across the dimension, making it fully dependent. When the flags were `False`, the full dimension was retained, resulting in independent sampling instead.

This PR corrects the logic by inverting the condition:

```python
if not indep_rows:
    rd_size[0] = 1
if not indep_cols:
    rd_size[1] = 1
```

This restores the intended behavior where independence retains full sampling dimensions and dependence enforces broadcasting.
No

- Whether the updated logic correctly reflects the intended behavior of `indep_rows` and `indep_cols`
- Whether the change is consistent with other sampling implementations in the codebase

Tested locally using larger mixture arrays (e.g., 20x2). The output matches expectations:
- dependent settings produce identical rows
- independent settings produce varied rows

#### Any other comments?

A verification screenshot demonstrating the behavior has also been attached for reference.
Screenshot 2026-03-20 at 11 48 05 AM

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/mixture.py
+++ b/skpro/distributions/mixture.py
@@ -221,9 +221,9 @@ class Mixture(BaseMetaObject, BaseDistribution):
         rd_size[0] *= N
         full_size[0] *= N
 
-        if indep_rows:
+        if not indep_rows:
             rd_size[0] = 1
-        if indep_cols:
+        if not indep_cols:
             rd_size[1] = 1
 
         n_dist = len(self._distributions)


ANANYA542:
@fkiraly —thanks for reviewing
this comes down to how rd_size controls the shape of the component draw before it gets broadcast.If a dimension is set to 1, np.random.choice draws a single component which then gets broadcast across that axis, making it dependent.
In the current implementation this happens when indep_rows=True, so all rows end up sharing the same component, which is the opposite of what we want.Flipping it to if not indep_rows keeps the full dimension when independence is requested, and only collapses it when dependence is intended. Same applies to columns as well.

ANANYA542:
Hi @fkiraly — just following up on this. I’ve addressed your question above and clarified the reasoning behind the change.Could you please take another look when you have time and let me know if any further clarification or changes are needed?

PR #983, opened 2026-03-20
 [ENH] Addition of Set representation and support-aware plotting

### LINKED ISSUE #244: [ENH] inspectable set-valued domains for distributions

Khushmagrawal:
#### Reference Issues/PRs
Fixes #244

I have implemented hierarchy of set classes to represent the mathematical support of distributions: `RealSet`, `IntegerSet`, `IntervalSet`, `FiniteSet`, and `EmptySet`, `UnionSet` and `IntersectionSet`. These sets are designed to be tabular, meaning a single set object can store parameters (bounds/values) as arrays 

The `BaseDistribution` now includes a `support` property). This would allow distributions to programmatically communicate where their probability mass/density resides. This method was overrided currently only in few distributions to test different types of set  - `Binomial` (IntegerSet), `Poisson` (IntegerSet), `Empirical` (FiniteSet), `TruncatedDistribution` (IntersectionSet), and `ZeroInflated` (UnionSet) have been updated.

The `plot` functionality has been changed to use `support.boundary()` to determine exact x-axis limits.  For discrete and mixed distributions, the plotter now queries the support for FiniteSet or IntegerSet components to identify exact coordinates for stem plots.

-  `_plot_single`, mixed distributions (like ZeroInflated) for pdf currently plot the continuous part as a line and the discrete mass as stems. Mathematically, a PDF of a mixed distribution includes Dirac delta spikes (infinite density). To distinguish Mass from Density, we mask the PDF line at spike points using np.isclose, would that be right?
- Broadcasting logic in  `_get_bc_params`  (We currently utilize `oned_as="col"` to force 1D parameter arrays  ) into column vectors `(N, 1)`)
- In UnionSet and IntersectionSet, we allow a mix of scalar and tabular children. We currently force the output to match the parent set's ndim
- Are the current Set types sufficient?

Yes, `test_set.py` has been added to verify the symbolic set logic

#### Any other comments?
Includes a minor fix for `ZeroInflated._iloc` to ensure scalar-scalar indexing correctly delegates to `_iat`, which was necessary for cell-level plotting in array distributions
##### For all contributions

### DIFF
--- a/skpro/distributions/base/__init__.py
+++ b/skpro/distributions/base/__init__.py
@@ -2,8 +2,30 @@
 # copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
 # adapted from sktime
 
-__all__ = ["BaseDistribution", "_DelegatedDistribution", "_BaseArrayDistribution"]
+__all__ = [
+    "BaseDistribution",
+    "_DelegatedDistribution",
+    "_BaseArrayDistribution",
+    "BaseSet",
+    "IntervalSet",
+    "FiniteSet",
+    "IntegerSet",
+    "UnionSet",
+    "IntersectionSet",
+    "EmptySet",
+    "RealSet",
+]
 
 from skpro.distributions.base._base import BaseDistribution
 from skpro.distributions.base._base_array import _BaseArrayDistribution
 from skpro.distributions.base._delegate import _DelegatedDistribution
+from skpro.distributions.base._set import (
+    BaseSet,
+    EmptySet,
+    FiniteSet,
+    IntegerSet,
+    IntersectionSet,
+    IntervalSet,
+    RealSet,
+    UnionSet,
+)
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -1355,6 +1355,44 @@ class BaseDistribution(BaseObject):
         else:
             return spl.mean().iloc[0]
 
+    @property
+    def support(self):
+        """Return the support of the distribution as a symbolic set.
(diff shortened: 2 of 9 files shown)


Khushmagrawal:
Adding two plotting examples here as well, the current plotting shows the continuous density and the point mass at zero separately and the support is `{0}` union `R`.

zin

For zero-inflated Poisson, the support is `{0}`  union the inner discrete support, and the plot shows the added mass at zero

zip

PR #995, opened 2026-03-25
[ENH] add truncated_mean interface to BaseDistribution

### LINKED ISSUE #221: [ENH] adding truncated means as an interface point?
API design and math question: would it make sense to add truncated means as general interface points? Or, more generally, truncated moments?

API design-wise, having it would help with:

* computation of energy, one of the terms is explicit in cdf and truncated means, see here: https://github.com/sktime/skpro/issues/22#issuecomment-2016599132
* truncation compositor, which then would have access to more analytical formulae

questions:

* if we add it, how? I would add additional arguments to the existing `mean` method.

Math-wise, the questions are:

* how easy are truncated mean and moments to obtain - matching against current literature coverage, and "potential" coverage in terms of analytic forms
* what are the analytic relations between these, and other distribution defining functions or methods currently present?

patelchaitany:
#### Reference Issues/PRs

Addresses #221. Closes #221.

Adds truncated_mean(lower, upper) to BaseDistribution. Returns E[X | lower < X < upper]. Closed-form implementations for Normal, Exponential, Laplace, and Uniform. Other distributions fall back to a ppf-based numerical approximation.

The base class _energy_x now uses truncated means when available: if a distribution has exact truncated_mean and cdf, it computes E[|X - c|] from the energy identity instead of Monte Carlo sampling.

TruncatedDistribution gets a _mean() that calls the inner distribution's truncated_mean, so TruncatedDistribution(Normal(...), lower=0).mean() is exact now.

#### Changes

- `base/_base.py` - new `truncated_mean` / `_truncated_mean` with ppf/MC default, updated `_energy_x`
- `normal.py` - exact `_truncated_mean`
- `exponential.py` - exact `_truncated_mean`
- `laplace.py` - exact `_truncated_mean`
- `uniform.py` - exact `_truncated_mean`
- `truncated.py` - `_mean()` via inner distribution's `truncated_mean`, updated tag logi

No

No
##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/distributions/base/_base.py
+++ b/skpro/distributions/base/_base.py
@@ -37,7 +37,7 @@ class BaseDistribution(BaseObject):
         "approx_spl": 1000,  # sample size used in other MC estimates
         "bisect_iter": 1000,  # max iters for bisection method in ppf
         # which methods are approximate (not numerically exact) should be listed here
-        "capabilities:approx": ["energy", "mean", "var", "pdfnorm"],
+        "capabilities:approx": ["energy", "mean", "var", "pdfnorm", "truncated_mean"],
         # broadcasting and parameter settings
         # -----------------------------------
         # used to control broadcasting of parameters
@@ -1244,7 +1244,48 @@ class BaseDistribution(BaseObject):
         -------
         2D np.ndarray, same shape as ``self``
             energy values w.r.t. the given points
+
+        Notes
+        -----
+        Uses an analytical shortcut when both ``"truncated_mean"`` and
+        ``"cdf"`` appear in the subclass's ``capabilities:exact`` tag.
+        In that case, the energy is computed via:
+
+        .. math::
+
+            \mathbb{E}[|X - c|]
+            = (\mathbb{E}[X \mid X > c] - c)(1 - F(c))
+            + (c - \mathbb{E}[X \mid X < c]) F(c)
+
+        When either capability is missing, falls back to Monte Carlo
+        via ``_energy_default``.
         """
+        exact_capas = self.get_tag("capabilities:exact", [])
+        has_exact_trunc = "truncated_mean" in exact_capas
+        has_exact_cdf = "cdf" in exact_capas
+
+        if has_exact_trunc and has_exact_cdf:
+            cdf_x = self._cdf(x)
+            x_float = np.asarray(x, dtype=float)
+            inf_arr = np.full_like(x_float, np.inf)
+            neg_inf_arr = np.full_like(x_float, -np.inf)
+
+            right_mean = self._truncated_mean(x_float, inf_arr)
+            left_mean = self._truncated_mean(neg_inf_arr, x_float)
+
+            right_mass = 1 - cdf_x
+            left_mass = cdf_x
+            zero_right = right_mass < 1e-15
+            zero_left = left_mass < 1e-15
+
+            right_term = np.where(zero_right, 0.0, (right_mean - x_float) * right_mass)
+            left_term = np.where(zero_left, 0.0, (x_float - left_mean) * left_mass)
+            energy_arr = right_term + left_term
+
+            if energy_arr.ndim > 0:
+                energy_arr = np.sum(energy_arr, axis=1)
+            return energy_arr
+
         return self._energy_default(x)
 
     def _energy_default(self, x=None):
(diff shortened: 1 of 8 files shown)


JiwaniZakir (commented):
The sampling-based fallback in `_truncated_mean` calls `self.sample(1)` inside a loop of `approx_spl_size` (default 1000) iterations, which is orders of magnitude slower than calling `self.sample(approx_spl_size)` once and filtering. The existing `_mean` fallback at the end of the file does exactly this with `_sample_mean`, so the pattern is already established — the loop approach here is inconsistent with the rest of the codebase and will be a significant performance bottleneck.

In `_energy_x`, the added docstring block (around line 1235) describes the truncated-mean identity used in the new code path, but it's inserted into the *existing* `_energy_x` docstring rather than being clearly separated from the original parameter/return docs. This makes the docstring structurally confusing — the formula section appears before the "Parameters" block, which is fine, but it references `_truncated_mean` and `cdf` without explaining that this only triggers when both are exact capabilities, which is important context for subclass implementors.

The `zero_mass` guard in the ppf-based approximation (`np.abs(cdf_u - cdf_l) < 1e-15`) correctly handles degenerate intervals, but the same guard is absent in the `_energy_x` shortcut path — if a point `x` lies exactly at a probability-0 region boundary, `right_mean` or `left_mean` could return `nan` and the `nan_to_num(..., nan=0.0)` silently masks what might be a meaningful divergence rather than a true zero-contribution interval.

fkiraly (changes requested):
Very nice!

Can you please add a test to `TestAllDistributions` that checks that the method works? I think the method just has to be added to the right places.

skpro/distributions/base/_base.py
```diff
@@ -1397,6 +1429,98 @@ def _mean(self):
...
+            f"{approx_spl_size} samples filtered to the truncation bounds"
+        )
+        warn(self._method_error_msg("truncated_mean", fill_in=approx_method))
+
+        spl = [self.sample(1) for _ in range(approx_spl_size)]
+        spl_arr = np.array([s.values if hasattr(s, "values") else s for s in spl])
+        if spl_arr.ndim == 1:
+            spl_arr = spl_arr.reshape(-1, 1, 1)
+        elif spl_arr.ndim == 2:
+            spl_arr = spl_arr.reshape(spl_arr.shape[0], 1, -1)
```
fkiraly: why is this correct? or is it a mistake?

patelchaitany: Yeah, those reshape branches are dead code on my end. sample(1) always returns a DataFrame, so .values gives 2D, stacking gives 3D neither ndim == 1 nor ndim == 2 ever fires. I added them as defensive guards but they're unnecessary. Happy to remove them.

patelchaitany:
@fkiraly, I have added truncated_mean to METHODS_SCALAR plus two new tests one checks that unbounded truncated_mean() matches mean(), the other verifies output format with actual bounds.

fkiraly (changes requested):
Reviewed in more detail - this is very very nice!

I think we could in-principle merge, but I would like to discuss two things:

* have you considered a design where, instead of adding a new public API point `truncated_mean`, you instead add two arguments to `mean`? Where, of course, internally, it could still link to `_mean` and `_truncated_mean` (or `_mean_truncated`), similar to how `energy` dispatches to two different private methods. I am not saying this is how we should do it, but I would like to see your thoughts on a discussion of pros/cons.
* can you quickly give me a summary where your analytical formulae are coming from for the various distributions? I agree with the quadrature in the base class, but am asking for the various distributions. Did you ask AI; Wolfram; or did you do this on paper? In either case, it may be worth documenting the exact formula somewhere in an md file, so we can add it to the docstring once the docstring injection PR is resolved.

patelchaitany:
It was mix of every thing AI, hand notes and Wikipedia.

patelchaitany:
@fkiraly where should i add this formula.md do i create seprate doc and share it or i create it on project it self on some folder?

patelchaitany:
Merged truncated_mean into mean(lower=None, upper=None) when both are None it routes to _mean, otherwise resolves to ±inf and routes to _truncated_mean. Same pattern as energy. No changes to private methods, tags, or subclass implementations

patelchaitany:
@fkiraly Is there any changes required?

felipeangelimvieira (approved):

PR #1010, opened 2026-03-30
[BUG] - bugfix, GridSearchCV scoring=None defaults to CRPS

MaazAli8460:
If scoring= None, set it to CRPS().

#### Reference Issues/PRs

Fixes #1009.

#### What does this implement/fix? Explain your changes.
If scoring= None, set it to CRPS(). in _tuning.py

#### Does your contribution introduce a new dependency? If yes, which one?
No.

#### Did you add any tests for the change?
No. I don't think a separate test is necessary here.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
  How to: add yourself to the [all-contributors file](https://github.com/sktime/skpro/blob/main/.all-contributorsrc) in the `skpro` root directory (not the `CONTRIBUTORS.md`). Common badges: `code` - fixing a bug, or adding code logic. `doc` - writing or improving documentation or docstrings. `bug` - reporting or diagnosing a bug (get this plus `code` if you also fixed the bug in the PR).`maintenance` - CI, test framework, release.
  See here for [full badge reference](https://allcontributors.org/docs/en/emoji-key)
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
- [ ] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [ ] I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant `Examples` section.
- [ ] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured
  dependency isolation, see the [estimator dependencies guide](https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

PR #1014, opened 2026-03-30
fix: handle scoring=None in GridSearchCV._fit to avoid AttributeError (#1009)

ekkoitac:
## Fix Issue #1009: GridSearchCV scoring=None gives AttributeError

**Problem:** When `scoring=None` is passed to `GridSearchCV.fit()`, line 111 crashes with `AttributeError: 'NoneType' object has no attribute 'name'` because `scoring.name` is called without checking if `scoring` is `None`.

**Fix:** Check if `scoring is not None` before accessing `scoring.name`. When `None`, use `"test_CPRS"` as the scoring name, matching the documented behavior that `scoring=None` defaults to CRPS.

```python
# Before
scoring_name = f"test_{scoring.name}"  # crashes when scoring is None

# After
scoring_name = f"test_{scoring.name}" if scoring is not None else "test_CPRS"  # safe fallback
```

### DIFF
--- a/skpro/model_selection/_tuning.py
+++ b/skpro/model_selection/_tuning.py
@@ -108,7 +108,7 @@ class BaseGridSearch(_DelegatedProbaRegressor):
 
         # scoring = check_scoring(self.scoring, obj=self)
         scoring = self.scoring
-        scoring_name = f"test_{scoring.name}"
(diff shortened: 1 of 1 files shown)


direkkakkar319-ops:
Hi @ekkoitac this pr looks similar to PR #1010

PR #917, opened 2026-03-14
[MNT] Refactor CyclicBoosting to eliminate feature validation code duplication

MayankSharma-2812:
Reference Issues/PRs
Fixes #916

What does this implement/fix? Explain your changes.
This refactoring eliminates code duplication in CyclicBoosting by extracting the repeated feature validation logic into a single helper method _validate_feature_groups(). The change:

Adds helper method _validate_feature_groups(X) that validates all feature groups exist in X columns
Replaces duplicated code in 5 methods with single helper call:
_fit()
_predict()
_predict_proba()
_predict_interval()
_predict_quantiles()
Maintains identical behavior - same error messages and validation logic
Adds a test to ensure missing feature groups raise the expected error.
Does your contribution introduce a new dependency? If yes, which one?
No new dependencies introduced.

What should a reviewer concentrate their feedback on?
Verify that the helper method logic is identical to the original duplicated code
Ensure all 5 methods correctly call the helper method
Check that error messages remain unchanged
Confirm the new test provides adequate coverage
Did you add any tests for the change?
Yes. Added test_cyclic_boosting_missing_feature_validation() to skpro/regression/tests/test_cyclic_boosting.py which:

Tests error messages for missing features
Validates tuple feature groups work correctly
Ensures the helper method is called consistently across all prediction methods
Any other comments?
This is a pure maintenance refactoring that improves code maintainability without changing any external behavior. The fix follows DRY principles and reduces future maintenance burden.

PR checklist
For all contributions
I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
The PR title starts with [MNT] - maintenance refactoring

PR #1038, opened 2026-04-15
Fix/issue 1037 [MNT]

### LINKED ISSUE #1037: [MNT] Refactor duplicated feature validation in CyclicBoosting and fix broken error message.
There’s a small bit of duplicated logic around feature validation in `CyclicBoosting` that currently appears in five places:

* `_fit`
* `_predict`
* `_predict_proba`
* `_predict_interval`
* `_predict_quantiles`

It may be worth pulling this into a shared private helper so the validation only needs to be maintained in one place.

Also, there seems to be an issue with the current error handling:

```python id="ph6l5d"
if not set(feature_names).issubset(set(X.columns)):
    raise ValueError(f"{feature} is not in X")
```

`feature` is not defined here, so instead of raising the intended validation error, this can fail with a different exception.

Something along these lines may be a bit more robust:

```python id="uhxwqy"
missing_features = set(feature_names) - set(X.columns)

if missing_features:
    raise ValueError(
        f"Missing required feature(s) in X: {sorted(missing_features)}"
    )
```

That would make it easier to understand exactly which columns are missing and avoid repeating the same validation logic across multiple methods.

krsatyamthakur-droid:
#### Reference Issues/PRs
Fixes #1037.

There was a bit of duplicated feature validation logic in `CyclicBoosting` that was copied across five different methods (`_fit`, `_predict`, `_predict_proba`, `_predict_interval`, and `_predict_quantiles`). This PR simply pulls that block into a shared private `_check_features` helper method to keep things DRY and easier to maintain. 

It also fixes a minor issue with the error handling in that block. Previously, it raised a `ValueError` using a loosely defined `feature` variable, which actually caused an unrelated exception to be thrown instead of the intended error. This was updated to directly compute the set difference of the missing columns and display exactly which features are missing in the error message.

#Does your contribution introduce a new dependency? If yes, which one?
No.

# What should a reviewer concentrate their feedback on?
Just a quick sanity check to ensure the new helper `_check_features` is correctly placed and replacing the right blocks in the respective methods. 

# Did you add any tests for the change?
No new tests were added since this is purely a refactor of existing private logic and a patch for an error message. Existing CI suite should be sufficient.

#Any other comments?
None. 

##PR checklist

## For all contributions

##For new estimators
- [x] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured dependency isolation.

### DIFF
--- a/skpro/regression/bootstrap.py
+++ b/skpro/regression/bootstrap.py
@@ -11,6 +11,7 @@ from sklearn.utils import check_random_state
 from skpro.distributions.empirical import Empirical
 from skpro.regression.base import BaseProbaRegressor
 from skpro.utils.numpy import flatten_to_1D_if_colvector
+from skpro.utils.sampling import _random_ss_ix
 from skpro.utils.sklearn import prep_skl_df
 
 
@@ -203,6 +204,3 @@ class BootstrapRegressor(BaseProbaRegressor):
         }
 
         return [params1, params2]
-
-
-from skpro.utils.sampling import _random_ss_ix  # noqa: E402, F811
--- a/skpro/regression/bootstrap.py
+++ b/skpro/regression/bootstrap.py
@@ -205,11 +205,4 @@ class BootstrapRegressor(BaseProbaRegressor):
         return [params1, params2]
 
 
-def _random_ss_ix(ix, size, replace=True, random_state=None):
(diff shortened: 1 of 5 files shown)


krsatyamthakur-droid:
Closing this PR as the duplicated logic and error message were already refactored and fixed by PR #917.

PR #1041, opened 2026-04-23
[MNT] Fix #1039: Eliminate global seeds and replace sklearn.utils.check_random_state

### LINKED ISSUE #1039: [MNT] Eliminate global np.random.seed calls to prevent global state mutation.
I found some `np.random.seed()` calls in a few regressors that are currently mutating the global NumPy state. This is generally bad practice for library code because it can cause sneaky side effects in a user's environment and makes it hard to get reproducible results when running things in parallel.

We should clean these up and switch to proper local random state management using `check_random_state`.

**Impacted files:**
- `skpro/regression/bootstrap.py`: Line 113
- `skpro/regression/enbpi.py`: Line 163
- `skpro/regression/ensemble/_bagging.py`: Line 133
- `skpro/regression/mdn.py`: Line 441

**Proposed fix:**
- Remove the `np.random.seed(...)` calls.
- Use `sklearn.utils.check_random_state` to handle the `random_state` parameter locally.
- For `mdn.py` and `_bagging.py`, we'll need to add the `check_random_state` initialization to store a local RNG.
- Ensure any sampling or noise logic uses that local RNG instead of the global `np.random` module.

`bootstrap.py` and `enbpi.py` already have a `self._random_state` object initialized, so there we just need to delete the global seed line.

krsatyamthakur-droid:
#### Reference Issues/PRs
Fixes #1039.

Cleaned up global `np.random.seed()` calls in several regressors to stop them from mutating the global state. 

Following the maintainer's suggestion, I also removed the `scikit-learn` dependency for random state management by adding a pure NumPy `check_random_state` helper in `skpro/utils/random_state.py`.

**Changes:**
- Added an internal `check_random_state` utility in `random_state.py` so we don't have to pull from `sklearn.utils`.
- Removed `np.random.seed()` from `bootstrap.py`, `enbpi.py`, `_bagging.py`, and `mdn.py`.
- Updated those estimators to use a local `self._random_state` for all sampling and noise logic.
- Switched imports to use our new internal utility.

No. This actually helps remove the `sklearn` dependency for these modules.

Check the `check_random_state` implementation in `skpro/utils/random_state.py` to make sure it covers all the edge cases for NumPy seeds. Also, double-check that I didn't miss any sampling steps in the regressors that still use the global `np.random`.

I ran the existing tests to make sure I didn't break reproducibility:
- `pytest skpro/regression/tests/test_mdn.py` (Passed)
- `pytest skpro/regression/tests/test_all_regressors.py` (63 tests for the affected regressors all passed).

#### Any other comments?
This should make these regressors thread-safe and stop them from messing with the user's local environment.

#### PR checklist
- [x] I've added myself to the list of contributors.
- [x] The PR title starts with [MNT].


krsatyamthakur-droid:
Hii @fkiraly "I've refactored this to avoid the scikit-learn dependency as you suggested. I added a local NumPy version of check_random_state so we can handle the seeds internally. Should be good now!"

PR #1042, opened 2026-04-25
[ENH] MDN Improved Losses (nGem)

joshdunnlime:
#### Reference Issues/PRs

No Issue but referenced [here](https://github.com/sktime/skpro/pull/796#issuecomment-4027898053).

Mixture density networks typically uses the negative log-likelihood (nll) objective, which can suffer from slow convergence and mode collapse. The natural gradient expectation maximization (ngem) objective can achieve up to 10x faster convergence while adding almost zero computational overhead, and scales well to high-dimensional data where nll fails.

[Learning Mixture Density via Natural Gradient Expectation Maximization](https://arxiv.org/html/2602.10602v1)

No.

~General API structure for adding ngem specific learning rate. ngem typically performs better with lower lr, so adding an automated lr multiplier has been considered. Is this designed correctly?~

I have removed `ngem_lr_scaling` as it obscures true `lr` and sets a bad standard if we wish to add more objectives that also benefit different magnitudes of scaling.

Existing MDN tests to still pass. Add specific ngem tests also.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Nice! Though, should the log-loss - previous behaviour - not also be an option?

I think we should not break user behaviour, so I would suggest to retain previous usage of the log-loss as an option.

Or does it not get removed? Would appreciate explanation if I misunderstood.

joshdunnlime:
No we don't remove log-loss (or rather [nll](https://github.com/sktime/skpro/pull/1042/changes#diff-289ef28ac2f44b1654b2f73e2d22c7a590e8674d5407d1a80adb3b38d82964c5R154)). `ngem` is provided as an additional option to the loss argument.

fkiraly (approved):
Ah, ok. You are just returning unnormalized logits from the network now, right?

fkiraly (changes requested):
Can you kindly look into the failure message? Looks related to the changes. Have you run `check_estimator`?

joshdunnlime:
> Can you kindly look into the failure message? Looks related to the changes. Have you run `check_estimator`?

Exactly. We just pass those to the loss the users selects.

I'll fix the test ASAP. Thanks!

fkiraly:
kind reminder

joshdunnlime:
@fkiraly - this should be fixed now. It passes locally.

fkiraly (approved):
Looks good now!

PR #1045, opened 2026-05-03
[ENH] add _normalize_dist_str utility for consistent distribution string aliases

paramsureliya:
Part of #1023

#### Reference Issues/PRs

Part of #1023 first step toward uniformizing distribution string handling across all probabilistic regressors.

Adds a shared `_normalize_dist_str` utility (`skpro/regression/_dist_utils.py`) that maps case-insensitive distribution string aliases to the canonical capitalized class name used internally in skpro, e.g. `"gaussian"` → `"Normal"`, `"lognormal"` → `"LogNormal"`, `"t"` → `"TDistribution"`.

Integrates this into `NGBoostAdapter` (inherited by both `NGBoostRegressor` and `NGBoostSurvival`) as a proof of concept. Normalization happens at point-of-use inside the adapter methods, not in `__init__`, so `get_params()`/`clone()` compatibility is preserved.

The remaining regressors from #1023 (XGBoostLSS, ResidualDouble, CyclicBoosting, GAMRegressor, GLMRegressor, GlumRegressor) will be addressed in follow-up PRs.

No new dependencies.

- The alias map in `_dist_utils.py`  any missing or incorrectly mapped aliases?
- The integration pattern in `NGBoostAdapter`  normalization into a local `dist` variable at point-of-use, leaving `self.dist` unchanged for sklearn compatibility.

Yes `skpro/regression/tests/test_dist_utils.py` with 57 tests across 3 classes:
- `TestNormalizeDistStr`: unit tests for every alias, canonical passthrough, non-string passthrough, unknown string warning, and idempotency
- `TestCrossRegressorAliasConsistency`: invariant tests ensuring the same alias resolves identically regardless of which regressor calls the function
- `TestNGBoostRegressorAliases`: end-to-end integration tests for NGBoostRegressor (auto-skipped if ngboost not installed)

#### Any other comments?

None.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -264,6 +264,12 @@
         "bug",
         "doc"
       ]
+    },
+    {
+      "login": "paramsureliya",
+      "name": "Param Sureliya",
+      "profile": "https://github.com/paramsureliya",
+      "contributions": ["code"]
     }
   ]
 }
--- a/skpro/regression/_dist_utils.py
+++ b/skpro/regression/_dist_utils.py
@@ -1,0 +1,130 @@
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+"""Distribution string normalisation utility for skpro regressors.
+
+Provides ``_normalize_dist_str``, which maps all known string aliases for a
+probability distribution to the canonical capitalized class name used in skpro
+(e.g. ``"gaussian"`` -> ``"Normal"``, ``"t"`` -> ``"TDistribution"``).
+
+Every probabilistic regressor adapter should call this before its own internal
+string -> object mapping so that users can pass any reasonable alias and have it
+work uniformly across regressors and in GridSearchCV / RandomizedSearchCV.
+"""
+
+_DIST_ALIAS_MAP: dict[str, str] = {
+    # Normal / Gaussian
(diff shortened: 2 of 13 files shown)


fkiraly (changes requested):
This is great!

If it is already done though, we should apply it to all places where distribution (string) aliasing happens, consistently.

Further, I would only advertise the "canonical" string in the docstrings, in order to uniformize future usage. As "canonical" strings, I would use the class names as in `skpro`. Alternative aliases for downwards compatibility can still function but be unadvertised.

paramsureliya:
Hi @fkiraly I've applied _normalize_dist_str to all remaining regressors. can you please review it.

fkiraly (changes requested):
Thanks!

Question, why are you reformatting or changing docstrings that looked like they aere good? E.g.,  why are you changing `NGBoostRegressor`, and why are you introducing newlines in places unrelated? Looks like AI use.

paramsureliya:
Hi @fkiraly fixed the isort failure.
Regarding the NGBoostRegressor docstring change: I updated it based on your earlier feedback to "only advertise the canonical string in the docstrings." Happy to revert it if you prefer to keep the original format there just let me know.

Issue #1050, opened 2026-05-17
[ENH] `update` capability for `GridSearchCV` and `RandomizedSearchCV`


fkiraly:
Online learning capability should be added to tuners `GridSearchCV` and `RandomizedSearchCV`.

For this, different algorithms can be used - similar to `sktime`, an `update_behaviour` parameter should be introduced - see there for a description of algorithms.

fkiraly:
FYI @patelchaitany

patelchaitany:
@fkiraly Does this approach look good? Happy to adjust if you have different preferences.

From looking at the codebase, BaseGridSearch currently has no capability:update tag and doesn't clone it from the inner estimator, so .update() is always a no-op.

My plan:

1. Clone capability:update from the inner estimator so the tuner automatically supports update when the wrapped estimator does.
2. Add an update_behaviour parameter (following sktime's pattern) that controls how updates are applied:
   - "update_best" -- update only best_estimator_
   - "update_all" -- update all candidates in n_best_estimators_
   - "refit_all" -- accumulate data and re-run the full grid search

This applies to both GridSearchCV and RandomizedSearchCV since they share BaseGridSearch.

fkiraly:
Yes, that is a sensible approach. Perhaps you can also make sure to use the same strategy names as in `sktime`, for consistency?

patelchaitany:
@fkiraly, should this issue be closed or there is more to this.

fkiraly:
I think one of the options should be updating only the best estimator, or the n best estimators and then re-evaluating which one is best. For that, we need the `river` adapter, i.e., #1068

PR #1053, opened 2026-05-21
[ENH] added `update` capability in `GridSearchCV` and `RandomizedSearchCV`

patelchaitany:
#### Reference Issues/PRs

#1050 

This add the update capability for GridSearchCV and RandomizedSearchCV

BaseGridSearch:

- Added update_behaviour="full_refit" parameter to __init__
- Added "capability:update" to tags_to_clone so the tag propagates from the inner estimator
- Added self._X, self._y, self._C storage at end of _fit() for data accumulation
- Added _update() method with three strategies: "full_refit", "inner_only", "no_update"
- Added _update_data() helper for concatenating old + new data

GridSearchCV: 
- Added update_behaviour parameter to __init__, passed through to super 
- Added update_behaviour docstring entry 
- Added test params with OnlineRefit-wrapped estimator

RandomizedSearchCV: 
- Same three changes as GridSearchCV (parameter, docstring, test params)

No

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/model_selection/_tuning.py
+++ b/skpro/model_selection/_tuning.py
@@ -14,10 +14,40 @@ from skpro.utils.parallel import parallelize
 
 
 class BaseGridSearch(_DelegatedProbaRegressor):
+    """Base class for hyperparameter search with cross-validation.
+
+    Parameters
+    ----------
+    estimator : estimator object
+        Probabilistic regressor to tune.
+    cv : cross-validation generator or an iterable
+        Cross-validation splitting strategy.
+    backend : str, optional (default="loky")
+        Parallel backend for candidate evaluation.
+    refit : bool, optional (default=True)
+        Whether to refit the best estimator on all data.
+    scoring : skpro metric, str, or callable, optional (default=None)
+        Metric used to score candidate parameter settings.
+    verbose : int, optional (default=0)
+        Verbosity level.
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks, good start!

Some change requests:

* please do *not* store the data by default, this causes memory leaks in current code. Instead, make the default option the current one - no update happens - and only store data if a strategy is chosen that requires this.
* No pull request is ready without docstrings. Please update the docstrings properly.

patelchaitany:
Thanks for the review - both points are addressed

fkiraly (changes requested):
Nice!

I think the tag `capability:update` is not correctly set though?

If `full_refit`, it should be True; if `no_update`, it should be False, If `inner_only`, it should be clone the tag of the inner estimator.

Currently, it is always cloned.

The class tag (in `_tags`) should be True, since for some setting of parameters the capability is True.

patelchaitany:
Thanks @fkiraly for catching this. Sorry I overlooked the capability:update tagging in the first pass.

I have pushed a fix that sets the class tag to True in _tags and sets the instance tag from update_behaviour: False for no_update, True for full_refit, and clones from the inner estimator for inner_only.

Happy to adjust if anything else should change.

fkiraly (approved):
Nice!

Issue #1049, opened 2026-05-17
[ENH] online versions of Jackknife+-after-bootstrap and bagging regressors


fkiraly:
The following two closely related regressors lend themselves to natural online (`update`) extensions:

* `BaggingRegressor` (bagging ensemble)
* `EnbpiRegressor` (Jackknife+-after-bootstrap)

These can be extended naturally to update-capable regressors, if the passed `estimator` is a `river` online estimator.

FYI @patelchaitany

Both estimators would be extended as follows:

Upon receiving new data in `update`, each inner `estimator` is `update`-d on a subsample of the new subsample that extends the statistical properties of the originally fitted subsample.

Column sampling is not repeated, the same columns are used as previously.

felipeangelimvieira:
@fkiraly are river estimators currently supported in skpro? 
We would need a dedicated river wrapper in skpro, right?

fkiraly:
> @fkiraly are river estimators currently supported in skpro?

No, they are not - this would be the first such example.

patelchaitany:
Hey, love to pick up this issue

patelchaitany:
@fkiraly I've been thinking about how we can add the `update` functionality in the online regressor, so this is what I have researched and what I have come up with.

### Current state

**`BaggingRegressor`** (`skpro/regression/ensemble/_bagging.py`):

1. On `fit`, creates `n_estimators` clones and fits each on a random row subsample (+ optional feature subsample).
2. Stores `estimators_` and `cols_` (per bag feature subsets for `predict_proba`).
3. Does **not** store training data, per bag row indices, or bootstrap membership.
4. Inherits `capability:update=False` from `BaseProbaRegressor`, so public `update()` is currently a no op.

**Clones and data:** Fitted clones only hold model state, not a copy of `(X, y)` unless the inner estimator does so internally. After `fit`, the ensemble cannot reconstruct each bag's training set. **Data responsibility today is entirely on the caller at fit time.**

### Intended semantics

The bagging meta estimator can only **route** new data to each clone. Whether that counts as continued learning depends entirely on the **inner regressor**.

**Bootstrap subsampling on the new batch** (same `n_samples_`, `bootstrap` as in `fit`, fixed `cols_[i]` from `fit`) is what `BaggingRegressor` would do on each `update` call:

1. Draw a row subsample from the **incoming** `(X, y)` only (not from stored fit data, unless a separate strategy chooses to store and refit).
[...]

PR #1064, opened 2026-06-03
[ENH] add online update to BaggingRegressor

patelchaitany:
#### Reference Issues/PRs

Partially addresses #1049.

Adds online `update` support to `BaggingRegressor`.

On `update`, each fitted bagged clone is updated on a row subsample of the incoming batch, using the same `n_samples` and `bootstrap` settings as in `fit`. Feature subsets `cols_[i]` from `fit` are reused (column sampling is not repeated). Sets `capability:update=True` on the meta-estimator so the public `update` path runs; meaningful incremental learning still depends on the inner regressor (batch-only inners effectively no-op). When `bootstrap=False`, subsample size is capped at the update batch size so small batches do not error.

Docstring updated to describe `update` behaviour.

No.

- Whether `_update` row subsampling matches `fit` semantics (including `bootstrap=False` on small update batches).

No dedicated tests added. Covered by existing `test_online_update` in `test_all_regressors.py` for `BaggingRegressor` via `get_test_params()`.

#### Any other comments?

Follow-up for #1049 may include `EnbpiRegressor` and a River wrapper for online inner estimators.

##### For all contributions

##### For new estimators

### DIFF
--- a/skpro/regression/ensemble/_bagging.py
+++ b/skpro/regression/ensemble/_bagging.py
@@ -21,6 +21,15 @@ class BaggingRegressor(BaseProbaRegressor):
 
     On ``predict_proba``, the mixture of probabilistic predictions is returned.
 
+    In ``update``, each fitted clone is updated on a row subsample of the new
+    batch. The row subsample fraction is the same as in ``fit`` (for integer
+    ``n_samples``, the fraction ``n_samples / n_fit`` is stored at ``fit`` and
+    applied to each update batch). The same ``bootstrap`` setting and feature
+    subset ``cols_[i]`` as at ``fit`` are reused (column sampling is not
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks!

For the algorithm to make sense, in the case where `n_samples` is an integer, a fraction `n_samples / n` should be computed in `_fit`, and applied to the sample in `_update`.

That is, the row selection probability should remain the same for `_fit` and `_update`.

With this insight, it should also be possible to merge the two "resolve size" methods into a single one.

patelchaitany:
Hey @fkiraly, I've addressed your review comments! Whenever you get some time, please take another look and let me know if it's good to go. Thanks!

fkiraly (approved):
Great!

PR #1065, opened 2026-06-08
[ENH] reject non-positive sigma/scale in Normal and Laplace #959

### LINKED ISSUE #959: [BUG]Normal and Laplace accept non-positive sigma/scale and return invalid pdf values
**Describe the bug**
Normal and Laplace accept non-positive values for their required scale parameters and can return mathematically invalid pdf values instead of rejecting the input.

For example:

- `Normal(mu=0, sigma=-1).pdf(0)` returns `-0.3989422804014327`

- `Laplace(mu=0, scale=-1).pdf(0)` returns `-0.5`

A probability density function should never be negative, so this is a silent correctness issue.

**To Reproduce**
```python
from skpro.distributions.laplace import Laplace

print(Normal(mu=0, sigma=-1).pdf(0))
print(Laplace(mu=0, scale=-1).pdf(0))
```

```python
Observed output in skpro:

-0.3989422804014327
-0.5
```
```python
SciPy comparison:

from scipy.stats import norm, laplace

print(norm.pdf(0, loc=0, scale=-1))
print(laplace.pdf(0, loc=0, scale=-1))
```
```python
SciPy returns:

nan
nan
```

**Expected behavior**
These distributions document `sigma` and `scale` as positive parameters. Passing non-positive values should raise a clear `ValueError`, ideally at construction time, instead of allowing invalid parameters through to pdf evaluation.

**Environment**
OS: Windows
Python: 3.9.13
skpro: current main branch (local clone)

**Additional context**
This appears to be missing parameter validation in the distribution constructors. I checked a few related distributions as well; the strongest confirmed cases from direct reproduction are `Normal` and `Laplace`.

I'd be happy to open a PR for this.

shubhumre777:
#### Reference Issues/PRs
Fixes #959

This PR adds input parameter validation to the constructors (`__init__`) of both the `Normal` and `Laplace` distributions to enforce strictly positive scale parameters (`sigma` for Normal and `scale` for Laplace). 

Previously, passing non-positive values allowed mathematically invalid negative values to propagate to the `_pdf` calculation. The input validation uses a robust vectorized numpy check `np.any(np.asarray(...) <= 0)` to seamlessly handle both scalar values and multi-dimensional array structures without crashing.

No.

The vector/array validation logic implementation using `np.any(np.asarray(...) <= 0)` inside the constructors.

Yes, added a new automated unit test file:
`skpro/distributions/tests/test_parameter_validation.py` which verifies that both distributions throw a descriptive `ValueError` on negative and zero boundary values.

#### Any other comments?
This is my first contribution to skpro ,  Excited to hear feedback.

##### For all contributions

### DIFF
--- a/skpro/distributions/laplace.py
+++ b/skpro/distributions/laplace.py
@@ -56,6 +56,8 @@ class Laplace(BaseDistribution):
 
     def __init__(self, mu, scale, index=None, columns=None):
         self.mu = mu
+        if np.any(np.asarray(scale) <=0 ) :
+            raise ValueError("scale must be strictly positive .")
         self.scale = scale
 
         super().__init__(index=index, columns=columns)
--- a/skpro/distributions/normal.py
+++ b/skpro/distributions/normal.py
@@ -53,6 +53,8 @@ class Normal(BaseDistribution):
 
     def __init__(self, mu, sigma, index=None, columns=None):
         self.mu = mu
+        if np.any(np.asarray(sigma) <= 0) :
+            raise ValueError("sigma must be strictly positive.")
         self.sigma = sigma
 
         super().__init__(index=index, columns=columns)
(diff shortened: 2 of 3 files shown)


fkiraly:
Does not make sense as individual fix, and also does not comply with code formatting.

Issue #1066, opened 2026-06-08
[ENH] RiverRegressor adapter and online regressor support


patelchaitany:
**Is your feature request related to a problem? Please describe.**

skpro does not yet support online regressors that can update but only do point prediction, not full probabilistic output.

**Describe the solution you'd like**

Add an internal RiverRegressor adapter for [River](https://riverml.xyz/) models. It should support learn_one and learn_many, do point prediction only, and wrap River models automatically when passed to update-capable meta-estimators.

**Describe alternatives you've considered**

Users wrapping River models themselves, or making the adapter probabilistic - not preferred for now.

**Additional context**

Full design decisions are in the docs: [Design Doc](https://hackmd.io/@CZftYYlgR5uqgCJ29VNDOQ/BJqGrUx-Gl)

PR #1068, opened 2026-06-10
[ENH] Add online regressor support and River adapter

patelchaitany:
#### Reference Issues/PRs

#1066 

This PR adds the foundation for online regression in skpro, including a new regressor_online scitype and BaseOnlineRegressor with capability:update and capability:pred_int. It introduces the capability:pred_int tag and gates probabilistic prediction methods on regressors that do not support them. It also adds RiverRegressor as a point-prediction adapter for River models, with deepcopy-based cloning and coerce_to_skpro_regressor() for wrapping raw River estimators.

No
##### For all contributions

##### For new estimators
- [X] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [X] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured


fkiraly (changes requested):
Nice! Though I think there is somet duplication going on.

* I think the new base class, `BaseOnlineRegressor`, can be merged into `BaseProbaRegressor`, there is no substantial added functionality.
* The new jackknife-plus regressor can be merged into the existing one (the non-MAPIE regressor), simply add the `update` logic.

skpro/regression/enbpi.py
```diff
@@ -165,17 +172,60 @@ def _fit(self, X, y):
...
+        X : pandas DataFrame
+            feature instances to update with
+        y : pandas DataFrame, must be same length as X
+            labels to update with
+
+        Returns
+        -------
+        self : reference to self
+        """
+        errs, bs_vs_ix_new, estimators = self._bootstrap_batch(X, y, self.estimators_)
```
felipeangelimvieira: As I understand, `update` is behaving just like a brand new fit on the new batch. Shouldn't `_boostrap_batch` call `_update` on the new estimators instead of `fit`, if the user hass called EnbpiRegressor.update?

patelchaitany: I think I made mistake The original EnbpiRegressor dose not update the trained model it only update the residuals

skpro/registry/_base_classes.py
```diff
@@ -132,6 +132,28 @@ def get_test_class(cls):
         return TestAllRegressors
 
 
+class regressor_online(_BaseScitypeOfObject):
```
felipeangelimvieira: Now that there is no BaseOnlineRegressor, we can remove this object type.

skpro/registry/_tags.py
```diff
@@ -269,18 +269,30 @@ class capability__update(_BaseTag):
...
     _tags = {
         "tag_name": "capability:update",
-        "parent_type": "regressor_proba",
+        "parent_type": ["regressor_proba", "regressor_online"],
         "tag_type": "bool",
         "short_descr": "whether estimator supports online updates via update",
     }
 
 
+class capability__pred_int(_BaseTag):
```
felipeangelimvieira, line 277: Could you explain the motivation for this new tag?

felipeangelimvieira (changes requested):

skpro/registry/_base_classes.py
patelchaitany: Yes we can remove that but for the river using we are using separate tag named regressor_online and in future if create separate test for online_regressor it would be easy to add test, wdyt.

skpro/registry/_tags.py
patelchaitany: Until now all skpro regressors supported probabilistic prediction, so this tag wasn't needed. The River adapter is the first point-prediction-only estimator, so capability:pred_int is needed to distinguish which regressors support predict_interval/predict_proba - matching sktime's convention.

skpro/registry/_base_classes.py
felipeangelimvieira: Our API won't be supporting online regression directly, right? Except for update. I believe it would better to follow YAGNI than add this tag at the moment. In addition, this would need to be a separate tag, not a object type value.

felipeangelimvieira (approved):

skpro/regression/adapters/river/_utils.py
```diff
@@ -0,0 +1,102 @@
...
+    ----------
+    obj : object
+        Object to check.
+
+    Returns
+    -------
+    bool
+        True if ``obj`` is a River estimator, False otherwise.
+    """
+    if obj is None:
```
fkiraly: I think you can preface this with a `_check_soft_dependencies` call, then you can assume that `river` is present. That is: if check_soft_dependencies("river"...) is False, return False. Anything that comes after can assume that `river` is present in the environment.

Why this works: if `river` is not present, the passed object cannot be a `river` instance.

skpro/regression/base/_base.py
```diff
@@ -294,6 +298,7 @@ def predict_proba(self, X):
         y : skpro BaseDistribution, same length as `X`
             labels predicted for `X`
         """
```
fkiraly, line 357: are we missing a `check_is_fitted` here? Unrelated to your PR but it is still missing and shoudl be fixed?

pyproject.toml
```diff
@@ -53,6 +53,7 @@ all_extras = [
     "matplotlib>=3.3.2",
     "polars<1.43.0",
     "pymc; python_version < '3.13'",
+    "river",
```
fkiraly: we should not add it here - this should be an estimator specific dependency.

fkiraly (changes requested):
Great, thanks!

Since this would now add "online-only" regressors to the zoo, we also need to check compatibility with:

* tuners, grid search, etc
* pipelines
* further compositions, if any

For instance, there also, the `capability:pred_int` tag needs to be set consistently with a potential online update regressor.

You could add this in this PR together with tests, or in a PR building on top of this.

PR #1069, opened 2026-06-11
[ENH] `VotingProbaRegressor` - heterogeneous ensemble compositor

Ashish-Kumar-Dash:
#### Reference Issues/PRs

  None 

  #### What does this implement/fix? Explain your changes.

  Adds `VotingProbaRegressor`, a heterogeneous ensemble that fits multiple
  probabilistic regressors on the same data and returns a `Mixture` distribution
  of their predictions.

  Probabilistic generalization of sklearn's `VotingRegressor` — instead of
  averaging point predictions, it combines full predictive distributions via
  weighted mixture.

  Key decisions for the basis of this PR:
  - Inherits from `BaseMetaEstimator, BaseProbaRegressor` — follows the
  `Pipeline` pattern for heterogeneous estimator lists
  (`named_object_parameters`, property-based `_estimators`)
  - Capability tags use AND logic across components
  - Delegates all distribution math to `Mixture`

  #### Does your contribution introduce a new dependency? If yes, which one?

  No.

  #### What should a reviewer concentrate their feedback on?

  - The AND logic for capability tag cloning (lines 80-84) — this is different
  from single-estimator `clone_tags` used elsewhere
  - Whether `_estimators` as a property (mirroring `Pipeline._steps`) is the
  preferred pattern here

  #### Did you add any tests for the change?

  The estimator is auto-discovered by the existing test framework via
  `get_test_params` (3 parameter sets). All pass:
  - `test_all_estimators` — 67 passed
  - `test_all_regressors` — 21 passed

  #### Any other comments?

  No.

  #### PR checklist

  ##### For all contributions
  - [ ] I've added myself to the list of contributors with any new badges I've
  earned :-)
  - [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG].

  ##### For new estimators
  - [x] I've added the estimator to the API reference - in
  `docs/source/api_reference/regression.rst`
  - [x] I've added one or more illustrative usage examples to the docstring, in
  a pydocstyle compliant `Examples` section.
  - [x] If the estimator relies on a soft dependency, I've set the
  `python_dependencies` tag and ensured dependency isolation. (N/A — no soft
  dependencies)

PR #1075, opened 2026-06-19
[ENH] `StackingProbaRegressor` - probabilistic stacking ensemble

Ashish-Kumar-Dash:
#### Reference Issues/PRs

  Follows #1069  (VotingProbaRegressor).

 Adds `StackingProbaRegressor`, a heterogeneous stacking ensemble for probabilistic regression. Generalizes sklearn's `StackingRegressor` to the distributional prediction setting.

 During `fit`, base regressors produce out-of-fold distributional predictions via cross-validation. Mean and variance are extracted as meta-features to train a final probabilistic regressor (meta-learner). During `predict_proba`, base predictions are transformed to meta-features and passed to the fitted meta-learner.

  Key design choices:
  - Same `BaseMetaEstimator` + `BaseProbaRegressor` pattern as `VotingProbaRegressor`
  - OOF loop follows `ResidualDouble`'s CV pattern
  - Meta-features: mean and variance from each base distribution
  - `passthrough` option to include original features
  - Default meta-learner: `ResidualDouble(LinearRegression())`

  No.

  - Meta-feature choice (mean + variance) — sufficient for location-scale families
  - Tag logic: missing = AND over base estimators, survival = OR

  Three test parameter sets via `get_test_params` covering:
  default final_estimator, custom final_estimator, passthrough + auto-naming.
  All 89 tests pass (67 generic + 21 regressor + 1 doctest).

  #### PR checklist

  ##### For all contributions
  - [x] I've added myself to the list of contributors
  - [x] The PR title starts with [ENH]

  ##### For new estimators
  - [x] I've added the estimator to the API reference
  - [x] I've added illustrative usage examples to the docstring

### DIFF
--- a/docs/source/api_reference/regression.rst
+++ b/docs/source/api_reference/regression.rst
@@ -171,6 +171,7 @@ Heterogeneous ensembles of probabilistic regressors
     :template: class.rst
 
     VotingProbaRegressor
+    StackingProbaRegressor
 
 
 Distributional boosting
--- a/skpro/regression/ensemble/__init__.py
+++ b/skpro/regression/ensemble/__init__.py
@@ -3,6 +3,12 @@
 
 from skpro.regression.ensemble._bagging import BaggingRegressor
 from skpro.regression.ensemble._ngboost import NGBoostRegressor
+from skpro.regression.ensemble._stacking import StackingProbaRegressor
 from skpro.regression.ensemble._voting import VotingProbaRegressor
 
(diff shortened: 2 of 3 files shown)


fkiraly (changes requested):
Great contribution!

I think there should be an additional parameter, `features`, that determines what the parameters should be. Maybe `"meanvar"` can be the default, as current - I would also allow `"params"`, which uses the `to_df` method of the distribution.

Later, this will also allow to use distribution-to-feature transformations.

fkiraly (approved):
Great, thanks!

PR #1092, opened 2026-07-01
ENH: Fork cyclic_boosting into skpro and update NumPy compatibility (Fixes #1090)

### LINKED ISSUE #1090: [ENH] fork `cyclic_boosting` into `skpro` and update `numpy` compatibility
It seems `cyclic_boosting` is stale or unmaintained, because it calls `np.product` which is long deprecated and now no longer present.

We should investigate what the status of `cyclic_boosting` is, and fork it into `skpro` if now unmaintained.

https://github.com/Blue-Yonder-OSS/cyclic-boosting

---
Update: the project looks dead. We should fork and fix it.

For now, I would clean fork it into `skpro.libs` (as in the `sktime.libs` pattern), and update the `numpy` imports to ensure compatibility.

There is not too much gained in taking on the package itself, it has approx 300 downloads per month (so no proper user base).

utsab345:
## Summary

Fixes #1090.

This PR vendors the `cyclic_boosting` library into `skpro.libs` and updates it for compatibility with recent NumPy versions. Since the upstream project appears to be unmaintained, this allows `skpro` to maintain compatibility independently while preserving the existing functionality.

## Changes

* Forked the `cyclic_boosting` package into `skpro.libs`.
* Updated deprecated NumPy API usage (e.g. replacing removed/deprecated functionality such as `np.product` where applicable).
* Updated imports and internal references to use the vendored implementation.
* Preserved the original behavior while ensuring compatibility with current NumPy releases.

## Why

The upstream `cyclic_boosting` project appears to be inactive and is no longer compatible with newer NumPy versions due to deprecated API usage. Vendoring the library into `skpro` enables continued maintenance, avoids dependency on an unmaintained package, and ensures compatibility with supported NumPy versions.

## Testing

* Verified that the vendored implementation imports successfully.
* Confirmed compatibility with current NumPy versions.
* Ran the relevant test suite to ensure existing functionality remains unchanged.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,7 @@ repos:
     hooks:
       - id: pydocstyle
         args: ["--config=setup.cfg"]
+        exclude: "^skpro/libs/"
 
   # We use the Python version instead of the original version which seems to require Docker
   # https://github.com/koalaman/shellcheck-precommit
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ ignore = E121, E123, E126, E226, E24, E704, W503, W504
 max-line-length = 88
 exclude =
     skpro/_contrib/*
+    skpro/libs/*
 extend-ignore =
     # See https://github.com/PyCQA/pycodestyle/issues/373
     E203
--- a/skpro/libs/__init__.py
+++ b/skpro/libs/__init__.py
@@ -1,0 +1,1 @@
+"""Libraries bundled with skpro for compatibility."""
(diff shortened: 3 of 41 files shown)


fkiraly (changes requested):
Thanks!

It would be appreciated if you could run pre-commit across the files to fix the issues.

utsab345:
Ran pre-commit across all 38 changed files. All hooks now pass:

- Added exclude patterns for vendored code (`skpro/libs/`) in `setup.cfg` (flake8) and `.pre-commit-config.yaml` (pydocstyle) since the forked cyclic_boosting library should stay close to upstream.
- Applied black/isort formatting to both vendored and integration files.
- All other hooks (pyupgrade, flake8, pydocstyle, etc.) pass clean.

PR #1093, opened 2026-07-01
[ENH] Add distfitter scitype to base class registry

### LINKED ISSUE #1084: [ENH] add distribution fitters to base class registry
Distribution fitters currently do not work with `check_estimator`, this is due to missing registration in `registry._base_classes`.

This should be added.

FYI @patelchaitany, I would think you could take care of this since you already worked on #897 ?

utsab345:
Distribution fitters (`NormalFitter`, `MOMFitter`) were missing from the `registry._base_classes` registration, causing `check_estimator` to fail for these estimators.

Fixes #1084.

Added the `distfitter` scitype with:
- `BaseDistFitter` as its base class
- `TestAllDistFitters` as its test class
- `parent_scitype`: `estimator`

Follows the existing pattern used by `regressor_proba` and other registered scitypes.

### DIFF
--- a/skpro/registry/_base_classes.py
+++ b/skpro/registry/_base_classes.py
@@ -133,6 +133,35 @@ class regressor_proba(_BaseScitypeOfObject):
 
 
 # ----------------------------------
+# Distribution fitters scitypes
+# ----------------------------------
+
+
+class distfitter(_BaseScitypeOfObject):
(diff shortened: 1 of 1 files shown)


patelchaitany:
LGTM, wdyt @fkiraly

fkiraly (approved):

Issue #1083, opened 2026-06-29
[ENH] update `skpro` base framework with `sktime` upgrades


fkiraly:
We should update `skpro` with the upgraded base API patterns from `sktime`. These can be added individually.

- [ ] tag registry and tag documentation, as per `sktime.registry._tags` and concomitant framework code
- [ ] `__post_init__` and `__dynamic_tags__` to collect non-parameter-init logic; handling dependency checks more leniently; adding these methods to extension templates

ishan372or:
Hi @fkiraly ,I'd like to work on the remaining tag registry tasks. While comparing `skpro` and `sktime`, I noticed that `registry.all_tags()` behaves differently when filtering by object. In `sktime`, filtered results include shared "object" and "estimator" tags, whereas `skpro` currently appears to filter only by the requested object type.
Would implementing `all_tags()` in skpro be the way to go?

fkiraly:
hm, that is a good question. I think `all_tags` could be a separate contribution?

ishan372or:
I was looking into it and it seems the difference is related to how `all_tags()` handles scitype inheritance. In `sktime` this is handled by including "estimator" and "object" when filtering, but `skpro` has a deeper scitype hierarchy which might not be satisfied by a direct port.
Would you consider this part of the registry work tracked by this issue, or would you prefer it to have its own issue/PR?

PR #1094, opened 2026-07-01
[ENH] Add __dynamic_tags__ and __post_init__ to base framework

utsab345:
Part of #1083 - update skpro base framework with sktime upgrades.

Adds two lifecycle hooks to `BaseObject` and `BaseEstimator`:
- **`__dynamic_tags__`**: for conditional tag setting based on parameters (e.g., cloning tags from component estimators), called after `super().__init__()`
- **`__post_init__`**: for parameter validation and initialization of derived quantities, called after `__dynamic_tags__`

Updates all four extension templates (`regression.py`, `distfitter.py`, `distributions.py`, `survival.py`) to document and include these methods, with examples of conditional tag setting and default estimator initialization.

This is the first incremental upgrade of skpro's base framework following sktime's pattern.

### DIFF
--- a/extension_templates/distfitter.py
+++ b/extension_templates/distfitter.py
@@ -82,7 +82,8 @@ class ClassName(BaseDistFitter):
     # todo: fill init
     # params should be written to self and never changed
     # super call must not be removed, change class name
-    # parameter checks can go after super call
+    # parameter checks and component initialization should go in __post_init__
+    # dynamic tag logic should go in __dynamic_tags__
     def __init__(self, paramname, paramname2="paramname2default"):
         # todo: write any hyper-parameters and components to self
         self.paramname = paramname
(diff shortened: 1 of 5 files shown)


fkiraly (changes requested):
Great!

For consistency, could you copy the docstrings from `sktime`?

utsab345:
Replying to @fkiraly's review: The `__post_init__` docstring has been updated across all files (BaseObject, BaseEstimator, and all 4 extension templates) to match the exact sktime pattern used in `BaseForecaster` and the forecasting/transformer extension templates. The `__dynamic_tags__` docstring was already identical between the two repos — the only inconsistency was in `__post_init__`. All 4718 tests pass.

fkiraly (changes requested):
Contributor claims to have copied docstrings from `sktime` but they are not consistent.

Likely uncritical use of AI which does not see the other respository - a human would easily spot that the docstrings are not identical..

utsab345:
@fkiraly I've pushed the fix updating the `__post_init__` docstring to match the exact sktime pattern. Please re-review when you get a chance.

extension_templates/distfitter.py
```diff
@@ -91,9 +92,36 @@ def __init__(self, paramname, paramname2="paramname2default"):
...
+    # todo: optional, for parameter validation and derived quantities
+    # if not needed, delete this method
+    def __post_init__(self):
+        """Post-init constructor logic, can be used by inheriting classes.
+
+        This method should be used for:
+
+        * parameter validation
+        * initialization logic beyond self.param = param
+        * dynamic tag setting
```
fkiraly: I think this is a mistake? Dynamic tag setting happens in `__dynamic_tags__`. Might be a problem in `sktime` though.

fkiraly: yes, it is ap roblem in sktime - I am fixing it in https://github.com/sktime/sktime/pull/10506. You should also fix it here.

utsab345:
@fkiraly Fixed - removed 'dynamic tag setting' from `__post_init__` docstrings across all files. Matches the fix in sktime/sktime#10506.

fkiraly (approved):
Thanks!

utsab345:
> Thanks!

You're welcome! Thanks for the review and guidance throughout the process.

PR #1095, opened 2026-07-01
[ENH] `EmpiricalFitter` distribution fitter

### LINKED ISSUE #1087: [ENH] `EmpiricalFitter` distribution fitter
We should add an `EmpiricalFitter` which converts the full sample into an empirical distribution.

For now, the univariate case suffices (empirical per variable).

This is particularly simple - it simply wraps the data in an `Empirical` distribution. It is useful when used as a component, as a naive form of distribution fitting.

utsab345:
Fixes #1087.

Adds `EmpiricalFitter` which converts the full sample into an `Empirical` distribution. For the univariate case (empirical per variable), it simply wraps the data in an `Empirical` distribution.

Supports single-column and multi-column data, with a `time_indep` parameter passed through to `Empirical`.

Useful as a base component for naive distribution fitting in ensemble or reduction strategies.


fkiraly (changes requested):
I think this is not correct, since the result should be an 0D distribution, but it is a 2D distribution.

fkiraly:
please ensure to fix code quality failures too. Please look at the developer guide.

utsab345:
Fixed as per your review: `_fit` now flattens multi-column data via `X.values.ravel()` into a single-column DataFrame so `proba()` always returns a 0D Empirical distribution. Also ran pre-commit hooks — black, isort, flake8, pydocstyle all passed.

skpro/distfitter/_empiricalfitter.py
```diff
@@ -0,0 +1,92 @@
...
+    >>> X = pd.DataFrame([1.0, 2.0, 3.0, 4.0, 5.0])
+
+    >>> fitter = EmpiricalFitter()
+    >>> fitter.fit(X)
+    EmpiricalFitter()
+    >>> dist = fitter.proba()
+    """
+
+    _tags = {
+        "authors": ["fkiraly"],
```
patelchaitany: I think you are the author for this class

skpro/distfitter/_empiricalfitter.py
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    def __init__(self, time_indep=True):
+        self.time_indep = time_indep
```
patelchaitany: There is no need for this as we are flattening the data so this time_indep is going to make no difference as there is going to be an single row

patelchaitany (changes requested):

utsab345:
Addressed review feedback:

- Fixed `_tags` authors to `utsab345`
- Removed `time_indep` parameter — since `_fit` flattens all data via `X.values.ravel()`, only a single column is passed to `Empirical`, making `time_indep` a no-op
- Removed custom `__init__` (inherited one suffices)
- Updated `\_proba()` and `get_test_params` accordingly
- All linting (black, isort, flake8, pydocstyle) and tests pass

patelchaitany (approved):

Issue #1086, opened 2026-06-29
[ENH] interface `distfit` as distribution fitter


fkiraly:
It would be great to have an interface to `distfit` as a distribution fitter.

Relevant extension template: `distfitter`

pyarchana:
hi @fkiraly, i wouldd like to work on this interfacing distfit as a distribution fitter.
i'll follow the distfitter extension template once the base framework lands. could you assign this to me?

fkiraly:
sure!

PR #1097, opened 2026-07-02
Add DistfitFitter distribution fitter wrapping the distfit package

pyarchana:
#### Reference Issues/PRs
Towards #1086.

Adds `DistfitFitter`, a `skpro` distribution fitter that wraps the [`distfit`](https://github.com/erdogant/distfit) package's parametric fitting mode: it fits a set of candidate `scipy.stats` distributions to the data, scores them, and returns the best-scoring one as a fitted `skpro` scalar distribution. It follows the existing `EmpiricalFitter`/`NormalFitter` pattern (`_fit`/`_proba`, `get_test_params`).

Rather than reimplementing pdf/cdf/mean/var per distribution, the returned distribution reuses skpro's existing `_ScipyAdapter` (`skpro/distributions/adapters/scipy/`) via a small private wrapper (`_DistfitDistribution`) that takes the scipy distribution name and shape/loc/scale that `distfit` produces.

Currently scoped to `distfit`'s `method="parametric"` mode only. I checked `distfit`'s other three modes (`quantile`, `percentile`, `discrete`) and they return a differently shaped model dict that this adapter doesn't handle yet. Happy to extend if that's wanted before merge, or leave it for a follow-up.

`distfit`, but it's already listed in `pyproject.toml`'s `all_extras`, so no dependency-list changes were needed.

- Whether scoping to `method="parametric"` is acceptable for a first version, or whether the other three `distfit` modes should be wired up before merge.
- The `_DistfitDistribution` adapter design (reuse of `_ScipyAdapter` vs. an alternative approach).
- Whether this should wait on #1093 / #1094. I verified it does not currently depend on either (both `BaseDistFitter` and the `distfitter` registry entry are already on `main`).

Yes: `skpro/distfitter/tests/test_distfitfitter.py`, plus it's automatically picked up by the existing generic `TestAllDistFitters` contract test suite.

#### Any other comments?
Opened as draft for early feedback on scope. Full local test suite, doctest, and lint (black/isort/flake8/pydocstyle) pass.

Note: `docs/source/api_reference/distfitter.rst` doesn't exist yet. The whole `distfitter` module (`EmpiricalFitter`, `NormalFitter`, `MOMFitter`) is currently undocumented there, not just this PR's addition. Flagging it here rather than silently leaving the checklist item unchecked. Happy to add the doc file for the whole module as part of this PR if that's preferred, or leave it for a separate cleanup PR. Let me know which you'd rather I do.

##### For all contributions
- [x] I've added myself to the list of contributors

##### For new estimators
- [ ] I've added the estimator to the API reference
- [x] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured dependency isolation.


.all-contributorsrc
```diff
@@ -271,6 +271,16 @@
       "contributions": [
         "bug"
       ]
+    },
+    {
+      "login": "areychana",
+      "name": "Archana K.",
+      "avatar_url": "https://avatars.githubusercontent.com/u/93970069?v=4",
+      "profile": "https://github.com/areychana",
```
patelchaitany: i think your github is wrong.

patelchaitany:
@pyarchana, add DistfitFitter under Parametric fitters in that RST file

skpro/distfitter/_distfitfitter.py
```diff
@@ -0,0 +1,141 @@
...
+            bins=self.bins,
+            random_state=self.random_state,
+            verbose="warning",
+        )
+        dfit.fit_transform(vals)
+
+        model = dfit.model
+        self.dist_name_ = model["name"]
+        self.shape_args_ = tuple(model["arg"])
+        self.dist_loc_ = float(model["loc"])
```
patelchaitany, line 109: this are the not documentted, Should't we document this?

.all-contributorsrc
pyarchana: i have recently changed the username, i will update it here too!

fkiraly (changes requested):
Thanks!

* the distribution returned should be existing distributions in `skpro`, not a new distribution without any clear features.
* for tests, use `run_test_for_class` instead of an explicit dependency check

pyarchana:
> Thanks!
> 
> * the distribution returned should be existing distributions in `skpro`, not a new distribution without any clear features.
> * for tests, use `run_test_for_class` instead of an explicit dependency check

thanks!
two ques:

* distfit's "popular" list includes distributions skpro doesn't have like `genextreme` and `dweibull`. should i limit what distfit can pick from or raise when it picks one?
* and should `_distfit_adapter.py` go entirely or kept as a fallback?

pyarchana:
to be clear on where this is: `run_test_for_class` is done and conflicts with main are resolved. the distribution mapping is waiting on your answer to the two ques above.

PR #1108, opened 2026-07-07
[ENH] Add NaiveProbaRegressor using distribution fitter

### LINKED ISSUE #1085: [ENH] `NaiveProbaRegressor` using distribution fitter
We should add a `NaiveProbaRegressor` using a distribution fitter.

It pools all data and fits a single distribution that is then broadcast to a 2D array.

The signature would be `NaiveProbaRegressor(distfitter=MyDistFitter(42))`, with a default of `distfitter` being the normal fitter with unbiased estimator.

patelchaitany:
Adds a naive baseline probabilistic regressor that pools all training target values, fits a single distribution via a configurable distfitter, and broadcasts that distribution at prediction time.

#### Reference Issues/PRs

Fixes #1085

### DIFF
--- a/skpro/regression/naive.py
+++ b/skpro/regression/naive.py
@@ -4,8 +4,6 @@
 __author__ = ["patelchaitany"]
 __all__ = ["NaiveProbaRegressor"]
 
-import numpy as np
-
 from skpro.regression.base import BaseProbaRegressor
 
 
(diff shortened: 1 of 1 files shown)


skpro/regression/naive.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            distfitter = self.distfitter.clone()
+
+        distfitter.fit(y)
+        self.distfitter_ = distfitter
+        self.distribution_ = distfitter.proba()
+
+        return self
+
+    def _predict(self, X):
```
fkiraly: this is already the default, so it does not need to be copy-pasted

skpro/regression/naive.py
```diff
@@ -0,0 +1,178 @@
...
+        y_pred_proba : skpro BaseDistribution
+            Predicted distribution, same number of rows as ``X``.
+        """
+        X_ind = X.index
+        X_n_rows = X.shape[0]
+
+        dist = self.distribution_
+        params = dist.get_params()
+
+        broadcast_params = {}
```
fkiraly: this is brittle since it works only for some distributions.

I think there should be a general public API mechanism for this kind of broadcasting.

I believe this exists, by wrapping distributions in the `IID` class - not sure if there is a more elegant way.
I could imagine, for instance, if the distribution is `"parametric"` (as per the distribution type tag) that we can do what you are doing below. But, not for all distributions.

fkiraly (changes requested):
Very nice - see above for a comment on ensuring that broadcasting always works.

We should, in particular, test the `EmpiricalFitter`, since there the current logic would fail!

patelchaitany:
@fkiraly, Let me know if any more changes needed

fkiraly (approved):
Great!

Issue #1109, opened 2026-07-08
[ENH] expand distribution fitters module with new fitters


patelchaitany:
**Is your feature request related to a problem? Please describe.**

The distfitter module only has 2 fitters (NormalFitter, MOMFitter). This limits the baselines available for
  unconditional density estimation and the NaiveProbaRegressor (#1085). Key distribution families (Exponential, Uniform,Laplace, Empirical) have no dedicated fitters, and there's no generic MLE fitting path.

**Describe the solution you'd like**

  Add 5 new distribution fitters to skpro/distfitter/:

  ExponentialFitter -  Exponential(rate)  - MLE: rate = n / sum(x)     
  UniformFitter  - Uniform(lower, upper) - Sample min/max  
  LaplaceFitter - Laplace(mu, scale) - Median + MAD (robust) or MLE 
  EmpiricalFitter - Empirical(spl) - Nonparametric, wraps raw samples
  MLEFitter - Any scipy-backed dist - Generic MLE via scipy.stats.fit() with declarative param_map

  All follow the existing BaseDistFitter contract (fit(X) - proba() - scalar BaseDistribution).

**Describe alternatives you've considered**

**Additional context**

Relates to roadmap issue #7 (baselines / unconditional density estimation)

MoeenAsim:
@patelchaitany  
Hi! I'd like to work on this enhancement. Since it includes five fitters, would you prefer a single PR implementing all of them, or would you be open to splitting it into smaller PRs?

patelchaitany:
@MoeenAsim there is already pr raised for this, but you could look into other issues on the skpro repo

PR #1110, opened 2026-07-09
[ENH] expand distfitter module with 5 new distribution fitters

patelchaitany:
Add ExponentialFitter, UniformFitter, LaplaceFitter, EmpiricalFitter, and MLEFitter to the distfitter module. Update test data to use positive values for ExponentialFitter compatibility.

#1109

  - Add ExponentialFitter - fits Exponential(rate) via MLE (rate = n / sum(x))
  - Add UniformFitter - fits Uniform(lower, upper) from sample min/max
  - Add LaplaceFitter - fits Laplace(mu, scale) via robust (median + MAD) or MLE
  - Add EmpiricalFitter - nonparametric fitter returning Empirical distribution from raw samples
  - Add MLEFitter - generic MLE fitter via scipy.stats.fit() with declarative param_map for any scipy-backed
  distribution
##### For all contributions

##### For new estimators


fkiraly (approved):
Looks good - could you first check though whether this `EmpiricalFitter` should be merged?
https://github.com/sktime/skpro/pull/1095

patelchaitany:
Oh thanks for pointing out, I will review that PR and remove the EmpiricalFitter from this PR

fkiraly (approved):
Looks great!

Btw, if you have an `__init__` without any content other than a `super` call, you can also just leave it be - the behaviour will the same. Not blocking, because it might be cleaner to actually have it for consistency reasons.

fkiraly (changes requested):
Actually, a change request:

* can you change the `MLEFitter` to `ScipyMLEFitter`, to indicate this uses `scipy` specifically?
* It would also be great if you could hardcode the `param_map` to sensible defaults for most `skpro` distributions, if it is not too much of a hassle? And then list in the docstring which ones are supported.

patelchaitany:
Renamed to ScipyMLEFitter and added built-in defaults for param_map, fit_kwargs, and scipy_dist across 24 distributions -so something like ScipyMLEFitter(dist=Exponential) works out of the box now. Skipped distributions with algebraic transforms (TruncatedNormal, NegativeBinomial) and discrete ones (no scipy .fit()), but those still work with an explicit param_map.

skpro/distfitter/_mlefitter.py
```diff
@@ -0,0 +1,330 @@
...
+        ``{"rate": ("scale", "inverse")}``
+
+        Example for ``Gamma(alpha=..., beta=...)``:
+        ``{"alpha": ("a", "identity"), "beta": ("scale", "inverse")}``
+    scipy_dist : scipy.stats distribution or None, optional (default=None)
+        The scipy distribution object whose ``.fit()`` method will be
+        called.  When ``None``, auto-detected from the skpro ``dist``
+        class (works for ``_ScipyAdapter`` subclasses and for supported
+        distributions listed above).
+    fit_kwargs : dict or None, optional (default=None)
```
fkiraly, line 168: I think these need to be spelled out.

Also, I think the scipy fitter supports MSE fitting as well?

skpro/distfitter/_mlefitter.py
```diff
@@ -0,0 +1,330 @@
...
+      ``InverseGaussian``, ``Laplace``, ``Levy``, ``LogGamma``, ``Normal``,
+      ``SkewNormal``, ``TDistribution``.
+
+    Sensible ``fit_kwargs`` (e.g. ``floc=0``) and ``scipy_dist`` are also
+    supplied automatically for the distributions listed above when the
+    respective arguments are left at their default ``None``.
+
+    Parameters
+    ----------
+    dist : skpro distribution class
```
fkiraly, line 137: question, not a blocking request: should this also support strings and instances?

Like for other cases where `dist` can be passed

skpro/distfitter/_mlefitter.py
patelchaitany: Okay, confirmed Scipy also support the MSE also

skpro/distfitter/_mlefitter.py
patelchaitany: Keeping it class-only for now the fitter only needs the class to construct a new distribution with fitted params.Instances would have their values silently ignored, and strings would require internal imports/registry lookups for no real gain over the user just importing the class directly.

patelchaitany:
@fkiraly , let me know if any more changes needed

fkiraly (changes requested):
Thanks!

May I request more testing of the `ScipyMLEFitter`?

We should test integrity of the parameter maps, and perhaps more families of distributions, given the amount of distributions and the different parameters that this supports.

patelchaitany:
What I changed here:

Added tests for the parameter maps. Two check every row of the table, one fits the same data through the fitter and through scipy and compares them.

Found three wrong maps while running these tests. Alpha, Beta and LogGamma let scipy work out loc and scale and then dropped them, so the fitted distribution was wrong. Fixed in this PR.

get_test_params now covers 22 of the 24 distributions instead of 2. Beta is left out because the test data is not between 0 and 1, and TDistribution because its cdf gives nan for large df, which is an existing issue in the distribution itself.

Fixed the capabilities:undefined tags on Alpha, HalfCauchy, Levy and Cauchy. They were saying mean and var are exact while returning inf. Without this the mean/variance test had to be weakened. Happy to split this out if you prefer.

fkiraly (approved):
Changed the API reference structure, but otherwise, great!

PR #803, opened 2026-03-06
[ENH] add explicit energy computations to 10 distributions

sakun15:
## What this PR does

Closes #267 (partial — adds energy for 10 distributions that previously used Monte Carlo approximation).

Adds `_energy_self` and `_energy_x` methods to:

- `HalfNormal` — numerical CDF integration (2∫F(1-F)dt)
- `HalfLogistic` — numerical CDF integration
- `Erlang` — numerical CDF integration (promoted from approx to exact)
- `Fisk` (log-logistic) — numerical CDF integration (promoted from approx to exact)
- `LogLaplace` — numerical CDF integration
- `SkewNormal` — numerical CDF integration (promoted from approx to exact)
- `BurrXII` — numerical CDF integration
- `BurrIII` — numerical CDF integration
- `FatigueLife` — numerical CDF integration
- `GeneralizedPareto` — numerical CDF integration with adaptive bounds

All use `scipy.integrate.quad` with deterministic quadrature, so all distributions are now listed under `capabilities:exact`.

## Verification

Checked against Monte Carlo (n=200k-500k): all implementations within 2% of MC estimate.
All 320 `test_methods_scalar[*-energy-*]` pytest tests pass.

PR #698, opened 2026-01-11
[DOC] Add formula docstring injection hooks for distribution methods

Tusharjamdade:
#### Reference Issues/PRs

Fixes #689

#### What does this implement/fix? Explain your changes.

Adds a mechanism to inject distribution-specific closed-form formulas into method docstrings (`pdf`, `log_pdf`, `cdf`, `pmf`) via class-level hooks, handled automatically in `BaseDistribution.__init_subclass__`.
#### What should a reviewer concentrate their feedback on?

- Docstring injection logic
- Method cloning safety
- Inheritance behavior

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
  How to: add yourself to the [all-contributors file](https://github.com/sktime/skpro/blob/main/.all-contributorsrc) in the `skpro` root directory (not the `CONTRIBUTORS.md`). Common badges: `code` - fixing a bug, or adding code logic. `doc` - writing or improving documentation or docstrings. `bug` - reporting or diagnosing a bug (get this plus `code` if you also fixed the bug in the PR).`maintenance` - CI, test framework, release.
  See here for [full badge reference](https://allcontributors.org/docs/en/emoji-key)
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
- [ ] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [ ] I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant `Examples` section.
- [ ] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured
  dependency isolation, see the [estimator dependencies guide](https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

PR #688, opened 2025-12-19
[ENH] Add explicit energy computations for multiple distributions

arnavk23:
#### Reference Issues/PRs
Towards #267

#### What does this implement/fix? Explain your changes.
- Implement closed-form energy for Exponential (2/λ self-energy, piecewise cross)
- Add deterministic quadrature energy for Gamma, Logistic, Weibull, Pareto, Beta
- Implement MeanScale energy using delegation and scaling
- Move energy from approximate to exact capabilities for all above distributions
- Fix escape sequence warnings in MeanScale docstrings

#### What should a reviewer concentrate their feedback on?
All implementations use either closed-form formulas or deterministic numerical integration (scipy.integrate.quad) instead of Monte Carlo approximation.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/skpro/blob/main/CONTRIBUTORS.md) with any new badges I've earned :-)
  How to: add yourself to the [all-contributors file](https://github.com/sktime/skpro/blob/main/.all-contributorsrc) in the `skpro` root directory (not the `CONTRIBUTORS.md`). Common badges: `code` - fixing a bug, or adding code logic. `doc` - writing or improving documentation or docstrings. `bug` - reporting or diagnosing a bug (get this plus `code` if you also fixed the bug in the PR).`maintenance` - CI, test framework, release.
  See here for [full badge reference](https://allcontributors.org/docs/en/emoji-key)
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators
- [x] I've added the estimator to the API reference - in `docs/source/api_reference/taskname.rst`, follow the pattern.
- [x] I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant `Examples` section.
- [ ] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured
  dependency isolation, see the [estimator dependencies guide](https://www.sktime.net/en/latest/developer_guide/dependencies.html#adding-a-soft-dependency).

PR #1114, opened 2026-07-20
[BUG] TruncatedPareto: add closed-form mean and var

binggao1230:

`TruncatedPareto` declares `mean` and `var` as `capabilities:exact`, but it has no closed-form implementation of either, so both fall back to `BaseDistribution`'s numeric approximation (averaging the ppf over 1000 equidistant-in-probability nodes) and emit the "does not have a numerically exact implementation ... filled in by approximating" warning. The `exact` claim is therefore false, and the value is inaccurate: the equidistant-in-probability average under-resolves the heavy upper tail, so the variance is understated on wide truncation windows — up to ~33% low.

Reported vs. true variance on current `main` (true = `scipy.integrate.quad` over the bounded support):

| `b, scale, [lower, upper]` | true var | reported var | error |
|---|---|---|---|
| 2, 1, [1, 10]  | 1.345902 | 1.314102 | -2.36% |
| 3, 2, [2, 20]  | 1.972243 | 1.846493 | -6.38% |
| 2, 1, [1, 100] | 5.290077 | 3.531642 | **-33.24%** |
| 4, 1, [1, 1000] | 0.222220 | 0.187481 | -15.63% |

This adds closed-form `_mean` and `_var`. For shape `b`, scale `s`, support `[l, u]`, the raw moments are `E[X^k] = (b s^b / Z) ∫_l^u x^{k-b-1} dx` with `Z = (s/l)^b - (s/u)^b`. The integral is `(u^{k-b} - l^{k-b})/(k-b)` for `b != k`, and `ln(u/l)` at the removable singularity `b == k` (the log form the naive `α/(α-k)` expression misses — it applies to the mean at `b == 1` and to the second moment at `b == 2`). The tag is now true and no warning is emitted.

Verified against `scipy.integrate.quad` to `1e-9` across the parameter grid, including the `b == 1` and `b == 2` boundaries and a wide window. The new `test_truncated_pareto.py` fails on the current code (both on the mismatch and on the now-forbidden approximation warning) and passes with the fix; the full `test_all_distrs.py` battery stays green.

This is in line with the ongoing closed-form / capability-tag work — the explicit-closed-form drive (#688, #803, #698) and the sibling capability-tag fixes (#950, #962, #973).

#### Scope note — the QPD family shares the same false tag

Auditing all distributions for the same pattern, `QPD_S`, `QPD_B`, `QPD_U` and `QPD_Johnson` also list `mean`/`var` under `capabilities:exact` while falling back to the same numeric approximation. Unlike `TruncatedPareto` they have no elementary closed form, so the honest fix is either accurate quadrature or a tag correction (as in #950 / #962), not a closed form. I've left them out of this PR to keep it focused and would be happy to follow up on them separately if you'd like.

No.

#### PR checklist

- [x] The PR title starts with `[BUG]`.

### DIFF
--- a/skpro/distributions/tests/test_truncated_pareto.py
+++ b/skpro/distributions/tests/test_truncated_pareto.py
@@ -1,0 +1,77 @@
+"""Tests for TruncatedPareto distribution."""
+# copyright: skpro developers, BSD-3-Clause License (see LICENSE file)
+
+import warnings
+
+import numpy as np
+import pytest
+from numpy.testing import assert_allclose
+from scipy import integrate
+from scipy.stats import pareto
+
+from skpro.distributions.truncated_pareto import TruncatedPareto
+from skpro.tests.test_switch import run_test_module_changed
+
+# (b, scale, lower, upper), spanning wide windows and the b == 1 / b == 2
+# moment boundaries (where the naive power-law formula becomes logarithmic).
+MOMENT_CASES = [
+    (2.0, 1.0, 1.0, 10.0),
+    (3.0, 2.0, 2.0, 20.0),
+    (2.0, 1.0, 1.0, 100.0),  # wide window: base ppf-average is ~33% low here
+    (0.5, 1.0, 1.0, 50.0),
+    (1.0, 1.0, 1.0, 10.0),  # b == 1: mean uses the log form
+    (2.0, 1.0, 1.0, 50.0),  # b == 2: second moment uses the log form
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Thanks a lot - this looks great!

It would be nice if you could look at the QPD cases, too.
Noting that "exact" means numerically exact, up to a reasonable low tolerance. So, if the quadrature is "good enough", the tag should be "exact".

PR #1126, opened 2026-08-21
[DOC] add missing `Examples` sections to distribution docstrings

Ashish-Kumar-Dash:
#### Reference Issues/PRs

  None (found during a distribution-module audit).

  Nine distributions were out of line with the repo convention that each class docstring has an `Examples` section (which runs as a doctest under `--doctest-modules`):

  - **Seven had no `Examples`/doctest at all** — `BurrIII`, `BurrXII`, `FDist`,  `FatigueLife`, `LeftTruncated`, `Skellam`, `TruncatedPareto`. Added a minimal runnable `Examples` section (import + construct) to each.
  - **`GumbelL` and `GumbelR`** already had working doctests, but under a non-standard `Example` (singular) header, so they didn't register as the numpydoc `Examples` section. Fixed the header to `Examples`.

  Docs-only; no code or behavior change.

  No.

  - That each example uses a valid, representative parameter set.
  - `LeftTruncated`'s example composes another distribution  (`LeftTruncated(Poisson(mu=2), lower=0)`), confirm that's the clearest illustration.

  The added `Examples` blocks are themselves doctests 

  #### PR checklist

  - [x] Title starts with [DOC].
  - [x] Illustrative `Examples` section in each affected docstring (runs as doctest).
  - [x] `pytest --doctest-modules` passes on all nine files.
  - [x] pre-commit clean on changed files.

### DIFF
--- a/skpro/distributions/burr_iii.py
+++ b/skpro/distributions/burr_iii.py
@@ -24,6 +24,12 @@ class BurrIII(_ScipyAdapter):
         Shape parameter
     scale : float
         Scale parameter
+
+    Examples
+    --------
+    >>> from skpro.distributions import BurrIII
+    >>>
+    >>> d = BurrIII(c=2, d=3, scale=1)
     """
 
     _tags = {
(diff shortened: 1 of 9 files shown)


fkiraly (changes requested):
Perfect!

Please make sure that the examples are in a single box though.

That means also prefacing the line in-between with a `>>>` (and taking care that it has no trailing whitespace)

fkiraly (approved):
Thanks!

PR #1129, opened 2026-08-26
[ENH] `log_cdf` for distributions, exact for scipy-backed via `_ScipyAdapter`

Ashish-Kumar-Dash:

This one touches BaseDistribution (public API)
`BaseDistribution` had `log_pdf`/`log_pmf` but no `log_cdf`, so users needing left-tail log-probabilities had to compute `log(cdf(x))`, which underflows to `-inf` in the deep tail (`cdf` rounds to `0.0` in float64).

- Add `log_cdf` (public) and `_log_cdf` to `BaseDistribution`, mirroring `log_pdf`. The default `_log_cdf` returns `log(cdf(x))` with an approximation warning, so every distribution gains a `log_cdf`.
- Override `_log_cdf` in `_ScipyAdapter` with scipy's `logcdf`, which stays finite and accurate in the lower tail, giving all scipy-backed distributions an exact, tail-stable `log_cdf` (e.g. `norm.logcdf(-40) ≈ -804` vs `log(norm.cdf(-40)) = -inf`).

No, scipy is already a core dependency.

- The default `_log_cdf` placement/warning wording (mirrors the `_log_pdf` pattern).
- Whether `log_cdf` should also be wired into the generic `TestAllDistributions` contract and/or added to `capabilities:exact` for scipy-backed classes, I left both as follow-ups (subclass `_tags` replace rather than merge, so a central tag would not propagate). Happy to fold in if preferred.

`distributions/tests/test_log_cdf.py`:
- `test_scipy_log_cdf_exact`: `_ScipyAdapter.log_cdf` matches `scipy.stats.gamma.logcdf` exactly.
- `test_default_log_cdf_is_log_of_cdf`: base default equals `log(cdf)` for a non-scipy distribution (`Normal`).

#### PR checklist

- [x] Title starts with [ENH].
- [x] Tests for both the exact (scipy) and default (base) paths.
- [x] pre-commit + targeted pytest pass locally.


fkiraly (changes requested):
Great, thanks!

Can you please also add a documentation hook in the docstring, like for the other distribution methods? In the base class, `BaseDistribution`.

PR #1130, opened 2026-08-31
[MNT] add documentation link checker to CI

siddharth7113:

Adds a CI job that checks the external links in the documentation, using Sphinx's
built-in `linkcheck` builder.

Two changes:

**1. New workflow, `.github/workflows/linkcheck.yml`**

The job installs the `docs` extra and runs:

```
python -m sphinx -b linkcheck -j auto docs/source docs/_build/linkcheck
```

The `linkcheck` builder exits with a non-zero status when it finds a broken link, so
a broken link fails the job and reports it at `output.txt`.

This is a separate workflow from `test.yml`, because it sets
`paths-ignore: ['docs/**']`, so a job added to `test.yml` would not run on a
documentation-only pull request.

Triggers:

- Pull requests that touch `docs/`, `skpro/**/*.py`, `pyproject.toml`, or the workflow
  itself. Source files are included because docstrings are API reference pages.
- A weekly cron, matching the schedule already used by `update_contributors.yml`.

- `workflow_dispatch`, for manual runs.

**2. `linkcheck` options in `docs/source/conf.py`**

- `linkcheck_exclude_documents = [r"changelog"]`. The changelog holds 506 `:pr:` roles
  and 617 `:user:` roles, which `sphinx-issues` renders as roughly 570 unique
  github.com URLs. Checking them hits GitHub rate limits.
- `linkcheck_ignore` for `linkedin.com`, which returns HTTP 999 to non-browser clients.
- `linkcheck_anchors = False`, because anchors on pages rendered by JavaScript cannot be
  resolved from the raw HTML.

No.

It was verified by running the checker locally against the current `main`:

```
working 44   redirected 14   ignored 507   unchecked 189   broken 0
```

##### For all contributions

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,63 @@
+name: Docs link check
+
+on:
+  pull_request:
+    branches:
+      - main
+      - 'release**'
+    paths:
+      - 'docs/**'
+      - 'skpro/**/*.py'
+      - 'pyproject.toml'
+      - '.github/workflows/linkcheck.yml'
+  schedule:
+    - cron: '0 0 * * 6'
+
+  # Allows workflows to be manually triggered
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: linkcheck-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Great addition!

PR #1143, opened 2026-09-12
[ENH] ensure conditional execution of object-specific tests

### LINKED ISSUE #997: [ENH] Ensure conditional running of recently added tests
A larger number of tests have been recently added and merged. Most of these are specific for individual regressors or probability distributions.

To the object/estimator specific tests, we should add decorators that ensure they run on CI only if the underlying class has changed. The following decorator will do that:

```python
@pytest.mark.skipif(
    not run_test_for_class(ClassName),
    reason="run test only if tested object has changed",
)
```

adity1raut:
#### Reference Issues/PRs

Fixes #997.

A number of recently added tests are specific to a single estimator, distribution, or
module, but carry no conditional skip. They are therefore collected and executed in
every CI job, on every pull request, even when the code they cover is untouched. This
adds CI runtime without adding signal.

This PR adds the conditional skip decorators to those tests, using the existing
switches in `skpro/tests/test_switch.py`. No new machinery is introduced.

`run_test_for_class(ClassName)` is used for tests tied to specific objects:

| File | Tests | Object(s) |
| --- | --- | --- |
| `skpro/distributions/tests/test_normal_mixture.py` | 6 | `NormalMixture` |
| `skpro/regression/tests/test_mapie_v1.py` | 3 | the four MAPIE v1 regressors |
| `skpro/regression/tests/test_ondil.py` | 2 | `OndilOnlineGamlss` |
| `skpro/regression/tests/test_pipeline_transform_chain.py` | 1 | `Pipeline`, `ResidualDouble` |

`run_test_module_changed("skpro.<module>")` is used for tests that exercise a module or
base layer rather than one concrete class:

| File | Tests | Module |
| --- | --- | --- |
| `skpro/distributions/base/tests/test_multiindex.py` | 7 | `skpro.distributions` |
| `skpro/distributions/tests/test_proba_basic.py` | 7 | `skpro.distributions` |
| `skpro/regression/tests/test_bandwidth.py` | 5 | `skpro.regression` |

Both follow the decorator form already used across the test suite:

```python
@pytest.mark.skipif(
    not run_test_for_class(ClassName),
    reason="run test only if tested object has changed",
)
```

Besides change detection, `run_test_for_class` also checks the estimator's
`python_dependencies` tag. It therefore subsumes the explicit `_check_soft_dependencies`
guards that `test_mapie_v1.py` and `test_ondil.py` used, and those redundant guards are
removed in favour of the single switch.

Test logic is unchanged throughout. Only decorators, and the imports they need, are added.

No.

* The split between `run_test_for_class` and `run_test_module_changed`. The four
  object-specific files map cleanly onto single classes; `test_multiindex.py`,
  `test_proba_basic.py` and `test_bandwidth.py` cover a base layer or a module rather
  than one class, so the module-level switch is used there instead. Happy to move any of
  them to the class-level switch if you would prefer the finer granularity.
* The module granularity chosen for `test_bandwidth.py`. It is guarded on
  `skpro.regression`, following the existing convention of guarding on the top-level
  subpackage; `skpro.regression._bandwidth` would be tighter if you prefer.
* Removal of the `_check_soft_dependencies` decorators in `test_mapie_v1.py` and
  `test_ondil.py`, now covered by `run_test_for_class` via the `python_dependencies` tag.
  Verified below that those tests still skip when the dependency is absent.
* In `test_proba_basic.py`, `test_proba_example` is deliberately left untouched, as it is
  already unconditionally skipped via `pytest.mark.skip` pending #918. A conditional
  guard on top of that would be dead code.

No new tests are added. This PR changes only when existing tests run, so it is verified
by running the affected files in both switch states.

The seven affected test files, with the switch off (default):

```
31 passed, 9 skipped
```

The nine skips are the soft dependency skips for `mapie` and `ondil`, which are not
installed in the environment, plus the pre-existing `test_proba_example` skip. This
confirms that the dependency guards removed in favour of `run_test_for_class` still take
effect.

The same files with `--only_changed_modules True`:

```
24 passed, 16 skipped
```

The additional skips all report `run test only if tested object has changed`, which
confirms the new decorators engage. The module-level tests still run here because the
test files themselves live inside the modules they guard on and are modified by this PR,
which is the intended behaviour of `run_test_module_changed`.

Linting on the changed files, matching the `pre-commit` configuration:

```
flake8      -> passed
isort       -> passed
pydocstyle  -> passed
black       -> no new findings
```

The single pre-existing `black` finding on `test_bandwidth.py` is identical before and
after this change, and originates from a newer `black` than the pinned 23.7.0.

#### Any other comments?

With `--only_changed_modules True`, `run_test_for_class` runs a test only if the class,
one of its local parent classes, its covering test classes, its `tests:libs` modules, or
its dependency specifications in `pyproject.toml` have changed relative to `main`, and
only if its soft dependencies are installed. `run_test_module_changed` applies the
equivalent check at module granularity. When the flag is off, behaviour is unchanged and
everything still runs, so local runs and scheduled full runs are unaffected.

##### For all contributions

### DIFF
--- a/skpro/distributions/base/tests/test_multiindex.py
+++ b/skpro/distributions/base/tests/test_multiindex.py
@@ -8,6 +8,7 @@ import pandas as pd
 import pytest
 
 from skpro.distributions.normal import Normal
+from skpro.tests.test_switch import run_test_module_changed
 
 
 @pytest.fixture
@@ -16,6 +17,10 @@ def normal_dist():
     return Normal(np.array([[1, 2], [2, 3], [4, 5], [6, 7]]), 2, index=ix)
 
 
+@pytest.mark.skipif(
+    not run_test_module_changed("skpro.distributions"),
+    reason="run only if skpro.distributions has been changed",
+)
 def test_loc_partial_level(normal_dist):
     result = normal_dist.loc[1]
     expected_index = pd.MultiIndex.from_tuples([(1, 2), (1, 3)])
@@ -23,6 +28,10 @@ def test_loc_partial_level(normal_dist):
     assert result.mean().shape == (2, 2)
 
 
+@pytest.mark.skipif(
+    not run_test_module_changed("skpro.distributions"),
+    reason="run only if skpro.distributions has been changed",
+)
 def test_loc_full_tuple(normal_dist):
     result = normal_dist.loc[(2, 2)]
     expected_index = pd.MultiIndex.from_tuples([(2, 2)])
@@ -30,6 +39,10 @@ def test_loc_full_tuple(normal_dist):
     assert result.mean().shape == (1, 2)
 
 
+@pytest.mark.skipif(
+    not run_test_module_changed("skpro.distributions"),
+    reason="run only if skpro.distributions has been changed",
+)
 def test_loc_list_of_keys(normal_dist):
     result = normal_dist.loc[[(1, 2), (2, 3)]]
     expected_index = pd.MultiIndex.from_tuples([(1, 2), (2, 3)])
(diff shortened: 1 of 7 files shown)


adity1raut:
@fkiraly  PTAL

fkiraly (approved):
Thanks!

